PR #205, opened 2014-06-27
Update db-helper

kevinpostlewaite:
Update db-helper to pull ddl from installed plugins in addition to killbill modules.


ejde:
@kevinpostlewaite @pierre It looks like there are some POM issues with ci.

PR #314, opened 2015-04-17
Add DebugJettyRun to make it easier to debug the KillBill web application using Eclipse

gesterhuizen:


pierre:
Out of curiosity, what is the issue with `mvn jetty:run`?

FWIW, here is what I do:

```
MAVEN_OPTS="$MAVEN_OPTS -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=12345"
mvn -Dorg.killbill.server.properties=file:///path/to/killbill-server.properties jetty:run
```

Also, have you looked at our [embedded Jetty server](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/test/java/org/killbill/billing/jetty/HttpServer.java) we use for tests?

PR #325, opened 2015-05-19
Avoid NPE with Status#fromStatusCode

### LINKED ISSUE #316: Response.Status 408(Request Timeout) does not exist in jersey core 1.18.1 -> IllegalArgumentException
During updateSubscription (triggered by KillBillJavaClient) server throws TimeoutException while processing SubscriptionResource$EntitlementCallCompletion.withSynchronization. TimeoutException is catched by this method  but javax.ws.rs.core.Response.Status.fromStatusCode(408) returns NULL, because 408 is not yet implemented by jersey-core 1.18.1
So Response.status throws IllegalArgumentException if status is null!

http://docs.oracle.com/javaee/6/api/javax/ws/rs/core/Response.Status.html: 408 is not integrated
http://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Response.Status.html: 408 is integrated

When we recieve the timeout:
If timeout field, in the the client method (updateSubscription), is set to 0, all works fine -> no timeout exception. 
If timeout field, in the the client method (updateSubscription), is set to >=8, all works fine -> no timeout exception. 
If timeout field, in the the client method (updateSubscription), is set  to values between, timeout exception is thrown. 
I have no idea whats wrong with this behaviour, but this is not the point, the problem is that status 408 is not supported by jersey core even if the TimeoutException is raised.

We use killbill server 0.13.6 and java client 0.3.1.

timout

lstephen:
Fixes #316

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
@@ -447,7 +447,7 @@ public class SubscriptionResource extends JaxRsResourceBase {
             } catch (final InterruptedException e) {
                 return Response.status(Status.INTERNAL_SERVER_ERROR).build();
             } catch (final TimeoutException e) {
-                return Response.status(Status.fromStatusCode(408)).build();
+                return Response.status(408).build();
             } finally {
                 if (waiter != null) {
                     killbillHandler.unregisterCompletionUserRequestWaiter(waiter);


sbrossie:
It would be great to have a regression test so we don't have to run manual regression on our side.

lstephen:
Is there any current test covering the class change here, or should I be creating a new test for SubscriptionResource?

pierre:
We do already have some tests [here](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestEntitlement.java).

pierre:
@lstephen Any update?

lstephen:
Sorry, I haven't had a chance to look at adding a test case that covers this.

pierre:
No worries, let us know when you get back to it and if you need some help.

PR #384, opened 2015-08-24
Add -H support for db-helper

david-xie:
Sometimes user may use MySQL on other hosts, so that we should allow user to specify host for db-helper.

I added a new parameter "-H" for this. Actually, I prefer "-h", but it is already use for "help" action.

### DIFF
--- a/bin/db-helper
+++ b/bin/db-helper
@@ -19,7 +19,7 @@
 #                                                                                 #
 ###################################################################################
 
-#set -x
+# set -x
 
 HERE=`cd \`dirname $0\`; pwd`
 TOP=$HERE/..
@@ -27,6 +27,7 @@ TOP=$HERE/..
 POM="$TOP/pom.xml"
 
 ACTION=
+HOST="localhost"
 DATABASE="killbill"
 USER="root"
 PWD="root"
@@ -42,6 +43,7 @@ SKIP="(server)"
 function usage() {
     echo -n "./db_helper "
     echo -n " -a <create|clean|dump>"
+    echo -n " -H MySQL host (default = localhost)"
     echo -n " -d database_name (default = killbill)"
     echo -n " -u user_name (default = root)"
     echo -n " -p password (default = root)"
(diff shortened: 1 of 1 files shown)


david-xie:
Hi @pierre , I have finished postgres part. Before I submit that PR, I need to confirm one thing, that is our sql script does not have `create database` part, so that user has to create their database first. Is this our design or can we change this behaviour?

pierre:
>  I need to confirm one thing, that is our sql script does not have create database part, so that user has to create their database first. Is this our design or can we change this behaviour?

Yes, the current behavior expects the database (and schema for PostgreSQL) to exist. We can certainly change the behavior for the `CREATE` option (it would be nice for the operation to be idempotent, i.e. invoking `CREATE` on an existing database would simply re-apply the DDL and not fail if the database already exists).

Also, looking at the [script](https://github.com/killbill/killbill/blob/master/bin/db-helper#L163-L166), it looks like the `CLEAN` option is not implemented properly (it should simply run the `CLEAN_FILE`).

david-xie:
I tried to create database for PostgreSQL but it failed. Will find another way for that.

However, I got error using sql script for MySQL:

	root@ebfd95cc8923:/app# bin/db-helper -a create -h 192.168.99.101 -u root -p pass4u
	Applying new schema to database killbill
	Warning: Using a password on the command line interface can be insecure.
	ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 1
	root@ebfd95cc8923:/app# mysql --version
	mysql  Ver 14.14 Distrib 5.6.26, for Linux (x86_64) using  EditLine wrapper

The first line of sql script is:

    /*! use killbill; */

Any idea for this?

pierre:
> However, I got error using sql script for MySQL

Can you create a gist of the full DDL?

> Another question is, has our script already supported PostgreSQL?

You need to run the following [DDL](https://github.com/killbill/killbill/blob/master/util/src/main/resources/org/killbill/billing/util/ddl-postgresql.sql) first, before running the main (shared) one.

david-xie:
check this: https://gist.github.com/david-xie/5e43ae98094c7d5e7d53

PR #397, opened 2015-09-19
Metrics service

andrenpaes:
Makes sure killbill points to snapshot version of platform (with metrics service included).

See https://github.com/killbill/killbill-platform/pull/11


sbrossie:
Confused with the PR? Is that mistake?

andrenpaes:
This is an old PR related to the platform metrics change. I'll close it.

PR #402, opened 2015-10-05
issues:396 - UTC was set as default account timezone.

vp-maguero:
A default UTC value was added in AccountModelDao.
time_zone column was changed to NOT NULL.
An assert was added to TestAccountDao.java


pierre:
Thanks!

pierre:
v0.14.x to v0.18.x is a big upgrade! 😸 

I strongly encourage you to do (at least) 0.14.x -> 0.16.x -> 0.18.x. Migration files were introduced in the 0.16.x series (http://docs.killbill.io/latest/database_migrations.html). Before that, the wiki / release notes are the source of truth (e.g. https://github.com/killbill/killbill/wiki/Kill-Bill-Migration-0.14.x-to-0.16.x).

For that issue, in MySQL non-strict mode, if you have accounts without timezones, the alter will trigger a warning but I encourage you to backfill the table with what makes sense in your scenario.

If you hit any roadblock, feel free to start a thread on the mailing-list.

PR #403, opened 2015-10-05
Issue 119

jgomez-vp:
Adding Unit Test to verify the behavior when trying to Cancel a Subscription with an invalid Requested Date.
As requested:
https://groups.google.com/forum/?hl=en#!topic/killbilling-users/Vh0qYjw55ZM

### DIFF
--- a/subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCancel.java
+++ b/subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCancel.java
@@ -257,9 +257,5 @@ public class TestUserApiCancel extends SubscriptionTestSuiteWithEmbeddedDB {
         final DateTime invalidDate = subscription.getBundleStartDate().minusDays(3);
         // CANCEL in EVERGREEN period with an invalid Date (prior to the Creation Date)
         subscription.cancelWithDate(invalidDate, callContext);
-        assertListenerStatus();
-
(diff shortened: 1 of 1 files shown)


subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCancel.java
```diff
@@ -233,4 +233,33 @@ public void testUncancel() throws SubscriptionBillingApiException, SubscriptionB
...
+        currentPhase = subscription.getCurrentPhase();
+        assertEquals(currentPhase.getPhaseType(), PhaseType.EVERGREEN);
+
+        final DateTime invalidDate = subscription.getBundleStartDate().minusDays(3);
+        // CANCEL in EVERGREEN period with an invalid Date (prior to the Creation Date)
+        subscription.cancelWithDate(invalidDate, callContext);
+        assertListenerStatus();
+
+        testListener.pushExpectedEvent(NextEvent.CANCEL);
+        assertListenerStatus();
```
sbrossie: I am confused about that: If i understand correctly your test will throw at line 259 (since requestedDate is invalid), so then why would we expect a CANCEL event afterwards?

jgomez-vp: Sorry Stephane, you are right.
I've just updated the Test!

jgomez-vp:
Sorry Stephane, you are right.
I've just updated the Test!

Issue #162, opened 2014-03-04
Catalog name conflict


pierre:
It looks like the catalog system can be confused if one tries to load a catalog from a URI and an existing catalog with the same name exists in the classpath.

Details: https://groups.google.com/forum/#!topic/killbilling-users/BcYNdEQTBT0

jgomez-vp:
@pierre 
What are the expected solutions here? 
- Should we accept anyway the uploaded Catalog but rename it internally to avoid confussions?
- Should we reject the uploaded Catalog and drop an Exception informing the user that already exists a Catalog with the same name?

Thanks!

Javier.

pierre:
Note that this bug happens when loading the catalog from a [XML file](https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/DefaultCatalogService.java#L74), not when uploading a multi-tenant catalog.

The issue seems to be in the [VersionedCatalogLoader](https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/io/VersionedCatalogLoader.java#L65-L79): calling `loadDefaultCatalog` with `SpyCarBasic.xml`, `/var/tmp/SpyCarBasic.xml` or `file:///var/tmp/SpyCarBasic.xml` will always result in loading the `SpyCarBasic.xml` catalog from the classpath (the way [we build the URI](https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/io/VersionedCatalogLoader.java#L74) is wrong). Instead:

* `SpyCarBasic.xml` should be loaded from the classpath
* `/var/tmp/SpyCarBasic.xml` should be loaded from the classpath ([package](http://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getResourceAsStream%28java.lang.String%29) `var.tmp`)
* `file:///var/tmp/SpyCarBasic.xml` should be loaded from the filesystem

jgomez-vp:
Understood!
What I'm wondering now is how should I test the 3 different options...
All of them should be loaded from the catalina.properties file?

Like this:

Option 1:
org.killbill.catalog.uri=SpyCarCustom.xml

Option 2:
org.killbill.catalog.uri=/var/tmp/SpyCarCustom.xml

Option 3:
org.killbill.catalog.uri=file:///var/tmp/SpyCarCustom.xml

[...]

PR #404, opened 2015-10-06
Issue 162

jgomez-vp:
Fix for issue #162 
Removing an unnecessary block of comment that was always overwriting the Catalog (if exists) from the classpath.
Added Unit Tests to verify the 3 possible cases: 
* Load from Resource folder;
* Load from a folder inside the Resource folder;
* Load from external folder (file system)

### DIFF
--- a/catalog/src/test/java/org/killbill/billing/catalog/io/TestVersionedCatalogLoader.java
+++ b/catalog/src/test/java/org/killbill/billing/catalog/io/TestVersionedCatalogLoader.java
@@ -38,9 +38,9 @@ import org.killbill.billing.catalog.api.CatalogApiException;
 import org.killbill.billing.catalog.api.InvalidConfigException;
 import org.testng.Assert;
 import org.testng.annotations.Test;
-import com.google.common.io.Files;
 import org.xml.sax.SAXException;
 
+import com.google.common.io.Files;
 import com.google.common.io.Resources;
 
 public class TestVersionedCatalogLoader extends CatalogTestSuiteNoDB {
@@ -159,9 +159,8 @@ public class TestVersionedCatalogLoader extends CatalogTestSuiteNoDB {
 
     @Test(groups = "fast")
     public void testLoadCatalogFromExternalFile() throws CatalogApiException, IOException, URISyntaxException {
-        URL originURL = Resources.getResource("SpyCarBasic.xml");
-        File originFile = new File(originURL.toURI());
-        File destinationFile = new File("/var/tmp/SpyCarBasic.xml");
+        File originFile = new File(Resources.getResource("SpyCarBasic.xml").toURI());
+        File destinationFile = new File(Files.createTempDir().toString() + "/SpyCarBasicRelocated.xml");
         destinationFile.deleteOnExit();
         Files.copy(originFile, destinationFile);
         final VersionedCatalog c = loader.loadDefaultCatalog(destinationFile.toURI().toString());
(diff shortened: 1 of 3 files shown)


catalog/src/test/java/org/killbill/billing/catalog/io/TestVersionedCatalogLoader.java
```diff
@@ -37,6 +38,7 @@
 import org.killbill.billing.catalog.api.InvalidConfigException;
 import org.testng.Assert;
 import org.testng.annotations.Test;
+import org.weakref.jmx.internal.guava.io.Files;
```
pierre: This is an internal (relocated) class, use `com.google.common.io.Files` directly.

catalog/src/test/java/org/killbill/billing/catalog/io/TestVersionedCatalogLoader.java
```diff
@@ -126,4 +128,43 @@ public void testLoad() throws IOException, SAXException, InvalidConfigException,
...
+    @Test(groups = "fast", expectedExceptions = CatalogApiException.class)
+    public void testLoadCatalogFromInsideResourceWithBadFolderName() throws CatalogApiException {
+        loader.loadDefaultCatalog("com/acme2/SpyCarCustom.xml");
+    }
+
+    @Test(groups = "fast")
+    public void testLoadCatalogFromExternalFile() throws CatalogApiException, IOException, URISyntaxException {
+        URL originURL = Resources.getResource("SpyCarBasic.xml");
+        File originFile = new File(originURL.toURI());
+        File destinationFile = new File("/var/tmp/SpyCarBasic.xml");
```
pierre: Can you use a temporary directory (`/var/tmp` may not exist), using `File.createTempFile` for instance?

catalog/src/test/java/org/killbill/billing/catalog/io/TestVersionedCatalogLoader.java
```diff
@@ -126,4 +128,43 @@ public void testLoad() throws IOException, SAXException, InvalidConfigException,
...
+        destinationFile.deleteOnExit();
+        Files.copy(originFile, destinationFile);
+        final VersionedCatalog c = loader.loadDefaultCatalog(destinationFile.toURI().toString());
```
pierre: This test probably also passes even before the code change because the name of the xml file is the same (bug reported on the mailing-list). Can you change the name (e.g. `SpyCarBasicRelocated.xml`), just to ensure the right catalog is loaded?

jgomez-vp:
Done with fixes.

pierre:
Verified. Thanks!

FYI, I've [fixed](https://github.com/killbill/killbill/commit/3d2dfb5dcbd9efa8b03d9480ce8e5bdbf2ebca18) a few minor code style warnings. In case you are using [IntelliJ IDEA](https://www.jetbrains.com/idea/download/), settings to use are [checked-in](https://github.com/killbill/killbill/tree/master/.idea).

jgomez-vp:
Thanks @pierre !
Yes, I'm using IntelliJ IDEA and I have the updated settings in the .idea folder.
Do I have to import them from within IDEA or something?

pierre:
> Do I have to import them from within IDEA or something?

It should be detected automatically. In the preferences dialog, go to Editor then Code Style: the scheme "Project" should be selected.

One way to check if it has been taken into account is to go Editor, Code Style then Java. In the Imports tab, "Class count to use import with '\*'" and "Names count to use static import with '\*'" should both be set to 999.

Alternatively, you could try to reformat and optimize imports of `TestVersionedCatalogLoader.java` and there should be no change.

PR #405, opened 2015-10-06
Correct currency symbol for different Locale and Invoice Currency

### LINKED ISSUE #118: Invoice formatter display wrong currency symbol
First of all, the code is using the locale to set the currency formatter. For instance in 
DefaultInvoiceFormatter:

```
 public String getFormattedPaidAmount() {
        final NumberFormat number = NumberFormat.getCurrencyInstance(locale);
        return number.format(getPaidAmount().doubleValue());
 }
```

The number format returned is based on the locale which is 'good' so it knows how to display the dots, comma, ... but wrong it comes to display the currency symbol.

vp-maguero:
this fixed #118: Invoice formatter displays now correct currency symbol

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
@@ -17,12 +17,15 @@
 package org.killbill.billing.invoice.template.formatters;
 
 import java.math.BigDecimal;
-import java.text.NumberFormat;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.UUID;
 
 import org.joda.time.DateTime;
@@ -30,11 +33,6 @@ import org.joda.time.LocalDate;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 import org.killbill.billing.callcontext.InternalTenantContext;
-import org.killbill.billing.invoice.api.formatters.ResourceBundleFactory;
-import org.killbill.billing.tenant.api.TenantInternalApi;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import org.killbill.billing.catalog.api.Currency;
 import org.killbill.billing.currency.api.CurrencyConversion;
 import org.killbill.billing.currency.api.CurrencyConversionApi;
@@ -45,10 +43,13 @@ import org.killbill.billing.invoice.api.InvoiceItem;
 import org.killbill.billing.invoice.api.InvoiceItemType;
 import org.killbill.billing.invoice.api.InvoicePayment;
 import org.killbill.billing.invoice.api.formatters.InvoiceFormatter;
+import org.killbill.billing.invoice.api.formatters.ResourceBundleFactory;
 import org.killbill.billing.invoice.model.CreditAdjInvoiceItem;
 import org.killbill.billing.invoice.model.CreditBalanceAdjInvoiceItem;
 import org.killbill.billing.invoice.model.DefaultInvoice;
 import org.killbill.billing.util.template.translation.TranslatorConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Objects;
 import com.google.common.base.Strings;
@@ -70,6 +71,18 @@ public class DefaultInvoiceFormatter implements InvoiceFormatter {
     private final CurrencyConversionApi currencyConversionApi;
     private final InternalTenantContext context;
     private final ResourceBundleFactory bundleFactory;
+
+    public static Map<java.util.Currency, Locale> currencyLocaleMap;
+    {
+        currencyLocaleMap = new HashMap<java.util.Currency, Locale>();
+        for (Locale localeItem : Locale.getAvailableLocales()) {
+            try {
+                java.util.Currency currency = java.util.Currency.getInstance(localeItem);
+                currencyLocaleMap.put(currency, localeItem);
+            }catch (Exception e){
+            }
+        }
+    }
 
     public DefaultInvoiceFormatter(final TranslatorConfig config, final Invoice invoice, final Locale locale, final CurrencyConversionApi currencyConversionApi, final ResourceBundleFactory bundleFactory, final InternalTenantContext context) {
         this.config = config;
@@ -213,20 +226,35 @@ public class DefaultInvoiceFormatter implements InvoiceFormatter {
 
     @Override
     public String getFormattedChargedAmount() {
-        final NumberFormat number = NumberFormat.getCurrencyInstance(locale);
-        return number.format(getChargedAmount().doubleValue());
+        return getFormattedAmountByLocaleAndInvoiceCurrency(getChargedAmount());
     }
 
     @Override
     public String getFormattedPaidAmount() {
-        final NumberFormat number = NumberFormat.getCurrencyInstance(locale);
-        return number.format(getPaidAmount().doubleValue());
+        return getFormattedAmountByLocaleAndInvoiceCurrency(getPaidAmount());
     }
 
     @Override
     public String getFormattedBalance() {
-        final NumberFormat number = NumberFormat.getCurrencyInstance(locale);
-        return number.format(getBalance().doubleValue());
+        return getFormattedAmountByLocaleAndInvoiceCurrency(getBalance());
+    }
+
+    /**
+     * Returns the amount with the correct currency symbol
+     *
+     * @param amount
+     * @return
+     */
+    private String getFormattedAmountByLocaleAndInvoiceCurrency(BigDecimal amount) {
+        final DecimalFormat number = (DecimalFormat) DecimalFormat.getCurrencyInstance(locale);
+
+        java.util.Currency currency = java.util.Currency.getInstance(invoice.getCurrency().toString());
+        DecimalFormatSymbols dfs = number.getDecimalFormatSymbols();
+        dfs.setInternationalCurrencySymbol(currency.getCurrencyCode());
+        dfs.setCurrencySymbol(currency.getSymbol(currencyLocaleMap.get(currency)));
+        number.setDecimalFormatSymbols(dfs);
+
+        return number.format(amount.doubleValue());
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
```diff
@@ -71,6 +72,18 @@
     private final InternalTenantContext context;
     private final ResourceBundleFactory bundleFactory;
 
+    public static Map<java.util.Currency, Locale> currencyLocaleMap;
```
pierre: Could you put this initialization in the constructor of `DefaultInvoiceFormatterFactory` (and change the `DefaultInvoiceFormatter` constructor to accept it as an argument)? `static` initialization makes me nervous...

Also, feel free to expose a package protect getter in `DefaultInvoiceFormatterFactory`, so it's easier to update tests (e.g. `TestDefaultInvoiceFormatter`).

invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
```diff
@@ -213,20 +226,35 @@ public BigDecimal getBalance() {
...
-        final NumberFormat number = NumberFormat.getCurrencyInstance(locale);
-        return number.format(getBalance().doubleValue());
+        return getFormattedAmountByLocaleAndInvoiceCurrency(getBalance());
+    }
+
+    /**
+     * Returns the amount with the correct currency symbol
+     *
+     * @param amount
+     * @return
```
pierre: Either document the fields or simply put a non-Javadoc comment, e.g.:

```
// Returns the amount with the correct currency symbol
```

Otherwise, the build will fail on Java 8 where doclint is [enabled by default](http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html).

invoice/src/test/java/org/killbill/billing/invoice/template/formatters/TestDefaultInvoiceFormatter.java
```diff
@@ -187,6 +186,127 @@ public void testFormattedAmount() throws Exception {
     }
 
     @Test(groups = "fast")
+    public void testFormattedAmountFranceAndJPY() throws Exception {
+
+        final FixedPriceInvoiceItem fixedItem = new FixedPriceInvoiceItem(UUID.randomUUID(), UUID.randomUUID(), null, null,
+                                                                             UUID.randomUUID().toString(), UUID.randomUUID().toString(),
+                                                                             new LocalDate(), new BigDecimal("1500.00"), Currency.JPY);
+        final Invoice invoiceEUR = new DefaultInvoice(UUID.randomUUID(), new LocalDate(), new LocalDate(), Currency.JPY);
```
pierre: Rename to `invoiceJPY`, a bit confusing otherwise :smile_cat:

invoice/src/test/java/org/killbill/billing/invoice/template/formatters/TestDefaultInvoiceFormatter.java
```diff
@@ -187,6 +186,127 @@ public void testFormattedAmount() throws Exception {
...
+                    "</tr>",
+                    Locale.US);
+    }
+
+    @Test(groups = "fast")
+    public void testFormattedAmountUSAndGBP() throws Exception {
+        final FixedPriceInvoiceItem fixedItemEUR = new FixedPriceInvoiceItem(UUID.randomUUID(), UUID.randomUUID(), null, null,
+                                                                             UUID.randomUUID().toString(), UUID.randomUUID().toString(),
+                                                                             new LocalDate(), new BigDecimal("1499.95"), Currency.GBP);
+        final Invoice invoiceEUR = new DefaultInvoice(UUID.randomUUID(), new LocalDate(), new LocalDate(), Currency.GBP);
```
pierre: `invoiceGBP`.

invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
```diff
@@ -213,20 +226,35 @@ public BigDecimal getBalance() {
...
+     */
+    private String getFormattedAmountByLocaleAndInvoiceCurrency(BigDecimal amount) {
+        final DecimalFormat number = (DecimalFormat) DecimalFormat.getCurrencyInstance(locale);
```
pierre: Maybe rename it to `numberFormatter`?

invoice/src/test/java/org/killbill/billing/invoice/template/formatters/TestDefaultInvoiceFormatter.java
```diff
@@ -187,6 +186,127 @@ public void testFormattedAmount() throws Exception {
...
+                    "    <td class=\"amount\"><strong>{{invoice.formattedChargedAmount}}</strong></td>\n" +
+                    "</tr>\n" +
+                    "<tr>\n" +
+                    "    <td class=\"amount\"><strong>{{invoice.formattedPaidAmount}}</strong></td>\n" +
+                    "</tr>\n" +
+                    "<tr>\n" +
+                    "    <td class=\"amount\"><strong>{{invoice.formattedBalance}}</strong></td>\n" +
+                    "</tr>",
+                    "<tr>\n" +
+                    "    <td class=\"amount\"><strong>1 500,00 ￥</strong></td>\n" +
```
pierre: I'm not an expert but intuitively this seems wrong, as the Yen doesn't have cents. I believe you need to set in `getFormattedAmountByLocaleAndInvoiceCurrency`:

```
numberFormatter.setMaximumFractionDigits(currency.getDefaultFractionDigits());
```

To verify, could you add a test with the BTC currency, and verify we can display more than 2 digits (e.g. `105.28843439` is valid and probably shouldn't be rounded even with a French locale)?

invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
vp-maguero: No problem, I'll move it. Not sure what do you mean with add it to the constructor parameters because this method takes possible values from Locale.getAvailableLocales().

pierre: What I meant was that if `currencyLocaleMap ` becomes an instance variable of `DefaultInvoiceFormatterFactory`, you'll need to change the constructor of `DefaultInvoiceFormatter` to pass it as an argument (so it's available in `DefaultInvoiceFormatter`).

`DefaultInvoiceFormatterFactory` however should still have a no-argument constructor.

invoice/src/test/java/org/killbill/billing/invoice/template/formatters/TestDefaultInvoiceFormatter.java
vp-maguero: 1 - FractionDigits fixed.

2 - Running into an issue when try to create an instance of "BTC" java.util.Currency. This is not contains in java.util.Currency.getAvailableCurrencies()

Related [link] (http://stackoverflow.com/questions/20474336/implementing-bitcoin-and-java-util-currency).

Do you consider to refactor the code to have it working? or do you want to set a default ***currency*** for all not know currency codes?

pierre: > Running into an issue when try to create an instance of "BTC" java.util.Currency. This is not contains in java.util.Currency.getAvailableCurrencies()

Ah, interesting! For the test, let's use the Omani Rial instead (`OMR`), which has 3 decimal places.

> Do you consider to refactor the code to have it working? or do you want to set a default currency for all not know currency codes?

Thinking about it, we actually already use [Joda-Money](http://www.joda.org/joda-money/) in Kill Bill to handle such currencies. Could we refactor `getFormattedAmountByLocaleAndInvoiceCurrency` to simply use [MoneyFormatterBuilder](http://www.joda.org/joda-money/userguide.html) instead of `DecimalFormat` directly (it looks like it uses [DecimalFormat](https://github.com/JodaOrg/joda-money/blob/e8c83944f5d6407007d41de2fa89d79ab68f18af/src/main/java/org/joda/money/format/MoneyAmountStyle.java#L272) behind the scenes)?

vp-maguero: I was able to fix "BTC" currency and get more than 2 decimals depending on the currency format.

Another issue is that [DefaultAmountFormatter] (https://github.com/killbill/killbill/blob/master/util/src/main/java/org/killbill/billing/util/DefaultAmountFormatter.java#L23) is always [rounding] (https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java#L201) with scale = 2. 

Input:
```
new BigDecimal("1105.28843439")
```
Then:
```
BTC1,105.29000000
```

invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
vp-maguero: got it.

invoice/src/test/java/org/killbill/billing/invoice/template/formatters/TestDefaultInvoiceFormatter.java
pierre: > Another issue is that [DefaultAmountFormatter] (https://github.com/killbill/killbill/blob/master/util/src/main/java/org/killbill/billing/util/DefaultAmountFormatter.java#L23) is always [rounding] (https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java#L201) with scale = 2.

Weird, I cannot think of the reason why we forced that rounding (introduced by https://github.com/killbill/killbill/commit/0cab890a68df0b19c01886f2885ca9695d70600a). Maybe before we didn't enforce a default locale (https://github.com/killbill/killbill/commit/abed7b57fbdb6c2eef4bc188d50820a325b51f51)?

I would remove the rounding and verify other tests still pass.

vp-maguero:
Comments are fixed by last commit 43afb0759795e3ffdf748925e592b65738e00f25

pierre:
Thanks!

FYI, I've made a few small fixes:

* [trivial reformatting](https://github.com/killbill/killbill/commit/896ec16586e33e5867bbba8eaa21392042704149)
* [share the currency -> locale map in tests](https://github.com/killbill/killbill/commit/8ea9e88c4d4b6818801e07295315f2d6777cf286)
* [reduce catch block](https://github.com/killbill/killbill/commit/abe94d64e4f6a833e179519f0ea621dcc0d9f84b)
* [remove third digit](https://github.com/killbill/killbill/commit/fbf57b8265bc2cb143d46061988bf6c990aa5116) in euro test (only 2 digits is supported in practice for cents) and added the [OMR](https://github.com/killbill/killbill/commit/5927458b3546d6e0a2cdb6574ce58d318d0e1b7b) test

Issue #409, opened 2015-10-14
Enhance logging by print context user token


pierre:
To enhance https://github.com/killbill/killbill/pull/406, we should print the user token when calling the plugins.

We should also make sure the context token is set using the request id, if set in JaxRs.

sbrossie:
Work it to at least add the context user token into the payment api log line.

sbrossie:
I am wondering if we should not make the remaining of that task more generic and make sure every log line prints the X-Request-Id (maybe part of structured logging work).

pierre:
Fixed in enhanced-logging branch.

PR #419, opened 2015-10-22
Add support for propagating the X-Request-Id header

andrenpaes:
Keeps backward compatibility to the old "X-Killbill-Request-Id-Req" header.

### DIFF
--- a/profiles/killbill/src/main/java/org/killbill/billing/server/filters/RequestDataFilter.java
+++ b/profiles/killbill/src/main/java/org/killbill/billing/server/filters/RequestDataFilter.java
@@ -19,6 +19,8 @@ package org.killbill.billing.server.filters;
 
 import java.util.List;
 
+import javax.ws.rs.core.HttpHeaders;
+
 import org.killbill.billing.util.UUIDs;
 import org.killbill.commons.request.Request;
 import org.killbill.commons.request.RequestData;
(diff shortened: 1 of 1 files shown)


profiles/killbill/src/main/java/org/killbill/billing/server/filters/RequestDataFilter.java
```diff
@@ -32,12 +34,13 @@
 @Singleton
 public class RequestDataFilter implements ContainerRequestFilter, ContainerResponseFilter {
 
+    private static final String REQUEST_ID_HEADER = "X-Request-Id";
 
-    private static final String REQUEST_ID_HEADER_REQ = "X-Killbill-Request-Id-Req";
+    private static final String LEGACY_REQUEST_ID_HEADER = "X-Killbill-Request-Id-Req";
```
pierre: :+1: for backward compatibility.

Nit: maybe add a `@Deprecated`?

pierre:
Since you are touching that code, maybe you could also address the second part of https://github.com/killbill/killbill/issues/409, i.e. set the `userToken` of the `CallContext` to the `requestId`, if set (see [Context.java](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/util/Context.java#L51-L52)). This would make the request id flow to the database (audit logs, bus events, notifications, etc.).

Unclear how big of a change it is, we could also do it in a another PR. The first part of https://github.com/killbill/killbill/issues/409 is definitively much bigger though.

profiles/killbill/src/main/java/org/killbill/billing/server/filters/RequestDataFilter.java
```diff
@@ -32,12 +34,13 @@
 @Singleton
 public class RequestDataFilter implements ContainerRequestFilter, ContainerResponseFilter {
 
+    private static final String REQUEST_ID_HEADER = "X-Request-Id";
```
sbrossie: Should it be `X-Request-ID` or `X-Request-Id`. Googling shows me a lot of `X-Request-ID`. Unclear...

andrenpaes: For future reference, according to the HTTP spec, headers are case insensitive, so either way should work.

Issue #199, opened 2014-05-29
Combo call to create subscriptions with add-ons


pierre:
It would be nice to have a combo call to create in a single transaction a subscription with its add-ons, to minimize the amount of invoices created.

See discussion https://groups.google.com/forum/#!topic/killbilling-users/w0ZAxa9Z0BA. This was also requested by @kevinpostlewaite a while back.

pierre:
At a JAX-RS level, we've had other requests for combo calls (e.g. create account, add a payment method   and trigger a purchase in one go). While we have a long way to go to make that happen transactionally, we could introduce new resources (i.e. Java classes) specifically for these calls. Specifically, we could keep `AccountResource` and add `AccountPaymentResource` (still under `/1.0/kb/accounts`). We could even keep the `POST /1.0/kb/accounts` and have Jersey auto-detect wether it's combo or not based on some hint (content-type header, etc.), hidden in the client libraries.

As part of this work, we should also maybe breakdown the existing resources into multiple classes to remove the hacks in [killpay](https://github.com/killbill/killbill/blob/e768d81593120299f7d0f578f43537466f4d5488/profiles/killpay/src/main/java/org/killbill/billing/server/modules/KillpayServerModule.java#L93) (required for Guice injection) and have real profiles.

jgomez-vp:
Hi @pierre, @kevinpostlewaite and @sbrossie.
We've been trying to find solutions for this ticket and we've realized the following:
- Multiple invoices are being created after the process because for each Subscription creation (Base + add ons) an event is being posted into the Bus inside the [DefaultSubscriptionDao](https://github.com/killbill/killbill/blob/master/subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java#L496-L506).
To make a modification there, we will have to change also all the intermediate classes (APIs, services, etc.) that are being used between the Resource and the DAO, which we think would be a big refactor.

- In the other hand, we could base our work in the comments of the mailing list discussion and create a new Resource that will execute the following steps:
  - Receive a JSON object with all the information of the Base Subscription and a list of Add Ons (String names) that must be added.
  - Add the AUTO_INVOICING_OFF tag to the Account.
  - Create the Base Subscription
  - Create the Add On Subscriptions (one for each Add On)
  - Remove the AUTO_INVOICING_OFF to the Account.

As a result, we can confirm that only one invoice is being generated with 1 item for the base subscription and N items for the N Add Ons specified in the List.
[...]

PR #426, opened 2015-10-29
Combo call to create subscriptions with add-ons. #199

maguero:
Involves releases to:
* killbill-api
* killbill-client-java
* killbill-plugin-api
* killbill-oss-parent
* killbill

### DIFF
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
@@ -157,8 +157,9 @@ public class DefaultEntitlementApi extends DefaultEntitlementApiBase implements 
     }
 
     @Override
-    public Entitlement createBaseEntitlementWithAddOns(final UUID accountId, final Iterable<EntitlementSpecifier> entitlementSpecifier, final LocalDate effectiveDate,
-                                                             final Iterable<PluginProperty> properties, final CallContext callContext) throws EntitlementApiException {
+    public Entitlement createBaseEntitlementWithAddOns(final UUID accountId, final String externalKey, final Iterable<EntitlementSpecifier> entitlementSpecifier,
+                                                       final LocalDate effectiveDate, final Iterable<PluginProperty> properties, final CallContext callContext)
+            throws EntitlementApiException {
 
         final EntitlementSpecifier baseSpecifier = Iterables.tryFind(entitlementSpecifier, new Predicate<EntitlementSpecifier>() {
             @Override
@@ -176,7 +177,7 @@ public class DefaultEntitlementApi extends DefaultEntitlementApiBase implements 
                                                                                null,
                                                                                null,
                                                                                baseSpecifier.getPlanPhaseSpecifier(),
-                                                                               baseSpecifier.getExternalkey(),
+                                                                               externalKey,
                                                                                baseSpecifier.getOverrides(),
                                                                                effectiveDate,
                                                                                properties,
@@ -188,11 +189,11 @@ public class DefaultEntitlementApi extends DefaultEntitlementApiBase implements 
                 final InternalCallContext contextWithValidAccountRecordId = internalCallContextFactory.createInternalCallContext(accountId, callContext);
 
                 try {
-                    if (entitlementUtils.getFirstActiveSubscriptionIdForKeyOrNull(baseSpecifier.getExternalkey(), contextWithValidAccountRecordId) != null) {
-                        throw new EntitlementApiException(new SubscriptionBaseApiException(ErrorCode.SUB_CREATE_ACTIVE_BUNDLE_KEY_EXISTS, baseSpecifier.getExternalkey()));
+                    if (entitlementUtils.getFirstActiveSubscriptionIdForKeyOrNull(externalKey, contextWithValidAccountRecordId) != null) {
+                        throw new EntitlementApiException(new SubscriptionBaseApiException(ErrorCode.SUB_CREATE_ACTIVE_BUNDLE_KEY_EXISTS, externalKey));
                     }
 
-                    final SubscriptionBaseBundle bundle = subscriptionBaseInternalApi.createBundleForAccount(accountId, baseSpecifier.getExternalkey(), contextWithValidAccountRecordId);
+                    final SubscriptionBaseBundle bundle = subscriptionBaseInternalApi.createBundleForAccount(accountId, externalKey, contextWithValidAccountRecordId);
 
                     final DateTime referenceTime = clock.getUTCNow();
                     final DateTime requestedDate = dateHelper.fromLocalDateAndReferenceTime(updatedPluginContext.getEffectiveDate(), referenceTime, contextWithValidAccountRecordId);
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
@@ -188,6 +188,10 @@ public class DefaultEntitlementApi extends DefaultEntitlementApiBase implements 
                 final InternalCallContext contextWithValidAccountRecordId = internalCallContextFactory.createInternalCallContext(accountId, callContext);
 
                 try {
+                    if (entitlementUtils.getFirstActiveSubscriptionIdForKeyOrNull(baseSpecifier.getExternalkey(), contextWithValidAccountRecordId) != null) {
+                        throw new EntitlementApiException(new SubscriptionBaseApiException(ErrorCode.SUB_CREATE_ACTIVE_BUNDLE_KEY_EXISTS, baseSpecifier.getExternalkey()));
+                    }
+
                     final SubscriptionBaseBundle bundle = subscriptionBaseInternalApi.createBundleForAccount(accountId, baseSpecifier.getExternalkey(), contextWithValidAccountRecordId);
 
                     final DateTime referenceTime = clock.getUTCNow();
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
@@ -31,6 +31,7 @@ import org.killbill.billing.callcontext.InternalTenantContext;
 import org.killbill.billing.catalog.api.BillingActionPolicy;
 import org.killbill.billing.catalog.api.PlanPhasePriceOverride;
 import org.killbill.billing.catalog.api.PlanPhaseSpecifier;
+import org.killbill.billing.catalog.api.ProductCategory;
 import org.killbill.billing.entitlement.AccountEventsStreams;
 import org.killbill.billing.entitlement.EntitlementService;
 import org.killbill.billing.entitlement.EventsStream;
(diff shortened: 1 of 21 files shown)


entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -156,6 +157,58 @@ public Entitlement doCall(final EntitlementApi entitlementApi, final Entitlement
...
+                                                                               properties,
+                                                                               callContext);
+
+        final WithEntitlementPlugin<Entitlement> createBaseEntitlementWithAddOn = new WithEntitlementPlugin<Entitlement>() {
+            @Override
+            public Entitlement doCall(final EntitlementApi entitlementApi, final EntitlementContext updatedPluginContext) throws EntitlementApiException {
+                final InternalCallContext contextWithValidAccountRecordId = internalCallContextFactory.createInternalCallContext(accountId, callContext);
+
+                try {
+                    final SubscriptionBaseBundle bundle = subscriptionBaseInternalApi.createBundleForAccount(accountId, baseSpecifier.getExternalkey(), contextWithValidAccountRecordId);
```
pierre: We need to check if a bundle for that key already exists:

```

if (entitlementUtils.getFirstActiveSubscriptionIdForKeyOrNull(externalKey, contextWithValidAccountRecordId) != null) {
  throw new EntitlementApiException(new SubscriptionBaseApiException(ErrorCode.SUB_CREATE_ACTIVE_BUNDLE_KEY_EXISTS, externalKey));
}
```

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementSpecifier.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.List;
+
+import org.killbill.billing.catalog.api.PlanPhasePriceOverride;
+import org.killbill.billing.catalog.api.PlanPhaseSpecifier;
+import org.killbill.billing.catalog.api.PlanSpecifier;
+
+public class DefaultEntitlementSpecifier implements EntitlementSpecifier {
+
+    private String externalkey;
+    private PlanPhaseSpecifier PlanPhaseSpecifier;
```
pierre: Typo: should be lowercased (`planPhaseSpecifier`).

Also, all fields can be `final`.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementSpecifier.java
```diff
@@ -0,0 +1,53 @@
...
+    private List<PlanPhasePriceOverride> overrides;
+
+    public DefaultEntitlementSpecifier(final String externalkey, final org.killbill.billing.catalog.api.PlanPhaseSpecifier planPhaseSpecifier, final List<PlanPhasePriceOverride> overrides) {
```
pierre: No need for fully qualified class.

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApi.java
```diff
@@ -485,4 +489,106 @@ public void testCreateEntitlementInThePast() throws AccountApiException, Entitle
...
+
+        EntitlementSpecifier baseEntitlementSpecifier = new DefaultEntitlementSpecifier("baseExternalKey", baseSpec, null);
+        EntitlementSpecifier addOnEntitlementSpecifier = new DefaultEntitlementSpecifier("addOnExternalKey", addOnSpec, null);
+
+        final List<EntitlementSpecifier> specifierList = new ArrayList<EntitlementSpecifier>();
+        specifierList.add(baseEntitlementSpecifier);
+        specifierList.add(addOnEntitlementSpecifier);
+
+        // Keep the same object for the whole test, to make sure we refresh its state before r/w calls
+        testListener.pushExpectedEvents();
```
pierre: Unnecessary, since no event is expected?

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApi.java
```diff
@@ -485,4 +489,106 @@ public void testCreateEntitlementInThePast() throws AccountApiException, Entitle
...
+
+        EntitlementSpecifier addOnEntitlementSpecifier1 = new DefaultEntitlementSpecifier("addOnExternalKey", baseSpec, null);
+        EntitlementSpecifier addOnEntitlementSpecifier2 = new DefaultEntitlementSpecifier("addOnExternalKey2", addOnSpec, null);
+
+        final List<EntitlementSpecifier> specifierList = new ArrayList<EntitlementSpecifier>();
+        specifierList.add(addOnEntitlementSpecifier1);
+        specifierList.add(addOnEntitlementSpecifier2);
+
+        // Keep the same object for the whole test, to make sure we refresh its state before r/w calls
+        testListener.pushExpectedEvents();
```
pierre: Unnecessary, since no event is expected?

subscription/src/main/java/org/killbill/billing/subscription/api/user/SubscriptionEspecifier.java
```diff
@@ -0,0 +1,106 @@
...
+ * under the License.
+ */
+
+package org.killbill.billing.subscription.api.user;
+
+import org.joda.time.DateTime;
+import org.killbill.billing.catalog.api.PhaseType;
+import org.killbill.billing.catalog.api.Plan;
+
+public class SubscriptionEspecifier {
```
pierre: Typo, should be `SubscriptionSpecifier`.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
pierre: Also, this won't be atomic with the creation of the subscriptions. But it looks like we already have the issue for `createBaseEntitlement`?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -208,7 +212,122 @@ public Response doResponseOk(final Entitlement createdEntitlement) {
...
     }
 
+    @Timed
+    @POST
+    @Path("/createEntitlementWithAddOns")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Create an entitlement with addOn products")
+    @ApiResponses(value = {@ApiResponse(code = 400, message = "Invalid entitlement supplied")})
+    public Response createEntitlementWithAddOns(final List<SubscriptionJson> entitlements,
```
pierre: Could you add a test at the JAX-RS level (see [TestEntitlement.java](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestEntitlement.java)). You will have to update the [Java client](https://github.com/killbill/killbill-client-java) to introduce a new API.

pierre:
@sbrossie Should we get rid of redundant internal APIs (i.e. `SubscriptionBaseInternalApi#createSubscription`, `SubscriptionBaseApiService#createPlan` and `SubscriptionDao#createSubscription `) and always call the new APIs instead (passing an empty list if needed)?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -156,6 +157,58 @@ public Entitlement doCall(final EntitlementApi entitlementApi, final Entitlement
     }
 
     @Override
+    public Entitlement createBaseEntitlementWithAddOns(final UUID accountId, final Iterable<EntitlementSpecifier> entitlementSpecifier, final LocalDate effectiveDate,
+                                                             final Iterable<PluginProperty> properties, final CallContext callContext) throws EntitlementApiException {
+
```
sbrossie: I am wondering if it makes sense to return the BASE Entitlement only or if we should return a `Iterable< Entitlement > ` with all the subscription created. Let's keep it that way for now.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -156,6 +157,58 @@ public Entitlement doCall(final EntitlementApi entitlementApi, final Entitlement
...
+        final EntitlementSpecifier baseSpecifier = Iterables.tryFind(entitlementSpecifier, new Predicate<EntitlementSpecifier>() {
+            @Override
+            public boolean apply(final EntitlementSpecifier specifier) {
+                return specifier.getPlanPhaseSpecifier() != null && ProductCategory.BASE.equals(specifier.getPlanPhaseSpecifier().getProductCategory());
+            }
+        }).orNull();
```
sbrossie: We could rely on convention from top to bottom of the stack to always pass the BASE first in the list (and just do a check at each level to verify the first entry is for the BASE).

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -156,6 +157,58 @@ public Entitlement doCall(final EntitlementApi entitlementApi, final Entitlement
...
+
+        if (baseSpecifier == null) {
+            throw new EntitlementApiException(new IllegalArgumentException(), ErrorCode.SUB_CREATE_NO_BP.getCode(), "Missing Base Subscription.");
+        }
+
+        final EntitlementContext pluginContext = new DefaultEntitlementContext(OperationType.CREATE_SUBSCRIPTION,
+                                                                               accountId,
```
sbrossie: Should we introduce a new `OperationType` like `CREATE_SUBSCRIPTIONS_WITH_AO` and then should we modify the `EntitlementContext`: Today it is specific to one `Entitlement` and we see in that structure fields like:
* `getPlanPhaseSpecifier()`
* `getExternalKey()`
* `getPlanPhasePriceOverride()`

Instead we should now probably have a `List<EntitlementSpecifier>`.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
sbrossie: @pierre Yes we already have the issue. I think we don't care too much about atomicity except for the `externalKey` that has been provided. So i would suggest to modify `SubscriptionBaseInternalApi` to have a `getOrCreateBundleForAccount`. That way if the cal fails right after we created the bundle, it can be retried and would work. In the worst caee it is not retried and we end up with an empty bundle (there is no real bad consequences for that).

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApi.java
```diff
@@ -485,4 +489,106 @@ public void testCreateEntitlementInThePast() throws AccountApiException, Entitle
...
+        } catch (EntitlementApiException e) {
+            assertEquals(e.getMessage(), "Missing Base Subscription.");
+        }
+        assertListenerStatus();
+
+        final List<Entitlement> allEntitlementsForAccount = entitlementApi.getAllEntitlementsForAccountId(account.getId(), callContext);
+        assertTrue(allEntitlementsForAccount.size() == 0);
+
+    }
+
```
sbrossie: Good tests but we will also need additional tests higher in the stack:
* [Beatrix level tests](https://github.com/killbill/killbill/tree/master/beatrix/src/test/java/org/killbill/billing/beatrix/integration) to test the new `EntitlementAPI` and verify that we indeed get only ONE invoice created  with all the invoice items.
* [JAXRS level tests](https://github.com/killbill/killbill/tree/master/profiles/killbill/src/test/java/org/killbill/billing/jaxrs) which will require implementing the java client as Pierre mentioned to verify the HTTP api works as expected.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
sbrossie: @pierre Agreed. See my previous remark about tests as well.

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
```diff
@@ -104,6 +109,52 @@ public DefaultSubscriptionBase createPlan(final SubscriptionBuilder builder, fin
...
+            }
+        }
+
+        final InternalCallContext internalCallContext = createCallContextFromBundleId(subscriptionBaseList.get(0).getBundleId(), context);
+        dao.createSubscriptionWithAddOns(subscriptionBaseList, eventsMap, internalCallContext);
+
+        try {
+            for (DefaultSubscriptionBase subscriptionBase : subscriptionBaseList) {
+                subscriptionBase.rebuildTransitions(dao.getEventsForSubscription(subscriptionBase.getId(), internalCallContext), catalogService.getFullCatalog(internalCallContext));
+            }
```
sbrossie: If we only return the BASE subscription (not the ADD_ON), we probably don't need to rehydrate (`rebuildTransitions `) all of them.

sbrossie:
@maguero Very good first pass!

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
maguero: Let me open a new killbill-plugin-api fork to work on this.

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApi.java
maguero: I'm working on Beatrix level test, but for JAXRS level I need to add the new service to killbill-client-java before.

maguero: done.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
maguero: done.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
maguero: > Instead we should now probably have a List<EntitlementSpecifier>.

My concern about this change is that when a killbill dev upgrade core version may have to refactor the plugins to work with this new List<EntitlementSpecifier> instead of original 3 values.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
sbrossie: @maguero Just talked to Pierre and we should keep the api the way it is. Ignore the comment.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
sbrossie: @maguero Just talked to Pierre and we should keep the api the way it is. Ignore the comment (i.e let's keep looking for right entry the way you did it).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
sbrossie: @maguero See api comment i added in the [github issue](https://github.com/killbill/killbill/issues/199#issuecomment-153828997)

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
sbrossie: @maguero Your concern is valid, the change in not backward compatible. This entitlement plugin api is not very stable (i think nobody uses it today but us internally). So i would proceed with the change:
1. Create new type `OperationType` `CREATE_SUBSCRIPTIONS_WITH_AO `
2. Modify context to pass a list of  `List<EntitlementSpecifier>` to reflect the multiple subscriptions (ideally BASE should be first entry).

sbrossie:
> @sbrossie Should we get rid of redundant internal APIs (i.e. SubscriptionBaseInternalApi#createSubscription, SubscriptionBaseApiService#createPlan and SubscriptionDao#createSubscription) and always call the new APIs instead (passing an empty list if needed)?

@pierre That would be a nice cleanup indeed. @maguero If you have a chance to do that, that would definitely help keep the code more maintainable.

sbrossie:
@maguero you will need to merge the latest master before we can merge your PR (unfortunately we had to touch the entitlement code recently (mostly moving code around).

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
sbrossie: @maguero Can you make that change as well (this is not a blocker but a nice to have)

maguero:
Merge is done. I'm going to work on entitlement-plugin-api change now.

maguero:
Hi guys, is there any pending task for this PR? Thanks.

pierre:
> Hi guys, is there any pending task for this PR? Thanks.

Sorry for the slow reply, but it is a big one! :smile_cat:

Everything looks good, we have other changes to push but then we'll be able to release all artifacts (api, plugin-api, etc.) and merge it.

Thanks!

maguero:
I supposed to! :sweat_smile:
Thank you!

Issue #385, opened 2015-08-26
JAXRS Metrics are incorrect sometimes


sbrossie:
We rely on `@Timed` annotation to gather automatically metrics from requests.

For most cases, this is very convenient and works well, but there are some situations where this breaks:

1. Multi purpose API calls like [createComboPayment](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/PaymentResource.java#L616). This call allows to do AUTH, PURCHASE and CREDIT, and all these operations would end up within the same counter

2. Multiple paths for the same call. An example is `voidPayment ` and `voidPaymentByExternalKey `.  Although the implementation is slightly different, the functionality is the same, and it *probably* make sense to combine those.

PR #432, opened 2015-11-03
Timed resource metrics

andrenpaes:
Replaces the `@Timed` annotations on jaxrs resources.

Solves #385 

Note: This PR adds the jersey-core jar as a dependency of killbill/jaxrs (see https://github.com/killbill/killbill/commit/c2eac6c4739db5b4e1b52942495105f5dd253913).
It's a transitive dependency that was removed when metrics-jersey was removed, but it's actually being directly used in the `org.killbill.billing.jaxrs.resources.PluginResource` class.

### DIFF
--- a/jaxrs/pom.xml
+++ b/jaxrs/pom.xml
@@ -55,16 +55,12 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>com.sun.jersey</groupId>
+            <artifactId>jersey-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.wordnik</groupId>
             <artifactId>swagger-annotations</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.dropwizard.metrics</groupId>
-            <artifactId>metrics-annotation</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.dropwizard.metrics</groupId>
-            <artifactId>metrics-jersey</artifactId>
         </dependency>
         <dependency>
             <groupId>javax.inject</groupId>
@@ -152,6 +148,10 @@
             <groupId>org.kill-bill.commons</groupId>
             <artifactId>killbill-embeddeddb-mysql</artifactId>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.kill-bill.commons</groupId>
+            <artifactId>killbill-metrics</artifactId>
         </dependency>
         <dependency>
             <groupId>org.kill-bill.commons</groupId>
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/ComboPaymentTransactionJson.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/ComboPaymentTransactionJson.java
@@ -45,6 +45,14 @@ public class ComboPaymentTransactionJson extends ComboPaymentJson {
         return transaction;
     }
 
+    public String getTransactionType() {
+        if (transaction != null) {
+            return transaction.getTransactionType();
+        }
+
+        return null;
+    }
+
     public Iterable<PluginPropertyJson> getTransactionPluginProperties() {
         return transactionPluginProperties;
     }
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
@@ -111,8 +111,8 @@ import org.killbill.billing.util.entity.Pagination;
 import org.killbill.billing.util.tag.ControlTagType;
 import org.killbill.billing.util.tag.Tag;
 import org.killbill.clock.Clock;
+import org.killbill.commons.metrics.TimedResource;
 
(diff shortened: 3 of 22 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/PluginResource.java
```diff
@@ -98,7 +97,6 @@ public PluginResource(@Named("osgi") final HttpServlet osgiServlet, // See Defau
         this.osgiServlet = osgiServlet;
     }
 
-    @Timed
     @DELETE
```
pierre: Missing `@TimedResource`?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -824,7 +823,7 @@ public PaymentJson apply(final Payment payment) {
         return Response.status(Response.Status.OK).entity(result).build();
     }
 
-    @Timed
+    @TimedResource
```
pierre: You probably want a `@MetricTag` here, as this will trigger an AUTH, PURCHASE or CREDIT.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -848,7 +847,7 @@ public Response processPaymentByExternalKey(final PaymentTransactionJson json,
         return processPayment(json, account, paymentMethodIdStr, paymentControlPluginNames, pluginPropertiesString, uriInfo, callContext);
     }
 
-    @Timed
+    @TimedResource
```
pierre: Same here.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/PaymentResource.java
```diff
@@ -228,7 +229,7 @@ public PaymentJson apply(final Payment payment) {
                                                );
     }
 
-    @Timed
+    @TimedResource
```
pierre: Same here.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
pierre: Also, for all these methods `byExternalKey`, you probably want to override the name in `@TimedResource` to combine with the `byId` method (point 2. of https://github.com/killbill/killbill/issues/385).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/PluginResource.java
andrenpaes: @sbrossie and I talked about this. We ended up deciding to remove the metric generation from this resource since they don't provide that much info.
Here's what the metric name would look like: `"kb_resource./plugins{subResources:.\*}.doPUT.PUT.2xx.230"`

Since plugins register servlets through OSGI to serve these requests, and we have metric for those, we'd be covered.

jaxrs/pom.xml
```diff
@@ -55,16 +55,12 @@
...
         <dependency>
-            <groupId>io.dropwizard.metrics</groupId>
-            <artifactId>metrics-annotation</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.dropwizard.metrics</groupId>
-            <artifactId>metrics-jersey</artifactId>
+            <groupId>com.wordnik</groupId>
+            <artifactId>swagger-annotations</artifactId>
         </dependency>
```
sbrossie: If we remove the `@Timed` annotation (and replace them with `@TimedResource`, do we still need the `swagger-annotations` ?

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/ComboPaymentTransactionJson.java
```diff
@@ -45,6 +45,14 @@ public PaymentTransactionJson getTransaction() {
         return transaction;
     }
 
+    public String getTransactionType() {
+        if (transaction != null) {
+            return transaction.getTransactionType();
+        }
+
+        return null;
```
sbrossie: nit: could be written in one line:
```
return transaction != null ? transaction.getTransactionType() : null;
```

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/PluginResource.java
pierre: :+1:

sbrossie:
Looks good over all. I would do a second pass to make sure that:
* There i no more `@Timed` anywhere
* All getById versus getByExternalKey apis (in all resources) are branching to the same metric
* The multi-operation calls are splitting to different metrics (using the `@Metrictag` logic)

jaxrs/pom.xml
andrenpaes: The `@Timed` annotation belongs to the metrics-annotation jar. Swagger has the annotations used for documenting the api.

Issue #315, opened 2015-04-29
Specifying a bundle external key when adding an AO throws a incorrect exception.


sbrossie:
See reported issue https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/killbilling-users/LIPQNoCSZQI/SgKiPlWt1ScJ

sbrossie:
We should obviously fix the wrong error code, but also allow users to ONLY specify the bundleExternalKey (in place of of the bundleId, which requires first fetching base subscription of bundle).

maguero:
this is fixed. PR: https://github.com/killbill/killbill/pull/434

sbrossie:
@maguero I think my comment was confusing. By *ONLY* i did not mean we should remove the ability to specify a `bundleId`, i meant that if a user were to *only* specify a `bundleExternalKey` things should work.

maguero:
No problem, I'll add bundleId back.
Thanks.

PR #434, opened 2015-11-04
#315 : allow users to ONLY specify the bundleExternalKey instead of b…

maguero:
allow users to ONLY specify the bundleExternalKey instead of bundleId when create an ADD_ON subscriptions.

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/killbilling-users/LIPQNoCSZQI/SgKiPlWt1ScJ

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
@@ -87,6 +87,8 @@ import org.killbill.commons.metrics.TimedResource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import com.google.inject.Inject;
 import com.wordnik.swagger.annotations.Api;
 import com.wordnik.swagger.annotations.ApiOperation;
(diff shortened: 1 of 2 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -162,14 +162,23 @@ public Response createEntitlement(final SubscriptionJson entitlement,
...
 
         final Iterable<PluginProperty> pluginProperties = extractPluginProperties(pluginPropertiesString);
+        final CallContext callContext = context.createContext(createdBy, reason, comment, request);
+
+        SubscriptionBundle bundle = null;
         final boolean createAddOnEntitlement = ProductCategory.ADD_ON.toString().equals(entitlement.getProductCategory());
         if (createAddOnEntitlement) {
-            verifyNonNullOrEmpty(entitlement.getBundleId(), "SubscriptionJson bundleId should be specified for ADD_ON");
+            verifyNonNullOrEmpty(entitlement.getExternalKey(), "SubscriptionJson externalKey should be specified for ADD_ON");
+            try {
```
sbrossie: You should allow either a `bundleId` or `bundleExternalKey`. If both are specified the code would use the `bundleId` and ignore the `bundleExternalKey` (ideally we should check the two point to the same object and throw an error if this is not the case, but i am sure there are many places where we don't check, so let's keep it simple).

sbrossie:
@maguero I did merge that a bit too quick. I made a small modification afterwards. You can take a [look](https://github.com/killbill/killbill/commit/ba6a41f2f94ad5ae48e3a67ea3e2182d9d759476) if you are interested

maguero:
It looks great!

Issue #306, opened 2015-03-22
'retired' option from catalog object does not seem to have been implemented


sbrossie:
The catalog `Product`,  `Plan` and `Pricelist` can take a `retired` parameter. However by looking through the code, i am not seeing any logic doing filtering on that attribute. However:
* Our main documentation claims this is mandatory to retire plans that way versus remove the entries from the catalog -- which in theory is not supported
 * Also, what should be the relationship between retiring at Product, Plan or Pricelist level. Do we need all of them for instance?
* Finally what should be the semantics associated with it? For e.g, hhould findPlan API return the `retired` entries or not?

sbrossie:
Actually there is some code that checks for that flag @see https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/VersionedCatalog.java#L337

But nobody seems to call that API.

maguero:
Hi @sbrossie, what we need to do for this issue? I was reviewing the code but I'm not sure what and where I need to implement it.

sbrossie:
@maguero We would like the ability to 'retire' plans in a multi-version catalog. Example:
* Customer subscribes to plan 'A' at date T1
* Later we decide to remove plan 'A at date T2
* Customer cannot subscribes to plan 'A' anymore but old subscription should still work as expected.

Initially the catalog code planned for that use case by introducing a `retired` flag but looking though the code it does not seem to be used very much. It also seems that by *just* removing the entry (`plan A` from the next catalog version, things should just work as expected.

So the work is:
1. Verify if today removing entries just produce the right result. We already have multi-version [catalog tests written in ruby](https://github.com/killbill/killbill-integration-tests/blob/master/killbill-integration-tests/core/test_catalog.rb), but we could also add some at the `beatrix` level in java.
2. If this is the case remove all `retired` code and update catalog doc to remove it; if not implement `retired` correctly

Note: The ability to retire catalog entries should work at the `Product`, `Plan` and `Pricelist` level.

maguero:
@sbrossie I tried your scenario and it is working as spected with Plans.

Steps:
* I setup versioned catalogs (4 "WeaponsHireSmall" catalogs)
* I created a subscription to plan 'Pistol' at date T1
* I move time to T2 (T2 > T1) and last versioned catalog started (without 'Pistol' plan).
[...]

PR #437, opened 2015-11-13
#306: "retired" option is removed.

maguero:


catalog/src/test/java/org/killbill/billing/catalog/TestCatalogService.java
```diff
@@ -33,7 +33,7 @@ public void testCatalogServiceDirectory() throws ServiceException, CatalogApiExc
         final DefaultCatalogService service = new DefaultCatalogService(new CatalogConfig() {
             @Override
             public String getCatalogURI() {
-                return "file:src/test/resources/versionedCatalog";
+                return "file:catalog/src/test/resources/versionedCatalog";
             }
```
sbrossie: I am confused about that change: what do we have a new `catalog` prefix ?

maguero: If I run it without "catalog/" it fails in my devbox :sweat:.
Can you please try it there?

sbrossie: We already have that issue with that [test](https://github.com/killbill/killbill/blob/master/catalog/src/test/java/org/killbill/billing/catalog/TestCatalogService.java#L36). If we run it from inside our IDE (IntelliJ, Eclipse,..) the test fails (unless you especially instruct your IDE to look from the module and not the top. The issue is because we specify relative path.

We need to make sure the tests pass when they are run using `mvn` from the top of the repo. I think you would need to remove that `catalog` piece for that to happen. Can you confirm?

maguero: ohh, interesting! You are right, the test without "catalog/" is failing on IntelliJ but it works fine on mvn command. Let me change it back.
Thank you!

PR #439, opened 2015-11-30
Create endpoint for refreshing of paymentmethods

holkra:
Haven't created any test so far. Feedback welcome which test actually would make sense here as the effect of this method call highly depends on the plugin that is responsible for the payment methods. And the endpoint doesn't do anything else than calling the service method.

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
@@ -766,7 +766,7 @@ public class AccountResource extends JaxRsResourceBase {
     }
 
     @TimedResource
-    @GET
+    @POST
     @Path("/{accountId:" + UUID_PATTERN + "}/" + PAYMENT_METHODS + "/refresh")
     @Produces(APPLICATION_JSON)
     @ApiOperation(value = "Refresh account payment methods", response = PaymentMethodJson.class, responseContainer = "List")
@@ -775,7 +775,6 @@ public class AccountResource extends JaxRsResourceBase {
     public Response refreshPaymentMethods(@PathParam("accountId") final String accountId,
                                           @QueryParam(QUERY_PAYMENT_PLUGIN_NAME) final String pluginName,
                                           @QueryParam(QUERY_PLUGIN_PROPERTY) final List<String> pluginPropertiesString,
-                                          @QueryParam(QUERY_AUDIT) @DefaultValue("NONE") final AuditMode auditMode,
                                           @HeaderParam(HDR_CREATED_BY) final String createdBy,
                                           @HeaderParam(HDR_REASON) final String reason,
                                           @HeaderParam(HDR_COMMENT) final String comment,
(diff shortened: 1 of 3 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -766,6 +766,30 @@ public PaymentMethodJson apply(final PaymentMethod input) {
     }
 
     @TimedResource
+    @POST
+    @Path("/{accountId:" + UUID_PATTERN + "}/" + PAYMENT_METHODS + "/refresh")
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Refresh account payment methods", response = PaymentMethodJson.class, responseContainer = "List")
```
pierre: Wrong `response` (nothing is returned).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -766,6 +766,30 @@ public PaymentMethodJson apply(final PaymentMethod input) {
...
+                                          @QueryParam(QUERY_PLUGIN_PROPERTY) final List<String> pluginPropertiesString,
+                                          @HeaderParam(HDR_CREATED_BY) final String createdBy,
+                                          @HeaderParam(HDR_REASON) final String reason,
+                                          @HeaderParam(HDR_COMMENT) final String comment,
+                                          @javax.ws.rs.core.Context final HttpServletRequest request) throws AccountApiException, PaymentApiException {
+        final Iterable<PluginProperty> pluginProperties = extractPluginProperties(pluginPropertiesString);
+        final CallContext callContext = context.createContext(createdBy, reason, comment, request);
+
+        final Account account = accountUserApi.getAccountById(UUID.fromString(accountId), callContext);
+        paymentApi.refreshPaymentMethods(account, pluginName, pluginProperties, callContext);
```
pierre: If `pluginName` is null, you need to call the other `refreshPaymentMethods` API.

pierre:
When you are done, can you squash the commits?

> Feedback welcome which test actually would make sense here as the effect of this method call highly depends on the plugin that is responsible for the payment methods.

Yes, the [mock plugin](https://github.com/killbill/killbill/blob/master/payment/src/test/java/org/killbill/billing/payment/provider/MockPaymentProviderPlugin.java#L288-L291) used in tests won't do much during a refresh. I would just check the return code (and try with a different input parameters, e.g. valid plugin name, invalid plugin name and no plugin name).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
holkra: Fixed.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
holkra: Fixed.

PR #440, opened 2015-11-30
Retrieve Catalog as JSON refactored.

maguero:


pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>killbill-oss-parent</artifactId>
         <groupId>org.kill-bill.billing</groupId>
-        <version>0.67</version>
+        <version>0.68-SNAPSHOT</version>
```
maguero: It contains the new version of killbill-client-java. PR: https://github.com/killbill/killbill-client-java/pull/22

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/CatalogJson.java
```diff
@@ -0,0 +1,549 @@
...
+            result = 31 * result + (billingPeriod != null ? billingPeriod.hashCode() : 0);
+            result = 31 * result + (phases != null ? phases.hashCode() : 0);
+            return result;
+        }
+    }
+
+    public static class PhaseJson {
+
+        private final String type;
+        private final List<PriceJson> prices;
```
sbrossie: We are missing the following:
* The Duration of the phase
* The FixedPrice (if any)

sbrossie:
Looks good. The PR does not address any of the usage sections. That would be a good addition but i also agree that make things more complicated. Maybe in a different bug/PR.

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/CatalogJson.java
sbrossie: @maguero I merged the PR as it was, but it would probably make a lot of sense to add those fields in the `PhaseJson` because otherwise the client of that api would miss critical info about the catalog. Is that something you could add in a separate PR?

matias-aguero-hs: @sbrossie Sure. I'll do it.

PR #500, opened 2016-02-28
use consistent length for plan_name, phase_name and usage_name in db

almalki:
This fixes an error while subscribing to a plan with very long name. I think there should be consistent validation between what the catalog accepts and what other APIs accept.

`
{
  "className": "java.lang.RuntimeException",
  "code": null,
  "message": "org.h2.jdbc.JdbcSQLException: Value too long for column \"PLAN_NAME VARCHAR(64) DEFAULT NULL\": \"'plan-with-very-long-name-and-very-very-very-very-very-very-very-very-very-very-long-slug-stuff-with-very-long-name-and-... "stackTrace": []
}`


pierre:
Thanks! I've merged this in the work-for-release-0.17.0 branch, as no more DDL change is allowed in the 0.16x release train.

Issue #502, opened 2016-02-29
Usage API idempotency


pierre:
The goal is to have the user pass an identifier (called `tracking_id`) when recording usage. Subsequent requests with that tracking id should not result in duplicated usage being recorded.

Tasks:

- Add a `tracking_id varchar(128) NOT NULL` column to `rolled_up_usage`.
- Index: `CREATE INDEX rolled_up_usage_tracking_id_subscription_id_tenant_record_id ON rolled_up_usage(tracking_id, subscription_id, tenant_record_id);`.
- Update `RolledUpUsageModelDao`, `UsageRecord` and `SubscriptionUsageRecordJson` (as well as the java client library model) to allow passing the tracking id when recording usage (the key won't be returned in the GET APIs, as current APIs aggregate usage rows). If the id isn't set, a random UUID should be generated using the `UUIDs` helper class. Note that all `RolledUpUsageModelDao` rows will share the same id from `SubscriptionUsageRecordJson`.
- Add a transactional bulk API to `DefaultRolledUpUsageDao` and `RolledUpUsageDao` **to record all row transactionally**. For performance reasons, we should use the `@SqlBatch` annotation in `RolledUpUsageDao` (see example in [QueueSqlDao](https://github.com/killbill/killbill-commons/blob/master/queue/src/main/java/org/killbill/queue/dao/QueueSqlDao.java#L107-L109)).
- Add a check in `DefaultUsageUserApi` when recording usage if we have (at least) one row with the supplied tracking id. If so, a checked exception should be thrown (similar to [DefaultAccountUserApi](https://github.com/killbill/killbill/blob/master/account/src/main/java/org/killbill/billing/account/api/user/DefaultAccountUserApi.java#L84), let's introduce a new `UsageApiException` and `ErrorCode`).

Due to the API and DDL changes, work should be done against the `work-for-release-0.17.0` branch.

sbrossie:
I am wondering if we should call it `external_key`. Usually our apis include a `getByExternalKey` and as you mentioned this will not be returned by GET (aggregation). Maybe a `tracking_id` (to make it clear this is simply used to avoid dups) ?

pierre:
Sure, :+1: for `tracking_id`.

sbrossie:
Note that initial description has been updated with comments below.

almalki:
Hi pierre,

I have few questions:

[...]

PR #507, opened 2016-03-06
Usage API Idempotency

almalki:
https://github.com/killbill/killbill/issues/502

### DIFF
--- a/usage/src/main/java/org/killbill/billing/usage/dao/RolledUpUsageModelDao.java
+++ b/usage/src/main/java/org/killbill/billing/usage/dao/RolledUpUsageModelDao.java
@@ -26,25 +26,35 @@ import org.killbill.billing.util.entity.Entity;
 import org.killbill.billing.util.entity.dao.EntityModelDao;
 import org.killbill.billing.util.entity.dao.EntityModelDaoBase;
 
+import com.google.common.base.Strings;
+
 public class RolledUpUsageModelDao extends EntityModelDaoBase implements EntityModelDao<Entity> {
 
     private UUID subscriptionId;
     private String unitType;
     private LocalDate recordDate;
     private Long amount;
+    private String trackingId;
 
     public RolledUpUsageModelDao() { /* For the DAO mapper */ }
 
-    public RolledUpUsageModelDao(final UUID id, final DateTime createdDate, final DateTime updatedDate, final UUID subscriptionId, final String unitType, final LocalDate recordDate, final Long amount) {
+    public RolledUpUsageModelDao(final UUID id, final DateTime createdDate, final DateTime updatedDate, final UUID subscriptionId, final String unitType, final LocalDate recordDate, final Long amount, final String trackingId) {
         super(id, createdDate, updatedDate);
         this.subscriptionId = subscriptionId;
         this.unitType = unitType;
         this.recordDate = recordDate;
         this.amount = amount;
+
+        if(Strings.isNullOrEmpty(trackingId)){
+            this.trackingId = UUIDs.randomUUID().toString();
+        }
+        else {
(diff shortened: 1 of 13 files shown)


usage/src/main/java/org/killbill/billing/usage/dao/RolledUpUsageModelDao.java
```diff
@@ -26,25 +26,35 @@
...
-    public RolledUpUsageModelDao(final UUID id, final DateTime createdDate, final DateTime updatedDate, final UUID subscriptionId, final String unitType, final LocalDate recordDate, final Long amount) {
+    public RolledUpUsageModelDao(final UUID id, final DateTime createdDate, final DateTime updatedDate, final UUID subscriptionId, final String unitType, final LocalDate recordDate, final Long amount, final String trackingId) {
         super(id, createdDate, updatedDate);
         this.subscriptionId = subscriptionId;
         this.unitType = unitType;
         this.recordDate = recordDate;
         this.amount = amount;
+
+        if(Strings.isNullOrEmpty(trackingId)){
+            this.trackingId = UUID.randomUUID().toString();
```
pierre: :warning: `UUIDs` should be used instead (from `org.killbill.billing.util`).

Because Kill Bill generates so many ids, we have a special, optimized, implementation.

usage/src/main/resources/org/killbill/billing/usage/dao/RolledUpUsageSqlDao.sql.stg
```diff
@@ -17,10 +18,20 @@ tableValues() ::= <<
 , :unitType
 , :recordDate
 , :amount
+, :trackingId
 , :userName
 , :createdDate
 >>
 
+getRecordsCountByTrackingId() ::= <<
```
pierre: tl;dr: it's probably fine, but we will have to performance test it (and run `EXPLAIN` on it) with a very large table.

Just for the record: because we will insert only one batch at most for a given `subscription_id` / `tracking_id`, this will scan at most a few hundred leaves in the index (unless the DB has some statistics on it). The performance should probably be the same as:

```
select
  1
where subscription_id = :subscriptionId
and tracking_id = :trackingId
<AND_CHECK_TENANT()>
limit 1
;
```

which is the approach I was thinking of intuitively originally (since we don't really need the count).

usage/src/main/resources/org/killbill/billing/usage/ddl.sql
```diff
@@ -8,6 +8,7 @@ CREATE TABLE rolled_up_usage (
     unit_type varchar(255),
     record_date date NOT NULL,
     amount bigint NOT NULL,
+    tracking_id varchar(128) NOT NULL,
```
pierre: Could you also update the [wiki](https://github.com/killbill/killbill/wiki/Kill-Bill-Migration-0.16.x-to-0.18.x) with the `ALTER` statement for the migration?

pierre:
A few comments, but otherwise, good job! :+1:

sbrossie:
@almalki If you could review/implement the proposed changes, we'd be more than happy to merge the PR.

almalki:
@sbrossie sure, I just got busy with other stuff, I will update the PR with the requested changes in a day or two.

usage/src/main/resources/org/killbill/billing/usage/ddl.sql
almalki: Done

almalki:
all comments were addressed

pierre:
Note that I had to fix the `create()` statement to make `SqlBatch` work properly. See c8ba522347161e12ba98221b830820a5f2bddd1f.

PR #511, opened 2016-03-16
Add Access-Control-Expose-Headers to ResponseCorsFilter

beccagaspard:
Add Access-Control-Expose-Headers to fix "Refused to get unsafe header" error when calling API from web browser via CORS


sbrossie:
@beccagaspard FYI: We are planning on releasing 0.16.3 early next week, so that should be part of it as well.

PR #517, opened 2016-03-25
Allow cookies for CORS requests

beccagaspard:
Would like to be able to use a session cookie for CORS requests to the KillBill API instead of including the `Authorization` header w/ username and password on every request.  Setting the `Access-Control-Allow-Credentials` header allows the use of cookies, but prohibits using `*` for  `Access-Control-Allow-Origin`, so I'm just echoing back the request origin.  I know this isn't a perfect solution.. ideally we could check against a configurable list of whitelisted origins.  Any thoughts? :smiley:

### DIFF
--- a/profiles/killbill/src/main/java/org/killbill/billing/server/filters/ResponseCorsFilter.java
+++ b/profiles/killbill/src/main/java/org/killbill/billing/server/filters/ResponseCorsFilter.java
@@ -26,12 +26,14 @@ import javax.servlet.FilterConfig;
 import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import org.killbill.billing.jaxrs.resources.JaxrsResource;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
+import com.google.common.net.HttpHeaders;
 
 @Singleton
 public class ResponseCorsFilter implements Filter {
@@ -39,8 +41,9 @@ public class ResponseCorsFilter implements Filter {
     private final String allowedHeaders;
 
     public ResponseCorsFilter() {
-        allowedHeaders = Joiner.on(",").join(ImmutableList.<String>of("Authorization",
-                                                                      "Content-Type",
+        allowedHeaders = Joiner.on(",").join(ImmutableList.<String>of(HttpHeaders.AUTHORIZATION,
+                                                                      HttpHeaders.CONTENT_TYPE,
+                                                                      HttpHeaders.LOCATION,
                                                                       JaxrsResource.HDR_API_KEY,
                                                                       JaxrsResource.HDR_API_SECRET,
                                                                       JaxrsResource.HDR_COMMENT,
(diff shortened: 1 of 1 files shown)


pierre:
Tricky... :sweat_smile: 

Question: have you verified [Swagger](http://127.0.0.1:8080/api.html) still works locally after the change? Quick testing shows that no `Origin` is set in that case (it is set however on our [website](http://killbill.io/api)).

Having an option to configure whitelisting sounds good regardless (probably in `KillbillServerConfig`).

Relevant:

* http://stackoverflow.com/questions/19743396/cors-cannot-use-wildcard-in-access-control-allow-origin-when-credentials-flag-i
* http://stackoverflow.com/questions/1653308/access-control-allow-origin-multiple-origin-domains (first answer)

sbrossie:
Ok not an export in CORS from reading the links, it seems like standard practice is to do the following:
1. Remove the `Access-Control-Allow-Origin = *` as it is too permissive
2. Allow the server (Kill Bill) to be configured with a whitelist of hosts that can then be returned in the `Access-Control-Allow-Origin`

A good place to add the whitelist config (as mentioned by Pierre above) is the `KillbillServerConfig`

@beccagaspard Does that make sense?

beccagaspard:
@pierre - I did verify that swagger was still working locally - I think it should be fine as long as swagger-ui and KillBill are running on the same server/port (according to their [documentation](https://github.com/swagger-api/swagger-ui/blob/master/README.md#cors-support))... but I will investigate how to get that header set for swagger.

@sbrossie - Yup, makes sense - I will work on adding config to the `KillbillServerConfig`, leaving the current behavior of `Access-Control-Allow-Origin = *` as a default so everything will be backwards compatible.   :sunglasses:

pierre:
@beccagaspard Trying to close the loop on this before 0.18 - any update regarding the (backward-compatible) changes we had talked about?

beccagaspard:
Sorry, totally lost track of this pull request... I seem to remember that as I was implementing this, I was wondering if we could use the container-provided CORS filter instead..?  It has all these feature and more implemented already.  Similar to the approach proposed for Location headers in #566. 

https://tomcat.apache.org/tomcat-7.0-doc/config/filter.html#CORS_Filter
https://tomcat.apache.org/tomcat-7.0-doc/api/org/apache/catalina/filters/CorsFilter.html

Any thoughts? 😄

sbrossie:
I like the proposal, but i think this needs to be validated through testing
(hard to really answer in the abstract).

pierre:
>  I seem to remember that as I was implementing this, I was wondering if we could use the container-provided CORS filter instead..?

It would be nice though that at least Swagger works out of the box, without any further configuration of Jetty, Tomcat, etc.

Since your patch is simple enough, I'm going to merge it as-is. Echoing the Origin back seems to work fine in my testing:

1. Open incognito window (to avoid any caching)
2. Go to google.com (chrome://newtab isn't a valid Origin)
3. Send the following request:
  ```
  var xhr = new XMLHttpRequest();
  xhr.open('GET', 'http://127.0.0.1:8080/1.0/kb/accounts/<UUID>');
  xhr.setRequestHeader('X-Killbill-ApiKey', 'bob');
  xhr.setRequestHeader('X-Killbill-ApiSecret', 'lazar');
  xhr.withCredentials = true;
  xhr.onload = function() { console.log('Success! body=' + xhr.responseText); };
  xhr.send();
  ```
* `OPTIONS` request doesn't send any Cookie nor Authorization header
* `GET` request prompts for a username and a password. The right Authorization header is sent and a Set-Cookie header is returned
4. Send the same script again. `GET` request submits both the Authorization header and the cookie - request is successful
5. Send a similar script with `xhr.setRequestHeader('Authorization', '');`. `GET` request submits both the empty Authorization header and the cookie - request is still successful
6. Swagger is unaffected

As far as configuring the Origin, happy to do that as well, although after further thoughts, I'm curious about the use-case. Do you have a custom UI making JS requests? Presumably, this UI is secured on a private network anyways?

PR #529, opened 2016-04-28
Map transaction status to different HTTP response codes

andrenpaes:


andrenpaes:
WIP - This is still missing mapping for Timeout exceptions during a request

Also, there's at least one test failing:
[org.killbill.billing.jaxrs.TestPayment#testWithFailedPayment](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPayment.java#L66)
The Kill Bill Java Client always expects an exception payload to be included in an HTTP response > 400:
https://github.com/killbill/killbill-client-java/blob/master/src/main/java/org/killbill/billing/client/KillBillHttpClient.java#L461

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -475,4 +477,32 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
+    private PaymentTransaction findCreatedTransaction(Payment payment, TransactionType transactionType, @Nullable String transactionExternalKey) {
+        for(PaymentTransaction transaction : payment.getTransactions()) {
+            if (transaction.getTransactionType() == transactionType
+                && Objects.equals(transaction.getExternalKey(), transactionExternalKey)) {
+                return transaction;
+            }
+        }
+        // If nothing is found, return last transaction
+        return Iterables.getLast(payment.getTransactions());
+    }
```
sbrossie: I think that works, but i then i am asking **myself**:
1. If `transactionExternalKey` was provided by user (is not null), then it should for sure be found and returned line 502 
2. If not, then it means  `transactionExternalKey` was **not** provided by user (is null), and then we default to last one

So, are we saying the `return Iterables.getLast(payment.getTransactions());` is the best we can return when no `transactionExternalKey` was provided, and that it **might return the wrong one** OR is it the case that it will always return the correct one, and in which case there is no need for the previous logic ;-)

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -475,4 +477,32 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
+            case PENDING:
+            case SUCCESS:
+                return uriBuilder.buildResponse(uriInfo, PaymentResource.class, "getPayment", payment.getId());
+            case PAYMENT_FAILURE:
+                return Response.status(402).build();
+            case PAYMENT_SYSTEM_OFF:
+            case UNKNOWN:
+                return Response.status(Status.SERVICE_UNAVAILABLE).build();
+            case PLUGIN_FAILURE:
+                return Response.status(502).build();
```
sbrossie: Sure `502 Bad gateway` works. I guess you are correct that most cases we get that is when we can't communicate correctly with the payment gateway.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -475,4 +477,32 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
+        final PaymentTransaction createdTransaction = findCreatedTransaction(payment, transactionType, transactionExternalKey);
+        switch (createdTransaction.getTransactionStatus()) {
+            case PENDING:
+            case SUCCESS:
+                return uriBuilder.buildResponse(uriInfo, PaymentResource.class, "getPayment", payment.getId());
+            case PAYMENT_FAILURE:
+                return Response.status(402).build();
+            case PAYMENT_SYSTEM_OFF:
+            case UNKNOWN:
+                return Response.status(Status.SERVICE_UNAVAILABLE).build();
```
sbrossie: I think `UNKNOWN` should be mapped the same way we map the timeout (and actually a timeout use case may very well result into an `UNKNOWN` transaction.

Then, the tricky cases left are the exception mapping.

sbrossie:
Looks good so far.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
andrenpaes: Should we at least check for the `transactionType` then? Return the last transaction of a transaction type. If that doesn't exist, we raise an exception

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -30,12 +30,14 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
```
pierre: ⚠️ That's Java 1.7+ only (we need to keep support for 1.6).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -475,4 +477,32 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
     }
 
+    protected Response createPaymentResponse(UriInfo uriInfo, Payment payment, TransactionType transactionType, @Nullable String transactionExternalKey) {
+        final PaymentTransaction createdTransaction = findCreatedTransaction(payment, transactionType, transactionExternalKey);
+        switch (createdTransaction.getTransactionStatus()) {
+            case PENDING:
+            case SUCCESS:
+                return uriBuilder.buildResponse(uriInfo, PaymentResource.class, "getPayment", payment.getId());
+            case PAYMENT_FAILURE:
+                return Response.status(402).build();
```
pierre: Could we keep the location header in failure scenarios too? That's the only way for the client to get the payment id (if no external key was specified).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -475,4 +477,32 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
+            case UNKNOWN:
+                return Response.status(Status.SERVICE_UNAVAILABLE).build();
+            case PLUGIN_FAILURE:
+                return Response.status(502).build();
+        }
+        // Should never happen
+        return Response.serverError().build();
+    }
+
+    private PaymentTransaction findCreatedTransaction(Payment payment, TransactionType transactionType, @Nullable String transactionExternalKey) {
```
pierre: Can `payment` be `null` here (aborted payment)? See the AccountResource.java diff above.

andrenpaes: I don't know. My idea was that it shouldn't be, and if it was null the code would fail with a 500. I'm not really sure what an aborted is and how should we handle it.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
andrenpaes: UKNOWN could mean other things, not necessarily a timeout. I think we should keep it as a 503 so we can differentiate from timeouts. And then, we can have the plugins throwing the right exception in case of a timeout, so it's handled by the exception mapper.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
andrenpaes: @pierre I've double checked this. If a payment is aborted by a payment plugin, then yes, `payment` will be `null` here. I'm not sure how we want to handle that case though, maybe throwing an exception is a better option?

pierre:
@andrenpaes Do you have the latest commits from the [work-for-release-0.16.4](https://github.com/killbill/killbill/commits/work-for-release-0.16.4) branch?

It looks like you are missing some of my changes regarding the handling of exceptions, such as:

* https://github.com/killbill/killbill/commit/eaa48eb5e946c2d235dd777cfd46bc3efab28f03
* https://github.com/killbill/killbill/commit/0d31b23cc7c2dea788e03da9d6c7b576b73f739b
* https://github.com/killbill/killbill/commit/d01484c5633db42ec2d79fb847617d4209cc6917
* https://github.com/killbill/killbill/commit/b12a00ea08e5be058f6b660dbaa94b0532e8d5ad
* https://github.com/killbill/killbill/commit/d84ba13acf27dcddd4dd31e555a28a98b9053c1c

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
pierre: It feels like it's similar to the `CANCELED` case, no (payment wasn't attempted)?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -475,4 +477,48 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
+            for (final PaymentTransaction transaction : matchingTransactions) {
+                if (Objects.equal(transaction.getExternalKey(), transactionExternalKey)) {
+                    return transaction;
+                }
+            }
+        }
+
+        // If nothing is found, return the latest transaction of given type
+        return Iterables.getFirst(matchingTransactions, null);
+    }
```
sbrossie: This is good, but this could be optimized a bit: everything could be done in one loop, by iterating from the end of it and checking for both `transactionType` and  `transactionExternalKey` if not null. 

This is fine to keep it that way as we expect small numbers of transactions/payments so multiple iterations should not impact perf very much.

payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java
```diff
@@ -124,6 +112,9 @@ private OperationException convertToUnknownTransactionStatusAndErroredPaymentSta
                                                                                                 null,
                                                                                                 null);
         paymentStateContext.setPaymentTransactionInfoPlugin(paymentInfoPlugin);
+        if (e instanceof OperationException) {
+            return (OperationException) e;
+        }
```
sbrossie: Does that case really exist?

andrenpaes: It could happen here:
https://github.com/andrenpaes/killbill/blob/http-error-responses/payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java#L86
But, to be honest, even though it's a checked exception, I'm not sure it's ever triggered.

The old code would re-wrap that into another OperationException, so I though it made sense to skip that.

andrenpaes:
I've rebased these changes against the 0.16.4 branch. I'll close this and I'll open a separate PR against the release branch.

Issue #394, opened 2015-09-05
Enhance logging of failed payment retries


pierre:
Currently, the stack trace looks like:

```
2015-09-05 16:30:16,024 [Thread-21] WARN  o.k.b.k.0.4.0.SNAPSHOT - [cybersource-plugin] Unsuccessful purchase: One or more fields contains invalid data
2015-09-05 16:30:16,046 [Thread-21] INFO  o.k.b.p.j.a.1.0.2.SNAPSHOT - Refreshing Analytics data for account XXX
2015-09-05 16:30:16,047 [notifications-th] WARN  o.k.b.payment.core.ProcessorBase - Failed to retry attempt YYY for plugins (__INVOICE_PAYMENT_CONTROL_PLUGIN__)
org.killbill.billing.payment.api.PaymentApiException: Internal payment error :
        at org.killbill.billing.payment.core.sm.PluginControlPaymentAutomatonRunner.run(PluginControlPaymentAutomatonRunner.java:126)
        at org.killbill.billing.payment.core.PluginControlPaymentProcessor.retryPaymentTransaction(PluginControlPaymentProcessor.java:208)
        at org.killbill.billing.payment.retry.DefaultRetryService.retryPaymentTransaction(DefaultRetryService.java:49)
        at org.killbill.billing.payment.retry.BaseRetryService$1.handleReadyNotification(BaseRetryService.java:74)
        at org.killbill.notificationq.NotificationQueueDispatcher.handleNotificationWithMetrics(NotificationQueueDispatcher.java:203)
        at org.killbill.notificationq.dispatching.NotificationCallableCallback.dispatch(NotificationCallableCallback.java:51)
        at org.killbill.notificationq.dispatching.NotificationCallableCallback.dispatch(NotificationCallableCallback.java:32)
        at org.killbill.queue.dispatching.Dispatcher$CallableQueue.call(Dispatcher.java:102)
        at org.killbill.queue.dispatching.Dispatcher$CallableQueue.call(Dispatcher.java:85)
        at org.killbill.commons.concurrent.WrappedCallable.call(WrappedCallable.java:42)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at org.killbill.commons.concurrent.WrappedRunnable.run(WrappedRunnable.java:40)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #534, opened 2016-05-03
Map transaction status to different HTTP response codes

andrenpaes:
Also, centralize exception handling for plugin calls thru `PaymentPluginDispatcher.dispatchWithExceptionHandling()`, so TimeOut exceptions happening during payment/control plugin calls are translated into a `ErrorCode.PAYMENT_PLUGIN_TIMEOUT` exception.

That simplified exception handling in general, but it has a few subtleties:
* `OperationControlCallback`'s dispatched task is expected to always throw an `OperationException` with the [correct result](https://github.com/killbill/killbill/pull/534/files#diff-f8373089685a563920750c610e88fbc3R148) (FAILURE, EXCEPTION, etc)
* An `OperationException` thrown by the dispatched task is wrapped in a `PaymentApiException` and unwrapped by [`unwrapExceptionFromDispatchedTask`](https://github.com/killbill/killbill/pull/534/files#diff-f8373089685a563920750c610e88fbc3R156)
* Time outs, interrupted and lock failed exceptions are wrapped in `PaymentApiException` and don't trigger a Plugin `onFailure` call (the part about not calling `onFailure` hasn't changed, it was already the behavior of the code)

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
@@ -544,12 +544,14 @@ public abstract class JaxRsResourceBase implements JaxrsResource {
             case SUCCESS:
                 return uriBuilder.buildResponse(uriInfo, PaymentResource.class, "getPayment", payment.getId());
             case PAYMENT_FAILURE:
+                // 402 - Payment Required
                 return Response.status(402).build();
             case PAYMENT_SYSTEM_OFF:
             case UNKNOWN:
-                // Return a 503
+                // 503 - Service Unavailable
                 return Response.status(Status.SERVICE_UNAVAILABLE).build();
             case PLUGIN_FAILURE:
+                // 502 - Bad Gateway
                 return Response.status(502).build();
         }
         // Should never happen
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
@@ -92,7 +92,6 @@ import org.slf4j.LoggerFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
-import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Strings;
@@ -570,7 +569,7 @@ public abstract class JaxRsResourceBase implements JaxrsResource {
 
         if (transactionExternalKey != null) {
             for (final PaymentTransaction transaction : matchingTransactions) {
-                if (Objects.equal(transaction.getExternalKey(), transactionExternalKey)) {
+                if (transactionExternalKey.equals(transaction.getExternalKey())) {
                     return transaction;
                 }
             }
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
@@ -30,14 +30,12 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.UUID;
 
 import javax.annotation.Nullable;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
-import javax.ws.rs.core.Response.StatusType;
 import javax.ws.rs.core.StreamingOutput;
 import javax.ws.rs.core.UriInfo;
 
@@ -94,12 +92,14 @@ import org.slf4j.LoggerFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
+import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Strings;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 
 public abstract class JaxRsResourceBase implements JaxrsResource {
 
@@ -535,8 +535,10 @@ public abstract class JaxRsResourceBase implements JaxrsResource {
         Preconditions.checkArgument(actual == expected, errorMessage);
     }
 
-    protected Response createPaymentResponse(UriInfo uriInfo, Payment payment, TransactionType transactionType, @Nullable String transactionExternalKey) {
+    protected Response createPaymentResponse(final UriInfo uriInfo, final Payment payment, final TransactionType transactionType, @Nullable final String transactionExternalKey) {
         final PaymentTransaction createdTransaction = findCreatedTransaction(payment, transactionType, transactionExternalKey);
+        Preconditions.checkNotNull(createdTransaction, "No transaction of type '%s' found", transactionType);
+
         switch (createdTransaction.getTransactionStatus()) {
             case PENDING:
             case SUCCESS:
@@ -553,14 +555,25 @@ public abstract class JaxRsResourceBase implements JaxrsResource {
         return Response.serverError().build();
     }
 
-    private PaymentTransaction findCreatedTransaction(Payment payment, TransactionType transactionType, @Nullable String transactionExternalKey) {
-        for(PaymentTransaction transaction : payment.getTransactions()) {
-            if (transaction.getTransactionType() == transactionType
-                && Objects.equals(transaction.getExternalKey(), transactionExternalKey)) {
-                return transaction;
+    private PaymentTransaction findCreatedTransaction(final Payment payment, final TransactionType transactionType, @Nullable final String transactionExternalKey) {
+        // Make sure we start looking from the latest transaction created
+        final List<PaymentTransaction> reversedTransactions = Lists.reverse(payment.getTransactions());
+        final Iterable<PaymentTransaction> matchingTransactions = Iterables.filter(reversedTransactions, new Predicate<PaymentTransaction>() {
+            @Override
+            public boolean apply(final PaymentTransaction input) {
+                return input.getTransactionType() == transactionType;
+            }
+        });
+
+        if (transactionExternalKey != null) {
+            for (final PaymentTransaction transaction : matchingTransactions) {
+                if (Objects.equal(transaction.getExternalKey(), transactionExternalKey)) {
+                    return transaction;
+                }
             }
         }
-        // If nothing is found, return last transaction
-        return Iterables.getLast(payment.getTransactions());
+
+        // If nothing is found, return the latest transaction of given type
+        return Iterables.getFirst(matchingTransactions, null);
     }
 }
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
@@ -30,12 +30,14 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 
 import javax.annotation.Nullable;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.core.Response.StatusType;
 import javax.ws.rs.core.StreamingOutput;
 import javax.ws.rs.core.UriInfo;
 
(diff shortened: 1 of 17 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -533,4 +535,48 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
+        final PaymentTransaction createdTransaction = findCreatedTransaction(payment, transactionType, transactionExternalKey);
+        Preconditions.checkNotNull(createdTransaction, "No transaction of type '%s' found", transactionType);
+
+        switch (createdTransaction.getTransactionStatus()) {
+            case PENDING:
+            case SUCCESS:
+                return uriBuilder.buildResponse(uriInfo, PaymentResource.class, "getPayment", payment.getId());
+            case PAYMENT_FAILURE:
+                // 402 - Payment Required
+                return Response.status(402).build();
```
pierre: Could we keep the location header in failure scenarios too? That's the only way for the client to get the payment id (if no external key was specified).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -533,4 +535,48 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
+        final Iterable<PaymentTransaction> matchingTransactions = Iterables.filter(reversedTransactions, new Predicate<PaymentTransaction>() {
+            @Override
+            public boolean apply(final PaymentTransaction input) {
+                return input.getTransactionType() == transactionType;
+            }
+        });
+
+        if (transactionExternalKey != null) {
+            for (final PaymentTransaction transaction : matchingTransactions) {
+                if (Objects.equal(transaction.getExternalKey(), transactionExternalKey)) {
```
pierre: Sure, that works, but since we know `transactionExternalKey != null`, you could simply have `transactionExternalKey.equals(transaction.getExternalKey())`.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -533,4 +535,48 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
         Preconditions.checkArgument(actual == expected, errorMessage);
     }
 
+    protected Response createPaymentResponse(final UriInfo uriInfo, final Payment payment, final TransactionType transactionType, @Nullable final String transactionExternalKey) {
+        final PaymentTransaction createdTransaction = findCreatedTransaction(payment, transactionType, transactionExternalKey);
+        Preconditions.checkNotNull(createdTransaction, "No transaction of type '%s' found", transactionType);
```
pierre: This will probably return 5xx. Shouldn't we return 404 instead?

payment/src/test/java/org/killbill/billing/payment/core/sm/TestPluginOperation.java
```diff
@@ -57,7 +57,7 @@
 
     private static final String PLUGIN_NAME_PLACEHOLDER = "pluginName";
 
-    private static final int TIMEOUT = 5;
+    private static final int TIMEOUT = 10;
```
pierre: Why this change?

payment/src/main/java/org/killbill/billing/payment/core/sm/control/OperationControlCallback.java
```diff
@@ -149,33 +145,19 @@ public OperationResult doOperationCallback() throws OperationException {
...
-            return (OperationException) originalExceptionOrCause;
-        } else if (originalExceptionOrCause instanceof LockFailedException) {
-            logger.warn("Failed to lock accountId='{}'", paymentStateContext.getAccount().getId());
-        } else if (originalExceptionOrCause instanceof TimeoutException) {
-            logger.warn("Call TIMEOUT for accountId='{}'", paymentStateContext.getAccount().getId());
-        } else if (originalExceptionOrCause instanceof InterruptedException) {
-            logger.warn("Call was interrupted for accountId='{}'", paymentStateContext.getAccount().getId());
-        } else {
-            logger.warn("Operation failed for accountId='{}'", paymentStateContext.getAccount().getId(), e);
+    protected OperationException unwrapExceptionFromDispatchedTask(final PaymentApiException e) {
```
pierre: Do we still need these `unwrapExceptionFromDispatchedTask`? There is no unwrapping done at this level anymore (at least the name should be changed).

payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java
```diff
@@ -86,35 +82,16 @@ public OperationResult doOperationCallback() throws OperationException {
         } else {
             try {
                 return doSimpleOperationCallback();
-            } catch (final Exception e) {
-                // We need to unwrap OperationException (see doSimpleOperationCallback below)
-                throw unwrapExceptionFromDispatchedTask(e);
-            }
-        }
+	        } catch (final OperationException e) {
```
pierre: Alignment issues?

pierre:
After re-reading the code and the diff, I don't see any major red flag (and 👍 if we can simplify our code). However, having recently spent a good amount of time in this piece of code, I also know it's quite easy to overlook a code path (we have a matrix of combinations, with the two state machines and the list of exceptions that can be thrown at various points)...

If it hasn't been done, we should at least do a round of manual testing. When I did my recent logging changes, I triggered manually:

* A state machine timeout with a control plugin taking too much time (before the request goes to the payment plugin)
* Ruby and Java Payment plugins timeouts
* Ruby and Java Payment plugins throwing a runtime exception

Each time, I verified the logs were meaningful (no unnecessary wrapping of exceptions or unnecessary logging of java.util exceptions -- see https://github.com/killbill/killbill/issues/394 for example).

Additionally:

* At the Java API level, what is the current status of our tests for this exception handling? Are we confident we have enough?
* Do the server tests pass as-is without any client changes? Can we add some tests when the mock plugin timeouts, throws an exception, etc. if we don't have any?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
sbrossie: That's actually a tricky one: Does it fail because client passed the wrong `transactionType` (=> `404`) or because our state is inconsistent (=> `5xx`). It would make sense to assume the former, but this is so not obvious.

Also it would be nice to use some kind of Preconditions in which case we might need to tweak the mapping into http status?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
sbrossie: I am actually not convinced about that one because the client has the ability to pass an `externalKey` which can be used to retrieve the payment. In theory `Location` header are specific to `201`, so unless we have good reason to twist http spec, i'd be more inclined to not return it (but i don't have a strong feeling, as long as we are consistent across our apis, which means that if we decide to return it, we should do a pass across all our apis to make sure this matches everywhere).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -533,4 +535,48 @@ protected void verifyNumberOfElements(int actual, int expected, String errorMess
...
+                if (Objects.equal(transaction.getExternalKey(), transactionExternalKey)) {
+                    return transaction;
+                }
+            }
+        }
+
+        // If nothing is found, return the latest transaction of given type
+        return Iterables.getFirst(matchingTransactions, null);
+    }
 }
```
sbrossie: Nit: Also as i mentioned yesterday on previous PR, this could be simplified into one iteration (instead of 3).

payment/src/test/java/org/killbill/billing/payment/core/sm/TestPluginOperation.java
andrenpaes: I was getting timeout exceptions while running some of the tests of this class.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
andrenpaes: We can definitely optimize it to do it in 1 iteration, but there's no performance gain in doing it.

My priority here was to communicate as clear as possible what the code is doing.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
andrenpaes: I don't expect this to ever be triggered, since we're always executing this code after a payment transaction was created, so a transaction should have been created, or at least already existed.

So, if this is triggered, it means something is wrong elsewhere in kb, unless I'm missing something.

payment/src/main/java/org/killbill/billing/payment/core/sm/control/OperationControlCallback.java
andrenpaes: We still need to either unwrap the `PaymentApiException` returned or wrap it into an `OperationException` instance.
Initially I've considered packing everything in `OperationCallbackBase` class, but the `PaymentOperation` class does some things of its own with errors thrown, so I'm not sure we can get rid of it.

payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java
andrenpaes: Tabs! :)
I'll fix it.

pierre: the-simpsons-episode-25 02-treehouse-of-horror-xxiv-promotional-photos-1_full

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
pierre: > So, if this is triggered, it means something is wrong elsewhere in kb, unless I'm missing something.

What about the aborted case? Wouldn't it trigger a NPE in `findCreatedTransaction`?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
pierre: One main use-case is to keep the `refresh` functionality in the clients. Even if we now return different return codes for the various states, the clients will still most likely want to access the payment object (to access the gateway error message for example).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
andrenpaes: Yeah, that is still broken, but now is broken uniformly :)

We could return a No Content response, but I'm not sure how helpful would that be without some hint on why the payment was aborted. Also, it's impossible to tell if the payment is null b/c it was aborted, or b/c some regression was introduced into code.

I believe throwing a NPE is probably the best (or least worst) way to go, since we don't ever expect aborted payments right now, and if it happens it will be easy it to notice.

pierre: > we don't ever expect aborted payments right now

Well, who is `we`? 😸  It's up to plugin developers to abort payments in their control plugin if needed (e.g. [fraud detection systems](https://github.com/killbill/killbill-accertify-plugin/blob/master/src/main/java/org/killbill/billing/plugin/accertify/api/AccertifyPaymentControlPluginApi.java#L71-L81)).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
sbrossie: I think we should talk about that and make sure we agree on what is the behavior we want. Probably, this is good enough for that upcoming release, and then we can do another pass on that topic.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
andrenpaes: I understand that, and I'm not saying we shouldn't fix. But I think we need a better solution to differentiate a `null` return from an aborted payment.
I think we'd benefit from having a separate issue to cover this as this PR is introducing a lot of changes already.

profiles/killbill/src/test/resources/killbill.properties
```diff
@@ -21,6 +21,8 @@ org.killbill.overdue.uri=overdue.xml
 
 org.killbill.payment.retry.days=8,8,8
 
+org.killbill.payment.plugin.timeout=8s
```
andrenpaes: The default timeout of the java client is 10 seconds, so I've lowered the threshold in kb (from the default 30s), so we could add a test case for timeouts.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
andrenpaes: Just to close the loop here, I'll be working on handling the abort payment scenario in another PR.

payment/src/main/java/org/killbill/billing/payment/core/sm/control/OperationControlCallback.java
andrenpaes: This part can be further simplified too, but I'll open a different PR so it's easier to test and review.

andrenpaes:
Ok, so I've added test cases covering most of the error handling scenarios. I've also cleaned up the logging and fixed a small issue in the `PaymentOperation`/`PaymentAutomaton` classes.

Regarding the responses, I've added the location header and error messages to the response payload. The java client is now able to handle the new error response correctly (by correctly, I mean raising a `KillBillClientException`).

Finally, I'll leave the abort payment handling to a different PR, since the changes here won't influence that case, as the latest changes for 0.16.4 introduced a regression in the `DefaultPaymentApi` class:
https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/api/DefaultPaymentApi.java#L128-L139

I think this is ready to be merged. Please take a look and let me know if you find any issues.

profiles/killbill/src/test/resources/killbill.properties
sbrossie: I think it would be better to see if we could overwrite the timeout value for that test specifically to something even smaller:

* `8 sec` is still quite long for a test that runs every time
* Also, that impacts all out tests (where default timeout is now 8 sec)

sbrossie:
@andrenpaes Looks good t me. 👍  on the new tests. Just have one comment about that default timeout value, but that's not a blocker, we can talk about it. If @pierre agrees, i am ready to merge

profiles/killbill/src/test/resources/killbill.properties
andrenpaes: I think this only impact the tests in the profile module. As far as I could tell, all tests that use this setting are in the slow group.

But yeah, it would be great to scope this to the test. I didn't want to set it lower since I was afraid we'd start getting timeouts in other tests that don't expect timeouts.

sbrossie:
This had been merged in branch `work-for-release-0.16.5`. Closing PR.

PR #536, opened 2016-05-06
Handle Control Plugin's Abort Payment instruction

andrenpaes:
Make sure an exception is thrown when a control plugin aborts a payment, and that the exception is correctly mapped to an HTTP response code*

\*Right now, the error is being mapped to [422 Unprocessable Entity](http://www.restpatterns.org/HTTP_Status_Codes/422_-_Unprocessable_Entity).

### DIFF
--- a/payment/src/main/java/org/killbill/billing/payment/core/sm/control/ControlPluginRunner.java
+++ b/payment/src/main/java/org/killbill/billing/payment/core/sm/control/ControlPluginRunner.java
@@ -73,7 +73,7 @@ public class ControlPluginRunner {
                                                              final boolean isApiPayment,
                                                              final List<String> paymentControlPluginNames,
                                                              final Iterable<PluginProperty> pluginProperties,
-                                                             final CallContext callContext) throws PaymentControlApiException, PaymentApiException {
+                                                             final CallContext callContext) throws PaymentControlApiException {
         // Return as soon as the first plugin aborts, or the last result for the last plugin
         PriorPaymentControlResult prevResult = new DefaultPriorPaymentControlResult(false, amount, currency, paymentMethodId, pluginProperties);
 
@@ -119,7 +119,7 @@ public class ControlPluginRunner {
                 inputPluginProperties = prevResult.getAdjustedPluginProperties();
             }
             if (prevResult.isAborted()) {
-                throw new PaymentApiException(ErrorCode.PAYMENT_PLUGIN_API_ABORTED, pluginName);
+                throw new PaymentControlApiAbortException(pluginName);
             }
             inputPaymentControlContext = new DefaultPaymentControlContext(account,
                                                                           inputPaymentMethodId,
--- a/payment/src/main/java/org/killbill/billing/payment/core/sm/control/ControlPluginRunner.java
+++ b/payment/src/main/java/org/killbill/billing/payment/core/sm/control/ControlPluginRunner.java
@@ -25,6 +25,7 @@ import javax.annotation.Nullable;
 import javax.inject.Inject;
 
 import org.joda.time.DateTime;
+import org.killbill.billing.ErrorCode;
 import org.killbill.billing.account.api.Account;
 import org.killbill.billing.callcontext.DefaultCallContext;
 import org.killbill.billing.catalog.api.Currency;
@@ -37,6 +38,7 @@ import org.killbill.billing.control.plugin.api.PaymentControlContext;
 import org.killbill.billing.control.plugin.api.PaymentControlPluginApi;
 import org.killbill.billing.control.plugin.api.PriorPaymentControlResult;
 import org.killbill.billing.osgi.api.OSGIServiceRegistration;
+import org.killbill.billing.payment.api.PaymentApiException;
 import org.killbill.billing.payment.api.PluginProperty;
 import org.killbill.billing.payment.api.TransactionType;
 import org.killbill.billing.payment.retry.DefaultFailureCallResult;
@@ -71,7 +73,7 @@ public class ControlPluginRunner {
                                                              final boolean isApiPayment,
                                                              final List<String> paymentControlPluginNames,
                                                              final Iterable<PluginProperty> pluginProperties,
-                                                             final CallContext callContext) throws PaymentControlApiException {
+                                                             final CallContext callContext) throws PaymentControlApiException, PaymentApiException {
         // Return as soon as the first plugin aborts, or the last result for the last plugin
         PriorPaymentControlResult prevResult = new DefaultPriorPaymentControlResult(false, amount, currency, paymentMethodId, pluginProperties);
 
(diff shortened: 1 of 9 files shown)


andrenpaes:
I wonder if it's a good idea to change this class:
https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/bus/PaymentBusEventHandler.java#L105-L107
In order to prevent error logging for `ErrorCode.PAYMENT_PLUGIN_API_ABORTED`.

payment/src/main/java/org/killbill/billing/payment/core/sm/control/OperationControlCallback.java
```diff
@@ -99,16 +100,14 @@ public OperationResult doOperationCallback() throws OperationException {
...
-                        // Transition to ABORTED
-                        return PluginDispatcher.createPluginDispatcherReturnType(OperationResult.EXCEPTION);
-                    }
+                    executePluginPriorCalls(paymentStateControlContext.getPaymentControlPluginNames(), paymentControlContext);
                 } catch (final PaymentControlApiException e) {
                     // Transition to ABORTED and throw PaymentControlApiException to caller.
                     throw new OperationException(e, OperationResult.EXCEPTION);
+                } catch (final PaymentApiException paymentAbortedException) {
+                    // Transition to ABORTED
+                    throw new OperationException(paymentAbortedException, OperationResult.EXCEPTION);
```
sbrossie: I think that works, but i would have preferred to not change the signature of `executePluginPriorCalls` (which now throws a PaymentApiException solely to support that case), and instead just throw the OperationException (with inside the PaymentApiException(ErrorCode. PAYMENT_PLUGIN_API_ABORTED) in the few lines above.

However, i think i understand why you did it that way: Probably you wanted to keep track of which plugin aborted (see remark on the message above).

payment/src/main/java/org/killbill/billing/payment/core/sm/control/ControlPluginRunner.java
```diff
@@ -117,7 +119,7 @@ public PriorPaymentControlResult executePluginPriorCalls(final Account account,
                 inputPluginProperties = prevResult.getAdjustedPluginProperties();
             }
             if (prevResult.isAborted()) {
-                break;
+                throw new PaymentApiException(ErrorCode.PAYMENT_PLUGIN_API_ABORTED, pluginName);
             }
```
sbrossie: Too bad we don't have the reason for the abort. I agree passing the pluginName is good, but maybe format a better message (e.g: "plugin name= 'foo'").

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
```diff
@@ -320,6 +320,23 @@ public void testCreatePurchaseWithControlPaymentPluginException() throws Excepti
...
+        try {
+            paymentApi.createPurchaseWithPaymentControl(
+                    account, account.getPaymentMethodId(), null, requestedAmount, Currency.AED,
+                    paymentExternalKey, transactionExternalKey, ImmutableList.<PluginProperty>of(), CONTROL_PLUGIN_OPTIONS, callContext);
+            fail();
+        } catch (PaymentApiException e) {
+            assertTrue(e.getCause() instanceof PaymentControlApiException);
+        }
+    }
+
```
sbrossie: Good test. I am assuming the goal was *not* to test the aborted state here but the case where plugin throws.

payment/src/main/java/org/killbill/billing/payment/core/sm/control/OperationControlCallback.java
andrenpaes: Yeah, my first implementation was based on what you first described, throwing an `OperationException` wrapping a `PaymentApiException`, but then we'd loose the plugin name that aborted the operation.

I agree that it can get kind of confusing, to only throw this exception for this particular case, but I'm not sure what's the best option.

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
andrenpaes: That's correct.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPayment.java
```diff
@@ -442,6 +443,24 @@ public void testComboAuthorization() throws Exception {
...
+        accountJson.setAccountId(null);
+        final String paymentExternalKey = UUID.randomUUID().toString();
+        final ComboPaymentTransaction comboPaymentTransaction = createComboPaymentTransaction(accountJson, paymentExternalKey);
+
+        mockPaymentControlProviderPlugin.setAborted(true);
+        try {
+            killBillClient.createPayment(comboPaymentTransaction, Arrays.asList(MockPaymentControlProviderPlugin.PLUGIN_NAME), ImmutableMap.<String, String>of(), createdBy, reason, comment);
+            fail();
+        } catch (KillBillClientException e) {
+            assertEquals(e.getResponse().getStatusCode(), 422);
```
sbrossie: Can you use `ErrorCode. PAYMENT_PLUGIN_API_ABORTED.getCode()` instead of numeral value (those sometimes change across major release.

andrenpaes: This is the http response code, not the `ErroCode.getCode`, uhh, code. :)

payment/src/main/java/org/killbill/billing/payment/core/sm/control/ControlPluginRunner.java
andrenpaes: Yeah, for getting the reason, we'd need to change the ControlPlugin API. We could do it, but it seems like a change with a broader scope...

For the error message, should we change here:
https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/ErrorCode.java#L258

Or just use a custom message while constructing the exception instance?

payment/src/main/java/org/killbill/billing/payment/core/sm/control/OperationControlCallback.java
andrenpaes: One option here is to throw create a specialized `PaymentControlApiException` for the abort case.
The code here would change to:
```java
try {
    executePluginPriorCalls(paymentStateControlContext.getPaymentControlPluginNames(), paymentControlContext);
} catch (final PaymentControlApiAbortException e) {
    // Transition to ABORTED
    final PaymentApiException paymentAbortedException = new PaymentApiException(ErrorCode. PAYMENT_PLUGIN_API_ABORTED, e.getPluginName());
    throw new OperationException(paymentAbortedException, OperationResult.EXCEPTION);
} catch (final PaymentControlApiException e) {
    // Transition to ABORTED and throw PaymentControlApiException to caller.
    throw new OperationException(e, OperationResult.EXCEPTION);
}

PR #544, opened 2016-05-17
Update length of field_name in custom_fields and custom_field_history tables

daliwei:
@sbrossie @pierre Please take a look. We have a case (isHPP_for_killbill-paypal-express) which is too long for the field name.


pierre:
Could you provide a migration file too? See http://docs.killbill.io/0.16/database_migrations.html.

daliwei:
@pierre Added the migration file. Please check.

pierre:
Thanks!

PR #550, opened 2016-05-21
Make sure Abort payment exceptions are handled correctly for HPP

### LINKED ISSUE #548: DefaultPaymentGatewayApi doesn't respect isAborted flag
We should make sure 422 is returned if the HPP payment is aborted by a control plugin.

andrenpaes:
Fixes #548


sbrossie:
👍

Issue #221, opened 2014-08-31
Allow deployments as non-root resources


pierre:
Today, Kill Bill only works when deployed as a ROOT resource.

The main culprit is the tenant filter:

https://github.com/killbill/killbill/blob/master/profiles/killbill/src/main/java/org/killbill/billing/server/security/TenantFilter.java#L130

maguero:
Assuming the server is running on http://127.0.0.1:8080/testPath/

``` Java
httpServletRequest.getRequestURI() //  ==>  http://127.0.0.1:8080/testPath/1.0/kb/tenants
// ------------
final String path = httpServletRequest.getRequestURI();  // ==> /testPath/1.0/kb/tenants
final String path = httpServletRequest.getPathInfo();   // ==> /1.0/kb/tenants
```

We can use **httpServletRequest.getPathInfo()** [here](https://github.com/maguero/killbill/blob/master/profiles/killbill/src/main/java/org/killbill/billing/server/security/TenantFilter.java#L138) to get the correct path value.

pierre:
Question: are there changes required in the [web.xml](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/main/webapp/WEB-INF/web.xml) as well?

For example, if we were to run both Kill Bill (under `http://127.0.0.1:8080/killbill`) and [Kaui](https://github.com/killbill/killbill-admin-ui-standalone) (under `http://127.0.0.1:8080/kaui`), we wouldn't want the `ShiroFilter` to apply to `/kaui/*`.

maguero:
I deployed both in tomcat, but now I'm trying to see why kaui is still calling root killbill...

Log:
```
28-Oct-2015 17:22:08.315 INFO [http-nio-8080-exec-3] org.apache.catalina.core.ApplicationContext.log   Parameters: {"utf8"=>"✓", "authenticity_token"=>"pkZZ+TIMD7tY78qMlfWpUlqaxRdQnBDvjyZJEHvXp8I=", "user"=>{"kb_username"=>"admin", "password"=>"[FILTERED]"}, "commit"=>"Sign in"}

28-Oct-2015 17:22:08.337 INFO [http-nio-8080-exec-3] org.apache.catalina.core.ApplicationContext.log ===> GET http://127.0.0.1:8080/1.0/kb/security/permissions

[...]

PR #559, opened 2016-06-13
#221 - Non root deployment issues

javier-gomez-hs:
- Added new system property to allow path-like Location in the Response
- Updated Resources to allow the new property
- Added new method to build the Response taking into account the new property
- Fixed issue of wrong Location value returned when Kill Bill deployed as non-root
- Added Tests to verify this new behavior

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
@@ -333,7 +333,7 @@ public class AccountResource extends JaxRsResourceBase {
 
         final AccountData data = json.toAccountData();
         final Account account = accountUserApi.createAccount(data, context.createContext(createdBy, reason, comment, request));
-        return uriBuilder.buildResponse(uriInfo, AccountResource.class, "getAccount", account.getId());
+        return uriBuilder.buildResponse(uriInfo, AccountResource.class, "getAccount", account.getId(), jaxrsConfig.getJaxrsReturnPathLikeUrl());
     }
 
     @TimedResource
@@ -979,7 +979,7 @@ public class AccountResource extends JaxRsResourceBase {
             default:
                 return Response.status(Status.PRECONDITION_FAILED).entity("TransactionType " + transactionType + " is not allowed for an account").build();
         }
-        return createPaymentResponse(uriInfo, result, transactionType, json.getTransactionExternalKey());
+        return createPaymentResponse(uriInfo, result, transactionType, json.getTransactionExternalKey(), jaxrsConfig.getJaxrsReturnPathLikeUrl());
     }
 
     /*
@@ -1032,8 +1032,8 @@ public class AccountResource extends JaxRsResourceBase {
                                        @HeaderParam(HDR_COMMENT) final String comment,
                                        @javax.ws.rs.core.Context final HttpServletRequest request,
                                        @javax.ws.rs.core.Context final UriInfo uriInfo) throws CustomFieldApiException {
-        return super.createCustomFields(UUID.fromString(id), customFields,
-                                        context.createContext(createdBy, reason, comment, request), uriInfo);
+        return super.createCustomFields(UUID.fromString(id), customFields, context.createContext(createdBy, reason,
+                                             comment, request), uriInfo, jaxrsConfig.getJaxrsReturnPathLikeUrl());
     }
 
     @TimedResource
@@ -1207,7 +1207,7 @@ public class AccountResource extends JaxRsResourceBase {
             accountUserApi.addEmail(accountId, json.toAccountEmail(UUIDs.randomUUID()), callContext);
         }
 
-        return uriBuilder.buildResponse(uriInfo, AccountResource.class, "getEmails", json.getAccountId());
+        return uriBuilder.buildResponse(uriInfo, AccountResource.class, "getEmails", json.getAccountId(), jaxrsConfig.getJaxrsReturnPathLikeUrl());
     }
 
     @TimedResource
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
@@ -333,7 +333,7 @@ public class AccountResource extends JaxRsResourceBase {
 
         final AccountData data = json.toAccountData();
         final Account account = accountUserApi.createAccount(data, context.createContext(createdBy, reason, comment, request));
-        return uriBuilder.buildResponse(uriInfo, AccountResource.class, "getAccount", account.getId(), jaxrsConfig.getJaxrsReturnPathLikeUrl());
+        return uriBuilder.buildResponse(uriInfo, AccountResource.class, "getAccount", account.getId());
     }
 
     @TimedResource
@@ -979,7 +979,7 @@ public class AccountResource extends JaxRsResourceBase {
             default:
                 return Response.status(Status.PRECONDITION_FAILED).entity("TransactionType " + transactionType + " is not allowed for an account").build();
         }
-        return createPaymentResponse(uriInfo, result, transactionType, json.getTransactionExternalKey(), jaxrsConfig.getJaxrsReturnPathLikeUrl());
+        return createPaymentResponse(uriInfo, result, transactionType, json.getTransactionExternalKey());
     }
 
     /*
@@ -1033,7 +1033,7 @@ public class AccountResource extends JaxRsResourceBase {
                                        @javax.ws.rs.core.Context final HttpServletRequest request,
                                        @javax.ws.rs.core.Context final UriInfo uriInfo) throws CustomFieldApiException {
         return super.createCustomFields(UUID.fromString(id), customFields, context.createContext(createdBy, reason,
-                                             comment, request), uriInfo, jaxrsConfig.getJaxrsReturnPathLikeUrl());
+                                             comment, request), uriInfo);
     }
 
     @TimedResource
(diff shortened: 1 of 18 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/util/JaxrsUriBuilder.java
```diff
@@ -74,6 +79,12 @@ public URI getUri() {
         return ri.entity(obj).build();
     }
 
+    private UriBuilder getUriBuilder(final String path, final Class<? extends JaxrsResource> theClassMaybeEnhanced, @Nullable final String getMethodName) {
+        final Class theClass = getNonEnhancedClass(theClassMaybeEnhanced);
+        return getMethodName != null ? UriBuilder.fromPath(path.equals("/") ? path.substring(1) : path).path(theClass).path(theClass, getMethodName) :
+               UriBuilder.fromPath(path).path(theClass);
+    }
+
     private UriBuilder getUriBuilder(final Class<? extends JaxrsResource> theClassMaybeEnhanced, @Nullable final String getMethodName) {
```
sbrossie: Is that one still used?

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestBuildResponse.java
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright 2010-2013 Ning, Inc.
+ * Copyright 2014 Groupon, Inc
+ * Copyright 2014 The Billing Project, LLC
+ *
```
sbrossie: Nit: You don't need to include `Ning` or `Groupon` for any of the new files you create. Also if you care about adding the name of your consulting company this is fine too.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -333,7 +333,7 @@ public Response createAccount(final AccountJson json,
 
         final AccountData data = json.toAccountData();
         final Account account = accountUserApi.createAccount(data, context.createContext(createdBy, reason, comment, request));
-        return uriBuilder.buildResponse(uriInfo, AccountResource.class, "getAccount", account.getId());
+        return uriBuilder.buildResponse(uriInfo, AccountResource.class, "getAccount", account.getId(), jaxrsConfig.getJaxrsReturnPathLikeUrl());
```
sbrossie: Instead of modifying the signature for `JaxrsUriBuilder #buildResponse` and then having to change the callers, would not it be simpler to inject the `JaxrsConfig` into  `JaxrsUriBuilder`. It looks like `JaxrsUriBuilder` is already injected through Guice (e.g in [AccountResource ](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java#L151)), however we miss the eager singleton binding in [DefaultJaxrsModule](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/glue/DefaultJaxrsModule.java#L35).

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestBuildResponse.java
```diff
@@ -0,0 +1,92 @@
...
+        URI uri = URI.create("http://localhost:8080/killbill");
+        when(uriInfo.getBaseUri()).thenReturn(uri);
+        when(uriInfo.getAbsolutePath()).thenReturn(uri);
+
+        JaxrsUriBuilder uriBuilder = new JaxrsUriBuilder();
+        Response response = uriBuilder.buildResponse(uriInfo, AccountResource.class, "getAccount", objectId, false);
+
+        assertEquals(response.getMetadata().get("Location").get(0).toString(), uri.toString() + "/1.0/kb/accounts/" + objectId.toString());
+    }
+}
```
sbrossie: 😎 for the tests!

sbrossie:
@javier-gomez-hs Looks good overall; i have a few comments. I also want @pierre to take a look.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestBuildResponse.java
pierre: Indeed - no need to add previous years. Just `Copyright 2016 The Billing Project, LLC`.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestBuildResponse.java
```diff
@@ -0,0 +1,92 @@
...
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriInfo;
+import java.net.URI;
+import java.util.UUID;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.testng.Assert.assertEquals;
+
+public class TestBuildResponse {
```
pierre: Needs to extend `GuicyKillbillTestSuiteNoDB`.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestBuildResponse.java
```diff
@@ -0,0 +1,92 @@
...
+        when(uriInfo.getBaseUri()).thenReturn(uri);
+        when(uriInfo.getAbsolutePath()).thenReturn(uri);
+
+        JaxrsUriBuilder uriBuilder = new JaxrsUriBuilder();
+        Response response = uriBuilder.buildResponse(uriInfo, AccountResource.class, "getAccount", objectId, false);
+
+        assertEquals(response.getMetadata().get("Location").get(0).toString(), uri.toString() + "/1.0/kb/accounts/" + objectId.toString());
+    }
+
+    @Test(groups = "fast", description = "Tests Uri Builder without Path Like URL and root Location")
```
pierre: Should be `non root Location`?

util/src/main/java/org/killbill/billing/util/config/JaxrsConfig.java
```diff
@@ -34,4 +34,9 @@
     @Description("Total timeout for all callables associated to a given api call (parallel mode)")
     TimeSpan getJaxrsTimeout();
 
+    @Config("org.killbill.jaxrs.return.path.like.url")
+    @Default("false")
+    @Description("Type of return for the jaxrs response location")
+    boolean getJaxrsReturnPathLikeUrl();
```
pierre: Nit: the naming / description is a bit confusing (not obvious it's a boolean). Maybe something more explicit like `isJaxrsLocationFullUrl`?

jaxrs/src/main/java/org/killbill/billing/jaxrs/util/JaxrsUriBuilder.java
javier-gomez-hs: @sbrossie yes, it is still in use by 2 methods [here](https://github.com/javier-gomez-hs/killbill/blob/jgomez-non-root-issues/jaxrs/src/main/java/org/killbill/billing/jaxrs/util/JaxrsUriBuilder.java#L70) and [here](https://github.com/javier-gomez-hs/killbill/blob/jgomez-non-root-issues/jaxrs/src/main/java/org/killbill/billing/jaxrs/util/JaxrsUriBuilder.java#L57).
A refactor might be needed, but will need to have the "path" value.

javier-gomez-hs:
@sbrossie @pierre 
Added fixes!! Thanks for your comments and suggestions!
Please let me know what do you think about it. 
Thanks!

jaxrs/src/main/java/org/killbill/billing/jaxrs/util/JaxrsUriBuilder.java
```diff
@@ -20,27 +20,36 @@
...
+        this.jaxrsConfig = jaxrsConfig;
+    }
+
     public Response buildResponse(final UriInfo uriInfo, final Class<? extends JaxrsResource> theClass,
-                                  final String getMethodName, final Object objectId, final boolean pathLikeUrl) {
-        final URI location = buildLocation(uriInfo, theClass, getMethodName, objectId, pathLikeUrl);
-        return pathLikeUrl ? Response.ok().header("Location", location.getPath()).build() : Response.created(location).build();
+                                  final String getMethodName, final Object objectId) {
+        final URI location = buildLocation(uriInfo, theClass, getMethodName, objectId);
+        return !jaxrsConfig.isJaxrsLocationFullUrl() ? Response.ok().header("Location", location.getPath()).build() : Response.created(location).build();
```
sbrossie: ⚠️  We want to always return '201' status `Status.CREATED`, but  `Response.ok()` will set the status to `Status.OK`. You probably need to set the status by hand calling the `status(Status.CREATED).location(.....)`

sbrossie:
@javier-gomez-hs I have one comment left about status. Looks good otherwise.

javier-gomez-hs:
@sbrossie @pierre 
Thanks for the suggestion! I've added the fix in the code and tests.

beccagaspard:
Hey Guys - just to add some context to my google group posting - we are using [AWS elastic load balancing](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/x-forwarded-headers.html) which sets the `X-Forwarded` headers.  I'm able to change the `Location` header returned by the jaxrs call by setting the `proxyPort`, `proxyName`, `scheme` in the tomcat server.xml, but not sure that will be ideal since we are using docker and we would have to bake in the server names or add extra environment variables or some magic.  Also if we had multiple domain names pointing to the same KillBill instance, tomcat would not be aware and just return the location as whatever is specified in `proxyName`.  I think a [tomcat filter](https://tomcat.apache.org/tomcat-8.0-doc/config/filter.html#Basic_configuration_to_handle_'x-forwarded-for'_and_'x-forwarded-proto') like @andrenpaes referenced should be able to transparently handle the `X-Forwarded` headers, but I think it would require modification of the `web.xml`.  Spring Boot has a [similar functionality](http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/html/howto-embedded-servlet-containers.html#howto-use-tomcat-behind-a-proxy-server) that you can enable or disable - pretty sure it is using servlet filters as well.  I think the use of `X-Forwarded` headers is pretty common with PaaS's like heroku, cloudfoundry, elastic beanstalk, etc. - would be awesome if KillBill could handle them somehow! :)

pierre:
Created https://github.com/killbill/killbill/issues/566 for tracking.

Issue #553, opened 2016-05-25
Revisit the credit story for HA parent


sbrossie:
In the HA (Hierarchical Account) model, when a child delegates its payment to its parent, the parent must first build a summary invoice (one for all its dependent children on any given day). Today this is built by adding a new `PARENT_SUMMARY` invoice item for each child and the amount is the child invoice amount. 

This model therefore includes any credit that child may have on the parent summary invoice , and this probably does not make sense as we don't want to use a given child's credit at the parent level by default.

The proposal is the following:
* Credits are not by default visible at the parent level
* We have a new api which allow to move a given child credit to the parent level (to support the use case when for instance a child cancels his subscription and is left with some 'recognizable credit'.

matias-aguero-hs:
This issue is because the [parent invoice creation event](https://github.com/killbill/killbill/blob/hierarchical-accounts/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L715) is using ``childInvoice.getChargedAmount()`` to calculate parent item amount, and it contemplates the following item types:

```Java
if (isCharge(invoiceItem) ||
    isInvoiceAdjustmentItem(invoiceItem, otherInvoiceItems) ||
    isInvoiceItemAdjustmentItem(invoiceItem) ||
    isParentSummaryItem(invoiceItem)) {
    amountCharged = amountCharged.add(invoiceItem.getAmount());
}
```
I suggest to create a new method in ``InvoiceCalculatorUtils`` (from ``computeInvoiceAmountCharged()``) and return the correct item amounts:

```Java
public static BigDecimal computeChildInvoiceAmountCharged(final Currency currency, @Nullable final Iterable<InvoiceItem> invoiceItems) {
        BigDecimal amountCharged = BigDecimal.ZERO;
        ...
        for (final InvoiceItem invoiceItem : invoiceItems) {
            ...
            if (isCharge(invoiceItem) ||
                isInvoiceItemAdjustmentItem(invoiceItem)) {
                amountCharged = amountCharged.add(invoiceItem.getAmount());
            }
        }
        return KillBillMoney.of(amountCharged, currency);
    }
     // checks removed
     // isInvoiceAdjustmentItem(invoiceItem, otherInvoiceItems)
     // isParentSummaryItem(invoiceItem)
```

matias-aguero-hs:
[...]

PR #561, opened 2016-06-16
Ha credit 553

matias-aguero-hs:
- Integration tests for different credit scenarios.
- Fix on Parent Invoice generation.
- Child Invoice adjustment listener added.
- CBA_ADJ item generation fixed for child/parent invoices.
- Use of credit for child invoices fixed.

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
@@ -1,0 +1,551 @@
+/*
+ * Copyright 2014-2016 Groupon, Inc
+ * Copyright 2014-2016 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.killbill.billing.beatrix.integration;
+
+import java.math.BigDecimal;
+import java.util.List;
+
+import org.joda.time.DateTime;
+import org.killbill.billing.account.api.Account;
+import org.killbill.billing.api.TestApiListener.NextEvent;
+import org.killbill.billing.catalog.api.BillingActionPolicy;
+import org.killbill.billing.catalog.api.BillingPeriod;
+import org.killbill.billing.catalog.api.Currency;
+import org.killbill.billing.catalog.api.ProductCategory;
+import org.killbill.billing.entitlement.api.DefaultEntitlement;
+import org.killbill.billing.entitlement.api.Entitlement;
+import org.killbill.billing.entitlement.api.Entitlement.EntitlementActionPolicy;
+import org.killbill.billing.invoice.api.Invoice;
+import org.killbill.billing.invoice.api.InvoiceItemType;
+import org.killbill.billing.invoice.api.InvoiceStatus;
+import org.killbill.billing.payment.api.Payment;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+public class TestIntegrationParentInvoice extends TestIntegrationBase {
+
+    @Test(groups = "slow")
+    public void testParentInvoiceGeneration() throws Exception {
+
+        final int billingDay = 14;
+        final DateTime initialCreationDate = new DateTime(2015, 5, 15, 0, 0, 0, 0, testTimeZone);
+        // set clock to the initial start date
+        clock.setTime(initialCreationDate);
+
+        log.info("Beginning test with BCD of " + billingDay);
+        final Account parentAccount = createAccountWithNonOsgiPaymentMethod(getAccountData(billingDay));
+        final Account child1Account = createAccountWithNonOsgiPaymentMethod(getChildAccountData(billingDay, parentAccount.getId(), true));
+        final Account child2Account = createAccountWithNonOsgiPaymentMethod(getChildAccountData(billingDay, parentAccount.getId(), true));
+
+        // CREATE SUBSCRIPTIONS AND EXPECT BOTH EVENTS EACH: NextEvent.CREATE NextEvent.INVOICE
+        createBaseEntitlementAndCheckForCompletion(child1Account.getId(), "bundleKey1", "Shotgun", ProductCategory.BASE, BillingPeriod.MONTHLY, NextEvent.CREATE, NextEvent.BLOCK, NextEvent.INVOICE);
+        createBaseEntitlementAndCheckForCompletion(child2Account.getId(), "bundleKey2", "Pistol", ProductCategory.BASE, BillingPeriod.MONTHLY, NextEvent.CREATE, NextEvent.BLOCK, NextEvent.INVOICE);
+
+        // First Parent invoice over TRIAL period
+        List<Invoice> parentInvoices = invoiceUserApi.getInvoicesByAccount(parentAccount.getId(), false, callContext);
+        assertEquals(parentInvoices.size(), 1);
+
+        Invoice parentInvoice = parentInvoices.get(0);
+        assertEquals(parentInvoice.getNumberOfItems(), 2);
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.DRAFT);
+        assertTrue(parentInvoice.isParentInvoice());
+        assertEquals(parentInvoice.getBalance().toString(), "0.00");
+
+        // Moving a day the NotificationQ calls the commitInvoice. No payment is expected
+        busHandler.pushExpectedEvents(NextEvent.INVOICE);
+        clock.addDays(1);
+        assertListenerStatus();
+
+        // reload parent invoice
+        parentInvoice = invoiceUserApi.getInvoice(parentInvoice.getId(), callContext);
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.COMMITTED);
+
+        // Move through time and verify new parent Invoice. No payments are expected yet.
+        busHandler.pushExpectedEvents(NextEvent.PHASE, NextEvent.PHASE,
+                                      NextEvent.INVOICE, NextEvent.INVOICE);
+        clock.addMonths(1);
+        assertListenerStatus();
+
+        // Second Parent invoice over Recurring period
+        parentInvoices = invoiceUserApi.getInvoicesByAccount(parentAccount.getId(), false, callContext);
+        assertEquals(parentInvoices.size(), 2);
+
+        parentInvoice = parentInvoices.get(1);
+        assertEquals(parentInvoice.getNumberOfItems(), 2);
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.DRAFT);
+        assertTrue(parentInvoice.isParentInvoice());
+        assertEquals(parentInvoice.getBalance().toString(), "279.90");
+        assertEquals(parentInvoice.getInvoiceItems().get(0).getInvoiceItemType(), InvoiceItemType.PARENT_SUMMARY);
+        assertEquals(parentInvoice.getInvoiceItems().get(1).getInvoiceItemType(), InvoiceItemType.PARENT_SUMMARY);
+
+        // Check Child Balance. It should be > 0 here because Parent invoice is unpaid yet.
+        List<Invoice> child1Invoices = invoiceUserApi.getInvoicesByAccount(child1Account.getId(), false, callContext);
+        assertEquals(child1Invoices.size(), 2);
+        assertTrue(child1Invoices.get(1).getBalance().compareTo(BigDecimal.ZERO) > 0);
+
+        // Moving a day the NotificationQ calls the commitInvoice. Payment is expected.
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.PAYMENT, NextEvent.INVOICE_PAYMENT);
+        clock.addDays(1);
+        assertListenerStatus();
+
+        parentInvoice = invoiceUserApi.getInvoice(parentInvoice.getId(), callContext);
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.COMMITTED);
+
+        // Check Child Balance. It should be = 0 because parent invoice had already paid.
+        child1Invoices = invoiceUserApi.getInvoicesByAccount(child1Account.getId(), false, callContext);
+        assertEquals(child1Invoices.size(), 2);
+        assertTrue(parentInvoice.getBalance().compareTo(BigDecimal.ZERO) == 0);
+        assertTrue(child1Invoices.get(1).getBalance().compareTo(BigDecimal.ZERO) == 0);
+
+    }
+
+    @Test(groups = "slow")
+    public void testParentInvoiceGenerationMultipleActionsSameDay() throws Exception {
+
+        final int billingDay = 14;
+        final DateTime initialCreationDate = new DateTime(2015, 5, 15, 0, 0, 0, 0, testTimeZone);
+
+        // set clock to the initial start date
+        clock.setTime(initialCreationDate);
+
+
+        log.info("Beginning test with BCD of " + billingDay);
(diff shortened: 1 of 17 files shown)


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -0,0 +1,763 @@
...
+        clock.addDays(29);
+        assertListenerStatus();
+
+        // Moving a day the NotificationQ calls the commitInvoice.
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.PAYMENT, NextEvent.INVOICE_PAYMENT);
+        clock.addDays(1);
+        assertListenerStatus();
+
+        // add credit to child account when invoice is still unpaid
+        busHandler.pushExpectedEvents(NextEvent.INVOICE);
```
sbrossie: I am guessing the comment is wrong here? (probably cur and paste from previous test?) : Should be "add credit to child account after invoice has been paid"

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -0,0 +1,763 @@
...
+        // expected child invoice
+        // RECURRING : $ 249.95
+        // ITEM_ADJ : $ -10
+
+        childInvoice = invoiceUserApi.getInvoice(childInvoice.getId(), callContext);
+        assertEquals(childInvoice.getInvoiceItems().size(), 2);
+        assertEquals(childInvoice.getBalance().compareTo(BigDecimal.valueOf(239.95)), 0);
+        assertEquals(childInvoice.getInvoiceItems().get(0).getInvoiceItemType(), InvoiceItemType.RECURRING);
+        assertEquals(childInvoice.getInvoiceItems().get(1).getInvoiceItemType(), InvoiceItemType.ITEM_ADJ);
+
```
sbrossie: I would have expected the parent `SUMMARY` invoice item to be modified (because invoice was not yet committed). I also don't think this is very important. @pierre Any thoughts?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -0,0 +1,763 @@
...
+        assertEquals(childInvoice.getInvoiceItems().get(2).getAmount().compareTo(BigDecimal.valueOf(10)), 0);
+
+        // check parent invoices
+        parentInvoice = invoiceUserApi.getInvoice(parentInvoice.getId(), callContext);
+        assertEquals(parentInvoice.getInvoiceItems().size(), 3);
+        assertEquals(parentInvoice.getPaidAmount().compareTo(BigDecimal.valueOf(249.95)), 0);
+        assertEquals(parentInvoice.getInvoiceItems().get(0).getInvoiceItemType(), InvoiceItemType.PARENT_SUMMARY);
+        assertEquals(parentInvoice.getInvoiceItems().get(1).getInvoiceItemType(), InvoiceItemType.ITEM_ADJ);
+        assertEquals(parentInvoice.getInvoiceItems().get(2).getInvoiceItemType(), InvoiceItemType.CBA_ADJ);
+
```
sbrossie: I think this is incorrect: This is the scenario 2 from #553. Nothing should be seen on the parent invoice.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -0,0 +1,763 @@
...
+        // expected invoices
+
+        // invoice 2:
+        // REPAIR_ADJ $ -15
+        // CBA_ADJ $ 15
+
+        // Invoice 1:
+        // RECURRING : $ 20
+        // CBA_ADJ $ -15
+
```
sbrossie: Confused with these comments. They don't seem to match the amount from the test?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -0,0 +1,763 @@
...
+        // check if parent invoice was updated
+        parentInvoices = invoiceUserApi.getInvoicesByAccount(parentAccount.getId(), false, callContext);
+        assertEquals(parentInvoices.size(), 2);
+
+        parentInvoice = parentInvoices.get(1);
+        assertEquals(parentInvoice.getNumberOfItems(), 1);
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.DRAFT);
+        assertTrue(parentInvoice.isParentInvoice());
+        assertEquals(parentInvoice.getBalance().compareTo(BigDecimal.valueOf(16.66)), 0);
+
```
sbrossie: Can we also check the items on the parent invoice? Do we see one or two (was `SUMMARY` 's amount modified or dod we add an `ITEM_ADJ`?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -0,0 +1,763 @@
...
+        // expected child invoices
+
+        // Invoice 2:
+        // REPAIR_ADJ $ -15
+        // CBA_ADJ $ 15
+
+        // Invoice 1: # unchanged
+        // RECURRING : $ 20
+
```
sbrossie: Same remark with comments/amount as above.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -0,0 +1,763 @@
...
```
sbrossie: Same remark on comments.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -0,0 +1,763 @@
...
+        parentInvoices = invoiceUserApi.getInvoicesByAccount(parentInvoice.getId(), false, callContext);
+        assertEquals(parentInvoices.size(), 4);
+
+        parentInvoice = parentInvoices.get(3);
+        assertEquals(parentInvoice.getNumberOfItems(), 1);
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.DRAFT);
+        assertTrue(parentInvoice.isParentInvoice());
+        assertEquals(parentInvoice.getBalance().compareTo(BigDecimal.valueOf(8.33)), 0);
+
+    }
```
sbrossie: Nit: You could have simplified the test by just starting with a account credit on the child and only start from line 734 (the rest was already tested previously), but this is fine to keep it the way you have it.

sbrossie:
@maguero I think there are a few issues with the tests (mostly thinking on expected behavior, not minor comments). When tests show correct behavior, i'll review the code diff itself.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
pierre: Indeed - if the invoice is still in draft mode, I would have expected the original item to be updated (to simplify the overall committed summary invoice). Agreed it's not super important at this stage though.

matias-aguero-hs: Parent invoice is updated. You can check the [assert](https://github.com/matias-aguero-hs/killbill/blob/d31aab07c5d2caff2145aad758226073249b0795/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java#L383-L383) method a few lines below.

```Java
// reload parent invoice
parentInvoice = invoiceUserApi.getInvoice(parentInvoice.getId(), callContext);
...
assertEquals(parentInvoice.getBalance().compareTo(BigDecimal.valueOf(239.95)), 0);
```

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
matias-aguero-hs: The implemented behaviour is what you suggested to do when parent invoice is already in COMMITTED status.

https://github.com/killbill/killbill/issues/553#issuecomment-225322061

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
matias-aguero-hs: You are right, but I wanted to ensure that new implementation (child-parent relationship) works for next invoices.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
sbrossie: @maguero Yes you are right. Ignore the comment.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
sbrossie: I don't think so. Do we agree this test case represents the scenario 2 is described in [the comment for issue 553](https://github.com/killbill/killbill/issues/553#issuecomment-224425217) ? Because invoice was already paid, the item adjustment  results into a child credit (`CBA_ADJ `), but nothing should be reflected on the parent `SUMMARY` (by default we don't want to transfer the credit to the parent). Also, in the current version, we would end up with 2 credits (one on the child and one of the parent which from an accounting point of view would be wrong).

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
sbrossie: No problem. This is fine to keep it the way you have it.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
matias-aguero-hs: > Do we agree this test case represents the scenario 2 is described in the comment for issue 553 ? 

Yes.

I have rolled back this change and pushed into repo.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -763,4 +765,102 @@ public void testParentInvoiceEarlyCancellationUseCredit() throws Exception {
...
+        assertEquals(parentInvoices.size(), 1);
+
+        final Invoice parentInvoice = parentInvoices.get(0);
+        assertEquals(parentInvoice.getNumberOfItems(), 3);
+        assertEquals(parentInvoice.getInvoiceItems().get(0).getInvoiceItemType(), InvoiceItemType.CREDIT_ADJ);
+        assertEquals(parentInvoice.getInvoiceItems().get(0).getAmount().compareTo(BigDecimal.valueOf(-250)), 0);
+        assertEquals(parentInvoice.getInvoiceItems().get(1).getInvoiceItemType(), InvoiceItemType.CBA_ADJ);
+        assertEquals(parentInvoice.getInvoiceItems().get(1).getAmount().compareTo(BigDecimal.valueOf(250)), 0);
+        assertEquals(parentInvoice.getInvoiceItems().get(2).getInvoiceItemType(), InvoiceItemType.PARENT_SUMMARY);
+        assertEquals(parentInvoice.getInvoiceItems().get(2).getAmount().compareTo(BigDecimal.ZERO), 0);
```
sbrossie: Why do we have also a $0 `PARENT_SUMMARY ` ?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -763,4 +765,102 @@ public void testParentInvoiceEarlyCancellationUseCredit() throws Exception {
...
+        final Account childAccount = createAccountWithNonOsgiPaymentMethod(getChildAccountData(billingDay, parentAccount.getId(), true));
+
+        BigDecimal childAccountCBA = invoiceUserApi.getAccountCBA(childAccount.getId(), callContext);
+        assertEquals(childAccountCBA.compareTo(BigDecimal.ZERO), 0);
+
+        BigDecimal parentAccountCBA = invoiceUserApi.getAccountCBA(parentAccount.getId(), callContext);
+        assertEquals(parentAccountCBA.compareTo(BigDecimal.ZERO), 0);
+
+        invoiceUserApi.transferChildCreditToParent(childAccount.getId(), callContext);
+
```
sbrossie: Don't we want to check in that case no invoices were generated (both child and parent) ?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -763,4 +765,102 @@ public void testParentInvoiceEarlyCancellationUseCredit() throws Exception {
...
+        // set clock to the initial start date
+        clock.setTime(initialCreationDate);
+
+        final Account account = createAccountWithNonOsgiPaymentMethod(getChildAccountData(billingDay, null, true));
+
+        BigDecimal childAccountCBA = invoiceUserApi.getAccountCBA(account.getId(), callContext);
+        assertEquals(childAccountCBA.compareTo(BigDecimal.ZERO), 0);
+
+        invoiceUserApi.transferChildCreditToParent(account.getId(), callContext);
+
```
sbrossie: Same remark here: Don't we want to check in that case no invoices were generated (both child and parent) ?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
sbrossie: Forget the above remark, i did not see you were expecting the `InvoiceApiException`...

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
sbrossie: Forget the above remark, i did not see you were expecting the `InvoiceApiException`...

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -1100,4 +1104,73 @@ public Void inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFa
             }
         });
     }
+
+    @Override
+    public void transferChildCreditToParent(final Account childAccount, final CallContext context) throws InvoiceApiException {
+        transactionalSqlDao.execute(new EntitySqlDaoTransactionWrapper<Void>() {
```
sbrossie: ⚠️ All the `InvoiceDao` apis use either  `InternalTenantContext` (read-only) or `InternalCallContext` (read-write), so the second argument `CallContext ` should really be in  `InternalCallContext`. Computation of such a context will require making a choice about whether we use the parent `accountId` or child `accountId`. I would suggest child unless t6here is a good reason not to (and inside the method you might need to create another one for the parent). This should be clearly documented inside the  `InvoiceDao` api.

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -1100,4 +1104,73 @@ public Void inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFa
...
+                                                                                 chargeDescription,
+                                                                                 effectiveDate.toLocalDate(),
+                                                                                 accountCBA,
+                                                                                 childAccount.getCurrency());
+                invoiceForExternalCharge.addInvoiceItem(externalChargeItem);
+
+                // create credit to parent account
+                final Invoice invoiceForCredit = new DefaultInvoice(childAccount.getParentAccountId(), effectiveDate.toLocalDate(), effectiveDate.toLocalDate(),
+                                                                    childAccount.getCurrency(), InvoiceStatus.DRAFT);
+                final String creditDescription = "Credit migrated from child account " + childAccount.getId();
```
sbrossie: Does it mean the parent invoice will end up in `DRAFT`? I would expect that parent invoice to be `COMMITTED`, no?

sbrossie:
@maguero Looks good overall (i had a few comments), but i believe after those are addressed we should be ready to merge.

Issue #163, opened 2014-03-04
Support for account with no payment methods.


sbrossie:
Today deleting a default payment method on an account is not allowed unless we specified to also switch the account into AUTO_PAY_OFF. The goal was to avoid having to make payment calls on accounts that have no payment method-- code would just spit a stack trace because it does not expect to see that.

javier-gomez-hs:
@sbrossie @pierre 
Hi guys!
In the meantime until you can answer my comment in https://github.com/killbill/killbill/issues/69, I've started working on this issue.

I was able to "allow" the removal of the Default Payment Method, and now I found where is the Exception being thrown when the Payment Method Id is null during the Payment process.
My question is: should I register the Payment Attempt in the DB / Timeline? Or just return an Exception with a new Payment Failure Status (like "NO_PAYMENT_METHOD" or something like that).
The last step before applying the Payment Transaction is [this one](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentLeavingStateCallback.java#L62)

Thanks!!

Javier.

sbrossie:
@javier-gomez-hs The short story is that we want to make sure that when there is payment operation running against a given invoice, and there is no default payment method on the account, the payment gets aborted. This should happen in the [pseudo invoice plugin](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java#L132).

javier-gomez-hs:
@sbrossie @pierre 
I've pushed some initial changes [here](https://github.com/killbill/killbill/pull/571).
Would you like to take a look?
That way the payment is being marked as "ABORTED" and the Server log is not showing any exception.

Is this the expected behavior?

Please let me know if I have to change anything else.

Thanks!

pierre:
> Would you like to take a look?

I've just posted some comments after sync'ing with @sbrossie on this feature.

> That way the payment is being marked as "ABORTED" and the Server log is not showing any exception.

Yes, it should be obvious though why the invoice wasn't paid (`log.warn` but a stacktrace would be useless though).

javier-gomez-hs:
@sbrossie @pierre 
[...]

PR #564, opened 2016-06-28
#69

javier-gomez-hs:
- Added new "withAttempts" parameter to the getPayments method in the Resource
- If it is true, added changes in the PaymentProcessor to get and handle the next Scheduled Payment Attempt
- Added changes to tests, Json and Resource files that includes this new attribute.

PR #571, opened 2016-07-05
Support for Account with no payment methods. #163

javier-gomez-hs:
- Removed control of null PaymentMethod in the PaymentProcessor
- Added control of null PaymentMethod in the PriorCall method and ABORTING payment in that case

### DIFF
--- a/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
+++ b/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
@@ -410,11 +410,6 @@ public class PaymentProcessor extends ProcessorBase {
             currentStateName = paymentModelDao.getLastSuccessStateName();
         }
 
-        // Sanity: no paymentMethodId was passed through API and account does not have a default paymentMethodId
-        if (paymentStateContext.getPaymentMethodId() == null) {
-            throw new PaymentApiException(ErrorCode.PAYMENT_NO_DEFAULT_PAYMENT_METHOD, paymentStateContext.getAccount().getId());
-        }
-
         final UUID nonNullPaymentId = paymentAutomatonRunner.run(paymentStateContext, daoHelper, currentStateName, transactionType);
 
         return getPayment(nonNullPaymentId, true, properties, callContext, internalCallContext);
--- a/payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
+++ b/payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
@@ -129,9 +129,14 @@ public final class InvoicePaymentControlPluginApi implements PaymentControlPlugi
         final InternalCallContext internalContext = internalCallContextFactory.createInternalCallContext(paymentControlContext.getAccountId(), paymentControlContext);
         switch (transactionType) {
             case PURCHASE:
-                // if the Payment Method is still null, abort payment
-                return (paymentControlContext.getPaymentMethodId() != null) ? getPluginPurchaseResult(paymentControlContext, pluginProperties, internalContext) :
-                        new DefaultPriorPaymentControlResult(true);
+                if (paymentControlContext.getPaymentMethodId() != null) {
+                    return getPluginPurchaseResult(paymentControlContext, pluginProperties, internalContext);
+                }
+                else {
+                    // if the Payment Method is still null, abort payment
+                    log.warn("Payment for Invoice {} was not triggered - Payment Method is null", getInvoiceId(pluginProperties));
+                    return new DefaultPriorPaymentControlResult(true);
+                }
             case REFUND:
                 return getPluginRefundResult(paymentControlContext, pluginProperties, internalContext);
             case CHARGEBACK:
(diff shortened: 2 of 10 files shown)


payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
```diff
@@ -129,7 +129,9 @@ public PriorPaymentControlResult priorCall(final PaymentControlContext paymentCo
         final InternalCallContext internalContext = internalCallContextFactory.createInternalCallContext(paymentControlContext.getAccountId(), paymentControlContext);
         switch (transactionType) {
             case PURCHASE:
-                return getPluginPurchaseResult(paymentControlContext, pluginProperties, internalContext);
+                // if the Payment Method is still null, abort payment
+                return (paymentControlContext.getPaymentMethodId() != null) ? getPluginPurchaseResult(paymentControlContext, pluginProperties, internalContext) :
```
pierre: @sbrossie Wouldn't we be automatically in `AUTO_PAY_OFF` today (unless maybe the account never had a payment method)?

payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
```diff
@@ -410,11 +410,6 @@ public boolean apply(final PaymentTransactionModelDao input) {
             currentStateName = paymentModelDao.getLastSuccessStateName();
         }
 
-        // Sanity: no paymentMethodId was passed through API and account does not have a default paymentMethodId
-        if (paymentStateContext.getPaymentMethodId() == null) {
-            throw new PaymentApiException(ErrorCode.PAYMENT_NO_DEFAULT_PAYMENT_METHOD, paymentStateContext.getAccount().getId());
```
pierre: After re-reading the code, it looks like it's safe indeed to remove this check (cannot happen).

However, the issue of trying to pay an invoice (or triggering a standalone payment) without a default payment method can happen, and we should return that specific `ErrorCode` in this case:

* In `JaxRsResourceBase#validatePaymentMethodForAccount`, instead of `verifyNonNull`, can you throw this `ErrorCode`?
* In `DefaultPaymentApi#createPurchaseWithPaymentControl`, instead of `PAYMENT_INVALID_PARAMETER`, we should use that `ErrorCode` instead

payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
pierre: > Wouldn't we be automatically in `AUTO_PAY_OFF` today (unless maybe the account never had a payment method)?

It's indeed the only use-case today.

@javier-gomez-hs Can you add a `log.warn` though, to make sure administrators understand from the logs why the payment wasn't triggered (maybe print the `invoiceId` too)?

pierre:
Additionally, `DefaultPaymentApi#deletePaymentMethod` should take an extra argument `boolean forceDefaultPaymentMethodDeletion` which should apply to `PaymentMethodProcessor#deletedPaymentMethod` (if `true`, `PAYMENT_DEL_DEFAULT_PAYMENT_METHOD` should not be thrown).

JAX-RS APIs need to be updated accordingly (default value is `false`).

payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
javier-gomez-hs: @pierre @sbrossie 

> In JaxRsResourceBase#validatePaymentMethodForAccount, instead of verifyNonNull, can you throw this ErrorCode?

I'll replace the **verifyNonNull** method throwing the **PaymentApiException** using the **PAYMENT_NO_DEFAULT_PAYMENT_METHOD** Error Code. Is that what we are expecting?

payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
javier-gomez-hs: @pierre 

Sure, I'll add the **log.warn** line.
Unfortunately the `invoiceId` is not present in the [PaymentControlContext](https://github.com/javier-gomez-hs/killbill/blob/13fd4001d59e59868186e2a306146d572be750dc/payment/src/main/java/org/killbill/billing/payment/core/sm/control/ControlPluginRunner.java#L260-L274).
Maybe I can use the **paymentId** instead and reply something like this:
 ```
log.warn("Payment was not triggered - Payment Method is null for PaymentId={}", paymentControlContext.getPaymentId());
```

Please let me know your thoughts.

Thanks!

Comment (pierre) [inline payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java]: The invoice id should be present in the plugin properties.

Try:

```
final UUID invoiceId = getInvoiceId(pluginProperties);
```

payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
pierre: Yes. The Java client should be able to catch it (similar to https://github.com/killbill/killbill/blob/master/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestAccount.java#L100-L102).

javier-gomez-hs:
@pierre @sbrossie 
I've added the requested changes.
As I commented here: https://github.com/killbill/killbill/issues/163#issuecomment-230876539 there is still some Integration tests that are not working because of the Overdue States.

Thanks!

pierre:
> I've added the requested changes

Could you also add new server tests for the new query parameter?

Thanks!

sbrossie:
@javier-gomez-hs Same remark as [here](https://github.com/killbill/killbill/pull/564#issuecomment-232441379). Please resubmit your PR against `work-for-release-0.17.x`, and then i'll close this one.

sbrossie:
Closed as work as been ported in PR #576

PR #574, opened 2016-07-11
Missing adjuntment for unpaid COMMITTED parent invoice

matias-aguero-hs:


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -777,41 +780,56 @@ private boolean shouldIgnoreChildInvoice(final Invoice childInvoice, final BigDe
...
         }
 
+        final Long parentAccountRecordId = internalCallContextFactory.getRecordIdFromObject(account.getParentAccountId(), ObjectType.ACCOUNT, buildTenantContext(context));
+        final InternalCallContext parentContext = internalCallContextFactory.createInternalCallContext(parentAccountRecordId, context);
+        final String description = "Adjustment for account ".concat(account.getExternalKey());
+
         final InvoiceItemModelDao childInvoiceItemAdjustment = Iterables.find(childInvoiceModelDao.getInvoiceItems(), new Predicate<InvoiceItemModelDao>() {
             @Override
             public boolean apply(@Nullable final InvoiceItemModelDao input) {
                 return input.getType().equals(InvoiceItemType.ITEM_ADJ) || input.getType().equals(InvoiceItemType.REPAIR_ADJ);
```
sbrossie: I am confused on why we look for either `ITEM_ADJ` OR `REPAIR_ADJ`. Shouldn't we **only** look for `ITEM_ADJ`, especially considering that we use the `childInvoiceItemAdjustment` to retrieve the amount (based on the item selected it would end up either positive or negative).

Also, while re-reading this code, i think we might have an issue, because in the case where we have multiple 
`ITEM_ADJ`, how do we know that the `childInvoiceItemAdjustment` is for the correct `ITEM_ADJ` ?

sbrossie: Thinking a bit more about it, we should probably pick the last created `ITEM_ADJ` on the child invoice. This is not a perfect solution as we could have scenarios where we have two concurrent `ITEM_ADJ` on the same child invoice.

A slightly better option would be to also also on the parent invoice to understand we we already had previous `ITEM_ADJ` and understand which one is missing. Code would become a bit more complex to write ( it think this is ok to implement the easier path for now as this is an edge case).

You should definitely enhance your previous test `testParentInvoiceItemAdjustmentUnpaidCommittedInvoice` to trigger several adjustments and make sure all of them are correctly accounted for.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -385,6 +385,89 @@ public void testParentInvoiceItemAdjustmentUnpaidInvoice() throws Exception {
...
+        // check parent invoice is updated and still in DRAFT status
+        assertEquals(parentInvoice.getNumberOfItems(), 2);
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.COMMITTED);
+        assertTrue(parentInvoice.isParentInvoice());
+        assertEquals(parentInvoice.getBalance().compareTo(BigDecimal.valueOf(239.95)), 0);
+        assertEquals(parentInvoice.getInvoiceItems().get(0).getInvoiceItemType(), InvoiceItemType.PARENT_SUMMARY);
+        assertEquals(parentInvoice.getInvoiceItems().get(1).getInvoiceItemType(), InvoiceItemType.ITEM_ADJ);
+
+    }
+
```
sbrossie: 👍 on the new test.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -385,6 +385,131 @@ public void testParentInvoiceItemAdjustmentUnpaidInvoice() throws Exception {
...
+        assertEquals(parentInvoice.getNumberOfItems(), 1);
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.COMMITTED);
+        assertTrue(parentInvoice.isParentInvoice());
+        assertEquals(parentInvoice.getBalance().compareTo(BigDecimal.valueOf(249.95)), 0);
+
+        // issue a $10 adj when invoice is unpaid
+        insertInvoiceItemAdjustmentToChildInvoice(childAccount, childInvoice, BigDecimal.TEN);
+        // make sure there is time difference between item adjustments.
+        // Otherwise they are created with same id and createdDate and it's used to sort them.
+        Thread.sleep(1000);
```
sbrossie: I would suggest to move the virtual clock by one sec instead of sleeping, `clock#addDeltaFromReality(1000)`, that way the test run faster (this makes a notable difference considering all the tests we have).

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -777,41 +782,67 @@ private boolean shouldIgnoreChildInvoice(final Invoice childInvoice, final BigDe
...
         });
 
+        // find last ITEM_ADJ invoice added in child invoice
+        Comparator<InvoiceItemModelDao> cmp = new Comparator<InvoiceItemModelDao>() {
+            @Override
+            public int compare(InvoiceItemModelDao o1, InvoiceItemModelDao o2) {
+                return o1.getCreatedDate().compareTo(o2.getCreatedDate());
+            }
+        };
+        final InvoiceItemModelDao lastChildInvoiceItemAdjustment = Collections.max(Lists.newArrayList(childAdjustments), cmp);
```
sbrossie: Nit: You could inline the comparator, this is more in sync with what we usually do:
```
Collections.max(Lists.newArrayList(childAdjustments), new Comparator() {
/// ...
});
```

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -777,41 +782,67 @@ private boolean shouldIgnoreChildInvoice(final Invoice childInvoice, final BigDe
...
+
+        final BigDecimal childInvoiceAdjustmentAmount = lastChildInvoiceItemAdjustment.getAmount();
+
+        if (parentInvoiceModelDao.getStatus().equals(InvoiceStatus.COMMITTED)) {
+            if (InvoiceModelDaoHelper.getBalance(parentInvoiceModelDao).compareTo(BigDecimal.ZERO) > 0) {
```
sbrossie: Nit: I think we could check for that earlier and avoid any previous processing. By the way the check 
```
if (InvoiceModelDaoHelper.getBalance(parentInvoiceModelDao).compareTo(BigDecimal.ZERO) == 0) {
   return;
}
```
would be true regardless of `InvoiceStatus` (because when this is not `COMMITTED` this can't be paid and therefore condition would not be hit.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationParentInvoice.java
```diff
@@ -385,6 +385,131 @@ public void testParentInvoiceItemAdjustmentUnpaidInvoice() throws Exception {
...
+        assertEquals(parentInvoice.getStatus(), InvoiceStatus.COMMITTED);
+        assertTrue(parentInvoice.isParentInvoice());
+        assertEquals(parentInvoice.getBalance().compareTo(BigDecimal.valueOf(209.95)), 0);
+        assertEquals(parentInvoice.getInvoiceItems().get(0).getInvoiceItemType(), InvoiceItemType.PARENT_SUMMARY);
+        assertEquals(parentInvoice.getInvoiceItems().get(1).getInvoiceItemType(), InvoiceItemType.ITEM_ADJ);
+        assertEquals(parentInvoice.getInvoiceItems().get(2).getInvoiceItemType(), InvoiceItemType.ITEM_ADJ);
+        assertEquals(parentInvoice.getInvoiceItems().get(3).getInvoiceItemType(), InvoiceItemType.ITEM_ADJ);
+        assertEquals(parentInvoice.getInvoiceItems().get(4).getInvoiceItemType(), InvoiceItemType.ITEM_ADJ);
+        assertEquals(parentInvoice.getInvoiceItems().get(5).getInvoiceItemType(), InvoiceItemType.ITEM_ADJ);
+
```
sbrossie: For each of the `parentInvoice#getInvoiceItems`, could you assert on the `amount` and also the `linkedInvoiceItem` ?

sbrossie:
👍 Looks good (I have one comment i think we should do for enhancing test assertion and some minor comments that you could implement if you want)

sbrossie:
@maguero Will merge shortly

Issue #573, opened 2016-07-09
"Invalid payment transition" Exception when calling notifyPendingTransactionOfStateChanged


andrenpaes:
While handling a notification request from Adyen, the killbill-adyen-plugin first stores the received data in the `adyen_response` table, including the result of the transaction (e.g PROCESSED).

Afterwards, the plugin calls the KillBill API to transition the payment into the expected status (`notifyPendingTransactionOfStateChanged` method). But that call fails with the following exception:
```
org.killbill.billing.payment.api.PaymentApiException: Invalid payment transition VOID from state VOID_SUCCESS
	at org.killbill.billing.payment.core.sm.PaymentAutomatonRunner.runStateMachineOperation(PaymentAutomatonRunner.java:239)
	at org.killbill.billing.payment.core.sm.PaymentAutomatonRunner.run(PaymentAutomatonRunner.java:208)
	at org.killbill.billing.payment.core.PaymentProcessor.performOperation(PaymentProcessor.java:418)
	at org.killbill.billing.payment.core.PaymentProcessor.notifyPendingPaymentOfStateChanged(PaymentProcessor.java:164)
	at org.killbill.billing.payment.api.DefaultPaymentApi.notifyPendingTransactionOfStateChanged(DefaultPaymentApi.java:586)
	at org.killbill.billing.util.glue.KillbillApiAopModule$ProfilingMethodInterceptor$1.execute(KillbillApiAopModule.java:52)
	at org.killbill.commons.profiling.Profiling.executeWithProfiling(Profiling.java:33)
	at org.killbill.billing.util.glue.KillbillApiAopModule$ProfilingMethodInterceptor.invoke(KillbillApiAopModule.java:49)
	at org.killbill.billing.plugin.adyen.core.KillbillAdyenNotificationHandler.transitionPendingTransaction(KillbillAdyenNotificationHandler.java:289)
	at org.killbill.billing.plugin.adyen.core.KillbillAdyenNotificationHandler.updateKillbill(KillbillAdyenNotificationHandler.java:227)
	at org.killbill.billing.plugin.adyen.core.KillbillAdyenNotificationHandler.updateKillbill(KillbillAdyenNotificationHandler.java:189)
	at org.killbill.billing.plugin.adyen.core.KillbillAdyenNotificationHandler.handleNotification(KillbillAdyenNotificationHandler.java:134)
[...]

PR #575, opened 2016-07-13
Add flag for conditionally invoking the Janitor

andrenpaes:
Fix for #573

Fixes the notifyPendingPaymentOfStateChanged method for the scenario where plugin status was updated before the method is called.


pierre:
👍

sbrossie:
👍

Issue #69, opened 2013-09-06
Display payment retry in billing timeline


sbrossie:

pierre:
See also discussion here: https://github.com/killbill/killbill/issues/393.

sbrossie:
True but #393 has been closed so this bug tracks the specific piece about exporting the payment retries. I think this is a nice feature to have.

pierre:
:+1: (I just wanted a link to your description of how retries work)

javier-gomez-hs:
@sbrossie @pierre 
I've been testing this issue, and as far as I saw, all the payment retries are being added to the **Account Timeline** with the status: "PAYMENT_FAILURE":
![image](https://cloud.githubusercontent.com/assets/16122005/16128355/cc422e8a-33d6-11e6-8695-1ccd9ce08167.png)

(I have added the **payment-test** plugin in order to simulate the payment failures)

Is this the expected behavior? 
When you say "Billing Timeline" you meant the "Account Timeline"?

> this bug tracks the specific piece about exporting the payment retries. I think this is a nice feature to have

I'm not sure what do you mean with exporting the payment retries.

Thanks!

sbrossie:
@javier-gomez-hs I would suggest to hold on on this task. We have a task on our end to revisit our story about payment retries and i think this will overlap with that work. I'll sync up with @pierre in the near future, to see if there is anything you could do on your end.

pierre:
Regarding this task, I think the original bug was to display the failed retries in the timeline. As @javier-gomez-hs pointed out, this has been fixed.

What's missing though (originally part of https://github.com/killbill/killbill/issues/393) is seeing the scheduled retry on the invoice (or payment) page.

Something like this would be nice:

![image](https://cloud.githubusercontent.com/assets/51940/16312960/db648d40-392b-11e6-91ab-09d6906eb5c6.png)

If there is no scheduled retry, this would implicitly mean that we gave up retrying.

To implement this though, we need a new API exposing the attempts. [Thinking out loud] Maybe something like `GET /1.0/kb/payments/<paymentId>/attempts?includeFuture=false` (and variations with keys like  `GET /1.0/kb/payments/attempts?paymentExternalKey=XXX&includeFuture=false`). When `includeFuture` is true, the code would need to search the `payment-service` queue for scheduled retries.

javier-gomez-hs:
Yes @pierre, it does make sense!
I'll work on that and let you know when it's done.

Thanks for your help!

[...]

PR #576, opened 2016-07-14
Changes for Issues #163 and #69

javier-gomez-hs:


payment/src/main/java/org/killbill/billing/payment/core/PaymentMethodProcessor.java
```diff
@@ -414,7 +415,7 @@ public void deletedPaymentMethod(final Account account, final UUID paymentMethod
                     try {
                         // Note: account.getPaymentMethodId() may be null
                         if (paymentMethodId.equals(account.getPaymentMethodId())) {
-                            if (!deleteDefaultPaymentMethodWithAutoPayOff) {
+                            if (!deleteDefaultPaymentMethodWithAutoPayOff && !forceDefaultPaymentMethodDeletion) {
                                 throw new PaymentApiException(ErrorCode.PAYMENT_DEL_DEFAULT_PAYMENT_METHOD, account.getId());
                             } else {
                                 final boolean isAccountAutoPayOff = isAccountAutoPayOff(account.getId(), context);
```
sbrossie: I think we are missing a `if` condition here (below the `else`):

```
} else {
    if (deleteDefaultPaymentMethodWithAutoPayOff && !isAccountAutoPayOff(account.getId(), context)) {
        log.info("Setting AUTO_PAY_OFF on accountId='{}' because of default payment method deletion", account.getId());
         setAccountAutoPayOff(account.getId(), context);
    }
    accountInternalApi.removePaymentMethod(account.getId(), context);
}
```

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
```diff
@@ -147,12 +147,43 @@ public void testAddRemovePaymentMethod() throws Exception {
...
+        final UUID paymentMethodId = account.getPaymentMethodId();
+
+        try {
+            paymentApi.deletePaymentMethod(account, paymentMethodId, false, false, ImmutableList.<PluginProperty>of(), callContext);
+        } catch (final PaymentApiException e) {
+            Assert.assertEquals(e.getCode(), ErrorCode.PAYMENT_INTERNAL_ERROR.getCode());
+        }
+        checkPaymentMethodPagination(paymentMethodId, baseNbRecords + 1, false);
+    }
+
```
sbrossie: Do we already have a test to verify the case ``deletePaymentMethod(account, paymentMethodId, true, false,...```  ? Might be a good addition if this is not the case (and verify that indeed we end up in auto_pay_off

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
```diff
@@ -147,12 +147,43 @@ public void testAddRemovePaymentMethod() throws Exception {
...
+        final Long baseNbRecords = paymentApi.getPaymentMethods(0L, 1000L, false, ImmutableList.<PluginProperty>of(), callContext).getMaxNbRecords();
+        Assert.assertEquals(baseNbRecords, (Long) 1L);
+
+        final Account account = testHelper.createTestAccount(UUID.randomUUID().toString(), true);
+        final UUID paymentMethodId = account.getPaymentMethodId();
+
+        checkPaymentMethodPagination(paymentMethodId, baseNbRecords + 1, false);
+
+        paymentApi.deletePaymentMethod(account, paymentMethodId, false, true, ImmutableList.<PluginProperty>of(), callContext);
+
```
sbrossie: Can we add a check to verify that account **does not** end up in AUTO_PAY_OFF (considering my remark earlier, i am pretty sure with current code it would fail).

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPayment.java
```diff
@@ -148,6 +153,111 @@ public void testWithTimeoutPayment() throws Exception {
...
+
+        Payments payments = killBillClient.searchPayments("", 0L, 100L, AuditLevel.NONE, inputOptions);
+        Assert.assertNotNull(payments.get(0));
+        Payment payment = killBillClient.getPayment(payments.get(0).getPaymentId(), inputOptions);
+
+        Assert.assertNotNull(payment.getPaymentAttempts());
+        Assert.assertEquals(payment.getPaymentAttempts().get(0).getStateName(), "RETRIED");
+        Assert.assertEquals(payment.getPaymentAttempts().get(1).getStateName(), "SCHEDULED");
+    }
+
```
sbrossie: 👍 for the tests!

sbrossie:
@javier-gomez-hs I wished you had kept the two issues on two separate PRs, but this is fine (i don't suggest you do anything else, we will merge it this way).

I have commented about one issue that needs to be addressed and a few test fixes. You are almost there (i expect after that we can merge). Thanks for your patience ✋

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
javier-gomez-hs: [This one](https://github.com/javier-gomez-hs/killbill/blob/e7c1e41dc2492843cbf5af7c57972f3b6fc053ef/payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java#L150) is using the "true, false" combination.
Should I add the verification of the auto_pay_off? Or write a new test?

Thanks!

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
javier-gomez-hs: To get the Account Tags and check if the account is in AUTO_PAY_OFF, I'd need to inject the TagUserApi into the PaymentTestSuiteWithEmbeddedDB.java.
Is that ok?
Is there any other way to get the AUTO_PAY_OFF tag status for an Account?

Thanks!

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
sbrossie: Cool, you can keep the test (just add an assertion to verify that in that case account ends up in `AUTIO_PAY_OFF` as requested.

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
sbrossie: There is an [api](https://github.com/killbill/killbill/blob/killbill-0.17.1/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java#L1153) that lets you get the tags associated with an account. I think this is implemented on the java-client side (if not add the call).

javier-gomez-hs: @sbrossie 
I've added (injected) the TagUserApi but when I try to get all the Tags associated with the Account, the list is empty.
I was debugging the test, and I have found out that when the tag is being "added" inside the [setAccountAutoPayOff](https://github.com/javier-gomez-hs/killbill/blob/60618ff139816a46465f01feabdfb4c769dfc276/payment/src/main/java/org/killbill/billing/payment/core/ProcessorBase.java#L103-L103) method, the **tagInternalApi** used is a Mock.
I wasn't able to find where is it being set as a Mock instead of the "real" **tagInternalApi**.
Any thoughts?

Thanks!

sbrossie: @javier-gomez-hs As i said earlier you should not inject `TagUserApi`. Instead you should rely on the [api](https://github.com/killbill/killbill/blob/killbill-0.17.1/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java#L1134) to retrieve the tags.

javier-gomez-hs: @sbrossie 
But in order to call that Resource, I'd need the java-client. And as far as I understand the client is not present in that Test package (in fact it is not present in any Api Test package).
Do you know how can I use it from outside that?
In addition, even if I'm able to access the client, I'll still have the issue with the Mocked Tag that I've described [here](https://github.com/killbill/killbill/pull/576#discussion_r70997937)

Thanks!

sbrossie:
@javier-gomez-hs Any status on this PR (i have seen one new [commit](https://github.com/killbill/killbill/pull/576/commits/6c6326e5a2efa4cfc24711c4b2d9674cdaad9429), but that did not address my previous remaining issues. Is everything clear on your end (if not let me know)?

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
sbrossie: Oh my bad, i did not realize this is was a `payment` test (inside the payment code). If you are having trouble making this test work from the payment code itself , then enhance your `profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPayment.java` to write that scenario.

javier-gomez-hs:
@sbrossie 
All PR have been updated with the latest changes.
Please feel free to review them again and let me know if there anything else I should add.

Thanks!

sbrossie:
👍 Will merge shortly

Issue #457, opened 2015-12-20
Introduce DRAFT Invoice (required for Hierarchical Account design)


sbrossie:
Invoice will now have a enum `status` (`DRAFT`, `COMMITTED`). 

Invoice DDL/DAO will need to be modified to reflect that new status.
By default all invoices will be created in `COMMITTED` (current use case), with the exception of apis that:

* `insertExternalCharges`
* `insertCredit`

We need 2 new invoice apis to transition from `DRAFT` to `COMMITTED`. The invoice bus event will only be sent for invoices that are in `COMMITTED` status. When the transition from `DRAFT` to `COMMITTED` occurs, this will send an invoice bus event.

For safety payment module should be modified to disregard non `COMMITTED` invoices.

Adding a user credit or a user external charge on a `COMMITTED` invoice should fail.

Clients libraries need to be updated as well (java + ruby)

Tests:

Invoice unit tests
Beatrix test to verify that a `DRAFT` invoice with a positive balance does not get handled by payment system, and then make the transition and check payment occurs.

matias-aguero-hs:
> We need to new invoice api to transition from DRAFT to COMMITTED

Do you want to receive the new invoice state by parameter or just assume that calling the new method the invoice will move from DRAFT to COMMITTED?

How the new api method will be called? Will be called automatically by a schedule job or manually by InvoiceResource?

pierre:
> How the new api method will be called? Will be called automatically by a schedule job or manually by InvoiceResource?

There are two scenarios:

* Hierarchical accounts: `DRAFT` to `COMMITTED` happens at the end of the day (in the account timezone) using the notification queue (when first creating the `DRAFT` invoice, the parent inserts an entry for that day)
* Because now credits and external charges are created in `DRAFT` mode, we need an API in `InvoiceResource` to manually transition the associated invoice (when the user wants to trigger a payment for example)

matias-aguero-hs:
Thanks, I added a new API in InvoiceResource. For first scenario I'll work on that when I work in #459.

matias-aguero-hs:
> Adding a user credit or a user external charge on a COMMITTED invoice should fail.

Adding this validation [here] (https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java#L352) prevents add a credit if the invoice status is COMMITTED, but there are two tests that fail:
[...]

PR #582, opened 2016-07-19
#457 : making balance = 0 when invoice status is DRAFT

matias-aguero-hs:


invoice/src/main/java/org/killbill/billing/invoice/model/DefaultInvoice.java
```diff
@@ -249,7 +249,9 @@ public BigDecimal getRefundedAmount() {
 
     @Override
     public BigDecimal getBalance() {
-        return hasZeroParentBalance() ? BigDecimal.ZERO : InvoiceCalculatorUtils.computeInvoiceBalance(currency, invoiceItems, payments, isWrittenOff() || isMigrationInvoice());
+        return getStatus().equals(InvoiceStatus.DRAFT) || hasZeroParentBalance() ?
+               BigDecimal.ZERO :
+               InvoiceCalculatorUtils.computeInvoiceBalance(currency, invoiceItems, payments, isWrittenOff() || isMigrationInvoice());
     }
```
matias-aguero-hs: Should we only ask `getStatus().equals(InvoiceStatus.DRAFT)` for parent invoices? Current code doesn't matter what kind of invoice is.

sbrossie: I am tempted to say it does not matter which kind of invoice it is (i.e current code seems good). We can revisit if needed. 👍

Issue #522, opened 2016-03-30
Change AccountResource.payAllInvoices should accept paymentExternalKey


almalki:
we would like to track external payments using their references, no sure if this is also applicable to transactionExternalKey

almalki:
@pierre what do you think, is this a valid use case and you would accept a PR?

pierre:
Are you suggesting to be able to pass a `paymentExternalKey` and/or `transactionExternalKey ` (https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java#L397-L398) to `payAllInvoices`?

This is actually quite tricky because you don't know how many payments will be triggered in advance. What change are you thinking of?

almalki:
@pierre My use case is simple: All payments we have are external, and amounts are not necessarily for single invoice, so the amount may cover multiple invoices and may also exceed and need to be added as credit to the account. We need to track these payments inside kill bill using their references (every external payment we receive has a reference), and we need to correlate payments inside kill bill with these external payments.

pierre:
> the amount may cover multiple invoices

This won't work: Kill Bill assumes that a payment is associated with a single invoice (an invoice could have multiple partial payments though). Without it, you could end up in undetermined scenarios (for example, a partial payment for 2 invoices: is any of them fully paid?).

Did you take a look at [InvoiceResource#createInstantPayment](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L592)? Could you trigger one (external) payment per invoice?

almalki:
@pierre 
I dont see how `createInstantPayment` is different from `payAllInvoices`, both generate random paymentExternalKey and random transactionExternalKey. I wanted to use `payAllInvoices` because it does pay all invoices starting from oldest to latest and if no more invoices it adds the remaining amount as credit. This combo call is very suitable for my case, except this paymentExternalKey problem.

pierre:
> I dont see how `createInstantPayment` is different from `payAllInvoices`, both generate random paymentExternalKey and random transactionExternalKey. 

I overlooked that `createInstantPayment` doesn't let you specify external keys for the payment, that could be fixed easily though.

[...]

Issue #349, opened 2015-07-14
Payment Control state machine: be consistent when creating external keys


pierre:
By convention, we want to set the external key as the payment id if unspecified (see [PaymentModelDao](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/dao/PaymentModelDao.java#L51) and [PaymentTransactionModelDao](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/dao/PaymentTransactionModelDao.java#L59])).

We don't respect this convention however in case of payment control (see [DefaultControlInitiated](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/core/sm/control/DefaultControlInitiated.java#L59-L72)). I don't think we need to set it that early: we can instead implement similar logic in [PaymentAttemptModelDao](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/dao/PaymentAttemptModelDao.java#L61) and set it in the context afterwards.

pierre:
See also [PaymentBusEventHandler](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/payment/src/main/java/org/killbill/billing/payment/bus/PaymentBusEventHandler.java#L106-L107).

sbrossie:
@pierre Can this be closed?

pierre:
We never addressed this. On one hand, it's not easy to fix, on the other, invoice payments and payments going through the control state machine look weird/different (the key is different than the id, yet the user didn't specify it explicitly).

PR #584, opened 2016-07-21
Issue #522 - Payments and Transactions with externalKey

javier-gomez-hs:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
```diff
@@ -501,13 +505,24 @@ public Response createExternalCharges(final Iterable<InvoiceItemJson> externalCh
         final LocalDate requestedDate = toLocalDateDefaultToday(account, requestedDateTimeString, callContext);
         final List<InvoiceItem> createdExternalCharges = invoiceApi.insertExternalCharges(account.getId(), requestedDate, sanitizedExternalChargesJson, autoCommit, callContext);
 
+        // if all createdExternalCharges point to the same invoiceId, use the provided paymentExternalKey and / or transactionExternalKey
+        boolean haveSameInvoiceId = Iterables.all(createdExternalCharges, new Predicate<InvoiceItem>() {
```
pierre: Nit: could be `final`.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
```diff
@@ -501,13 +505,24 @@ public Response createExternalCharges(final Iterable<InvoiceItemJson> externalCh
...
+            @Override
+            public boolean apply(@Nullable final InvoiceItem input) {
```
pierre: Nit: can't be `null` (not `@Nullable`)

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -688,7 +688,7 @@ public Response payAllInvoices(@PathParam("accountId") final String accountId,
                                            invoice.getBalance() : remainingRequestPayment;
             if (amountToPay.compareTo(BigDecimal.ZERO) > 0) {
                 final UUID paymentMethodId = externalPayment ? null : account.getPaymentMethodId();
-                createPurchaseForInvoice(account, invoice.getId(), amountToPay, paymentMethodId, externalPayment, pluginProperties, callContext);
+                createPurchaseForInvoice(account, invoice.getId(), amountToPay, paymentMethodId, externalPayment, UUIDs.randomUUID().toString(), UUIDs.randomUUID().toString(), pluginProperties, callContext);
```
pierre: We should probably pass `null` instead (I know the old code used to specify an `id`, but I'm not sure it's what we want): the payment code will simply set the key to the id of the payment (or transaction respectfully). Simpler / more consistent rather than generating yet another `id`.

javier-gomez-hs: @pierre 
You meant to pass both paymentExternalKey and transactionExternalKey parameters as null?
There is a check [here](https://github.com/javier-gomez-hs/killbill/blob/e7c1e41dc2492843cbf5af7c57972f3b6fc053ef/payment/src/main/java/org/killbill/billing/payment/api/DefaultPaymentApi.java#L295-L295) to verify that the transactionExternalKey is not null.
Should I remove that or send as null the paymentExternalKey and a random UUID as the transactionExternalKey?

javier-gomez-hs: @pierre 
I've made some changes. 
Please take a look and let me know your thoughts.

Thanks!

pierre: > There is a check [here](https://github.com/javier-gomez-hs/killbill/blob/e7c1e41dc2492843cbf5af7c57972f3b6fc053ef/payment/src/main/java/org/killbill/billing/payment/api/DefaultPaymentApi.java#L295-L295) to verify that the transactionExternalKey is not null.

That's odd and it doesn't look consistent. Let me check with @sbrossie and I'll get back to you. Good catch! 💯 

Probably related: https://github.com/killbill/killbill/issues/349.

pierre: The preconditions on the key could indeed be removed. We don't quite remember why they were here in the first place, just make sure after the change the transaction external key is not null (should be equal to the id).

Thanks!

javier-gomez-hs: @pierre @sbrossie 
So, I removed the precondition and tried (from a test) to make sure that the transaction external key is not null.
You said that should be equal to the **id**, which id did you mean? the transactionId? 🤔
At the end of the process, the transactionExternalKey has always an UUID value but does not match with any other ID value associated. So I debugged for quite a while until I got [this!](https://github.com/javier-gomez-hs/killbill/blob/b31a8ad2e463c5bf9869b4d2c69078e3ea5ecb55/payment/src/main/java/org/killbill/billing/payment/core/sm/control/DefaultControlInitiated.java#L94-L94)
Apparently, if at that time the transactionExternalKey is still null, a **random** value is being set 😒
What is your suggestion on that? 

Please let me know and I can make the changes!

Thanks!

pierre: > You said that should be equal to the id, which id did you mean? the transactionId?

Yes, in theory, that should be the behavior (whenever an object has both a key and id fields, if the key isn't specified, the system should set it to the auto-generated id).

> I debugged for quite a while until I got [this!](https://github.com/javier-gomez-hs/killbill/blob/b31a8ad2e463c5bf9869b4d2c69078e3ea5ecb55/payment/src/main/java/org/killbill/billing/payment/core/sm/control/DefaultControlInitiated.java#L94-L94)

Ah yes, that's exactly the issue https://github.com/killbill/killbill/issues/349. 

Is it possible to remove `stateContext.setPaymentTransactionExternalKey` from `DefaultControlInitiated`? The idea would be that the system will set it automatically when the model is built (see https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/dao/PaymentTransactionModelDao.java#L60 for instance). I'm not quite sure why it's different in that case. It's possible that it's because it's needed in `InvoicePaymentControlPluginApi` (in the `priorCall`), which is before the model is built.

Could you give it a try if it's easy to remove it? If it breaks all sorts of tests, don't worry about the check. Let's still pass `null` from JAX-RS and we can tackle the inconsistency in a separate PR.

Thanks!

javier-gomez-hs: @pierre 
There is only one integration test failing after removing the set.
I've been debugging and found out that [here](https://github.com/javier-gomez-hs/killbill/blob/b31a8ad2e463c5bf9869b4d2c69078e3ea5ecb55/payment/src/main/java/org/killbill/billing/payment/core/sm/control/DefaultControlInitiated.java#L120-L120) is being built the PaymentAttemptModelDao and it is throwing an Exception because the TransactionExternalKey is null.
Should I do something similar to [this](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/dao/PaymentTransactionModelDao.java#L60) to avoid this issue?

Thanks!

pierre: Which test is failing? What kind of exception do you get?

The pattern is different for `PaymentAttemptModelDao`: the object refers to a transaction, it doesn't _own_ `transactionId` / `transactionExternalKey`. Besides, setting it there would make the payment control context out-of-sync (unless you update it), because the transaction key would still be `null`.

That being said, I now remember why the inconsistency was hard to fix: the id / key are created by `PaymentAutomatonDAOHelper`, called by `PaymentLeavingStateCallback` (payment state machine). At this point, the `priorCall` from the payment control state machine was already invoked (when we need the key). So I would suggest not to change it, and leave `DefaultControlInitiated` as-is (still pass `null` for both keys from JAX-RS though).

Also, FYI, it looks like you'll need to re-base the PR (GitHub reports conflicts).

javier-gomez-hs: @pierre 
Ok sure! The DefaultControlInitiated will remain as-is.
I've updated the code with the removal of the not null verification of the transactionExternalKey (All integration tests are running fine) and I've also merged with the latest changes in work-for-release-0.17.x

Thanks!

pierre: The last patch looks good. Could you just remove the leftover random UUIDs created for the transaction external keys? I'll merge the PR then.

Thanks!

javier-gomez-hs: @pierre 
Done! 
Please let me know if there is anything else I should change.

Thanks!

Issue #572, opened 2016-07-08
Create new API to retrieve invoice items details associated to SUMMARY item.


sbrossie:
We started the work where the [DAO cal](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L1062)l has been implemented, but we need the new java apis, JAXRS api, java client, profile tests.

matias-aguero-hs:
@sbrossie I'll be working on this issue this week. (if nobody already started)
What parameters the api (resource) should receive? and what information it should return?

sbrossie:
@matias-aguero-hs Sounds good. Will reply with an API proposal. Also don't forget [this remaining issue](https://github.com/killbill/killbill/issues/457#issuecomment-233027870) (Or let me know if this has already been addressed).

sbrossie:
I am thinking we could just enhance our existing endpoint to retrieve invoice [by ID](https://github.com/killbill/killbill/blob/killbill-0.17.1/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L175) or [number](https://github.com/killbill/killbill/blob/killbill-0.17.1/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L197) so they support a new query param `QUERY_INVOICE_WITH_CHILDREN_ITEMS` (similar to existing `QUERY_INVOICE_WITH_ITEMS`):

* For a child invoice that new param would be ignored
* For the parent invoice the following could happen:
  * Neither `QUERY_INVOICE_WITH_CHILDREN_ITEMS` and `QUERY_INVOICE_WITH_ITEMS` => just return invoice with no items (same for child invoice)
  * Only `QUERY_INVOICE_WITH_ITEMS` => returns invoice with its `SUMMARY` items (similar to child behavior, with their own specific items)
  * `QUERY_INVOICE_WITH_CHILDREN_ITEMS` is specified (regardless of `QUERY_INVOICE_WITH_ITEMS`), return all parent and children items.

In that last case, the output would need to be modified (new json model) to have different sections per child:
* SUMMARY_ITEM (child1)
  * ITEM1 (child1)
  * ITEM2 (child1)
  * ...
* SUMMARY_ITEM (child2)
  * ITEM1 (child2)
  * ITEM2 (child2)

sbrossie:
@pierre Any thoughts?

[...]

PR #586, opened 2016-07-22
#572 : new API to retrieve invoice items details associated to SUMMARY item

matias-aguero-hs:


jaxrs/src/main/java/org/killbill/billing/jaxrs/json/InvoiceJson.java
```diff
@@ -22,15 +22,19 @@
 
 import javax.annotation.Nullable;
 
+import org.apache.shiro.util.CollectionUtils;
```
pierre: Wrong import?

Issue #285, opened 2015-03-02
Retries on push notifications


pierre:
`PushNotificationListener` could use its own notification queue to schedule retries.

See discussion at: https://groups.google.com/d/msgid/killbilling-users/CAJE4nfvdYty4SYyCfe1m94uoJWqaZwLGPCJwAAwCsB4Ht-o50A%40mail.gmail.com

matias-aguero-hs:
@pierre I'm working on this issue. I think we should check fails and only retry when the external server cannot be found. Do you think we can ask for a status = 404 [here](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java#L119)?

Does "retry" means insert this event into notification queue again? In this case, how to know what `serviceName` and `queueName` use.

```Java
    /**
     * Creates a new NotificationQueue for a given associated with the given service and queueName
     *  ...
     */
    NotificationQueue createNotificationQueue(final String svcName, final String queueName, final NotificationQueueHandler handler)
            throws NotificationQueueAlreadyExists;
```

Please correct me if I am not in the same page.

pierre:
> Do you think we can ask for a status = 404 [here](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java#L119)?

I think we should keep the logic as is: if the destination app is down (e.g. connect timeout), or unavailable (e.g. returns 500) or misconfigured (e.g. returns 301), we should assume the notification wasn't received and we should retry later (i.e. send it again).

> Does "retry" means insert this event into notification queue again?

We should create a new queue, independent of the queue of the original event, responsible for these events. The queue would be similar to the one handling [payment retries](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/retry/BaseRetryService.java) and do some sort of [exponential delay retries](https://github.com/killbill/killbill/blob/3e0648aebcac969d7c34ef76987b640b3bf231e2/payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentTransactionTask.java#L264-L272) over the span of, say, a few days (configurable).

[...]

PR #589, opened 2016-07-29
#285 : Retries on push notifications.

matias-aguero-hs:


profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java
```diff
@@ -112,13 +137,104 @@ public Response onCompleted(final Response response) throws Exception {
...
             response = futureStatus.get(timeoutSec, TimeUnit.SECONDS);
+        } catch (final TimeoutException toe) {
+            saveRetryPushNotificationInQueue(tenantId, url, body);
+            return false;
         } catch (final Exception e) {
             log.warn("Failed to push notification url='{}', tenantId='{}'", url, tenantId, e);
             return false;
         }
+
+        if ((response.getStatusCode() == 301) || (response.getStatusCode() == 500)) {
```
pierre: I would actually:

```
if (response.getStatusCode() >= 200 && response.getStatusCode() < 300) {
  return true;
} else {
  saveRetryPushNotificationInQueue(tenantId, url, body);
  return false;
}
```

profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java
```diff
@@ -112,13 +137,104 @@ public Response onCompleted(final Response response) throws Exception {
...
+            final NotificationJson notificationJson = mapper.readValue(body, NotificationJson.class);
+            final PushNotificationKey key = new PushNotificationKey(tenantId,
+                                                                    notificationJson.getAccountId() != null ? UUID.fromString(notificationJson.getAccountId()) : null,
+                                                                    notificationJson.getEventType(),
+                                                                    notificationJson.getObjectType(),
+                                                                    notificationJson.getObjectId() != null ? UUID.fromString(notificationJson.getObjectId()) : null,
+                                                                    0, url);
+            saveRetryPushNotificationInQueue(key);
+        } catch (IOException e) {
+            e.printStackTrace();
```
pierre: Logging instead.

profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java
```diff
@@ -112,13 +137,104 @@ public Response onCompleted(final Response response) throws Exception {
...
+        }
+
+        if ((response.getStatusCode() == 301) || (response.getStatusCode() == 500)) {
+            saveRetryPushNotificationInQueue(key);
+        }
+    }
+
+    private void saveRetryPushNotificationInQueue(final UUID tenantId, final String url, final String body) {
+        try {
+            final NotificationJson notificationJson = mapper.readValue(body, NotificationJson.class);
```
pierre: We have it accessible from `dispatchCallback` - could we pass it along to save a deserialization call? It would probably also remove the need to catch the `IOException`.

profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java
```diff
@@ -112,13 +137,104 @@ public Response onCompleted(final Response response) throws Exception {
...
+    private void saveRetryPushNotificationInQueue(final PushNotificationKey key) {
+        final TenantContext tenantContext = contextFactory.createTenantContext(key.getTenantId());
+        key.increaseAttemptNumber();
+        final DateTime nextNotificationTime = getNextNotificationTime(key.getAttemptNumber(), internalCallContextFactory.createInternalTenantContextWithoutAccountRecordId(tenantContext));
+
+        if (nextNotificationTime == null) {
+            log.warn("Max attempt number reached for push notification url='{}', tenantId='{}'", key.getUrl(), key.getTenantId());
+            return;
+        }
+        log.debug("Push notification is schedule to send at {} for url='{}', tenantId='{}'", nextNotificationTime, key.getUrl(), key.getTenantId());
```
pierre: Nit: is schedule**d**

util/src/main/java/org/killbill/billing/util/config/definition/NotificationConfig.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Config("org.killbill.billing.server.notifications.retries")
+    @Default("15m,30m,2h,12h,1d")
+    @Description("Delay before which unresolved push notifications should be retried")
+    List<TimeSpan> getPushNotificationsRetries();
+
+    @Config("org.killbill.billing.server.notifications.retries")
+    @Default("15m,30m,2h,12h,1d")
+    @Description("Delay before which unresolved push notifications should be retried")
+    List<TimeSpan> getPushNotificationsRetries(@Param("dummy") final InternalTenantContext tenantContext);
```
pierre: Nice work making it per-tenant right away!

That being said, to make it truly work with a global fallback (i.e. `getPushNotificationsRetries()` above), you need to do something similar to [MultiTenantPaymentConfig](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/payment/src/main/java/org/killbill/billing/payment/config/MultiTenantPaymentConfig.java).

profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationRetryService.java
```diff
@@ -0,0 +1,94 @@
...
+                                                                              if (!(notificationKey instanceof PushNotificationKey)) {
+                                                                                  log.error("Push Notification service got an unexpected notification type {}", notificationKey.getClass().getName());
+                                                                                  return;
+                                                                              }
+                                                                              final PushNotificationKey key = (PushNotificationKey) notificationKey;
+                                                                              final InternalCallContext callContext = internalCallContextFactory.createInternalCallContext(tenantRecordId, accountRecordId, retryService, CallOrigin.INTERNAL, UserType.SYSTEM, userToken);
+                                                                              try {
+                                                                                  pushNotificationListener.resendPushNotification(key, callContext);
```
pierre: Could we somehow re-use `doPost` to avoid duplicating the sending code?

profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java
sbrossie: Agree with @pierre but keep the `saveRetryPushNotificationInQueue` in case of exception. For those there is a question where we should only reinsert on  `TimeoutException` or on all `Exception`. I am tempted to say the later (on all `Exception`).

profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationKey.java
```diff
@@ -0,0 +1,98 @@
...
+
+public class PushNotificationKey implements NotificationEvent {
+
+    private final UUID tenantId;
+    private final UUID accountId;
+    private final String eventType;
+    private final String objectType;
+    private final UUID objectId;
+    private Integer attemptNumber;
+    private final String url;
```
sbrossie: Nit: I think it would be nicer to keep the field as private final and provide a constructor that takes both a `PushNotificationKey` and the next `attemptNumber `:

```
public PushNotificationKey(final PushNotificationKey input, final Integer nextAttemptNumber) {
 ....
}
```

profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java
```diff
@@ -112,13 +137,104 @@ public Response onCompleted(final Response response) throws Exception {
...
+        if ((response.getStatusCode() == 301) || (response.getStatusCode() == 500)) {
+            saveRetryPushNotificationInQueue(tenantId, url, body);
+            return false;
+        }
+
         return response.getStatusCode() >= 200 && response.getStatusCode() < 300;
     }
 
+    public void resendPushNotification(final PushNotificationKey key, final InternalCallContext callContext) throws JsonProcessingException {
+
```
sbrossie: Most of the code to send the notification is the same one that was used when sending initially. Is there a way to reuse that code instead to avoid duplication?

profiles/killbill/src/main/java/org/killbill/billing/server/notifications/PushNotificationListener.java
matias-aguero-hs: @pierre @sbrossie I refactor the code as Pierre suggested. Can you please confirm that you want to schedule a retry for all exceptions? Code will look like this:
```Java
try {
    final ListenableFuture<Response> futureStatus =
        builder.execute(new AsyncCompletionHandler<Response>() {
            @Override
            public Response onCompleted(final Response response) throws Exception {
                return response;
        }
    });
    response = futureStatus.get(timeoutSec, TimeUnit.SECONDS);
} catch (final Exception e) {
    log.warn("Failed to push notification url='{}', tenantId='{}'", url, tenantId, e);
    saveRetryPushNotificationInQueue(tenantId, url, notification, attemptRetryNumber);
    return false;
}
```

util/src/main/java/org/killbill/billing/util/config/definition/NotificationConfig.java
matias-aguero-hs: done! MultiTenantNotificationConfig class was created.

pierre:
:shipit:

PR #602, opened 2016-08-11
Issue 304 - Add an endpoint to invalidate Kill Bill ehcache

javier-gomez-hs:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/TestResource.java
```diff
@@ -206,6 +210,28 @@ public Response updateTestClockTime(@QueryParam("days") final Integer addDays,
         return getCurrentTime(timeZoneStr);
     }
 
+    @POST
```
pierre: Wondering if we should use `@DELETE` instead. @sbrossie: thoughts?

pierre:
One remark - otherwise LGTM! :shipit:

sbrossie:
@javier-gomez-hs Actually i have not been following the work on that issue very closely, so i am catching up, and i do have a few remarks:
1. I would move the code from `TestResource` to `AdminResource`
2. I did not have in mind that would clear at the cache level i think a common scenario is to clear across all caches on a per account level or per tenant level.  So in summary, here is what i would propose:
* Keep ability to clear on a per-cache level (but also provide ability to not specify a cache in which case all caches would be cleaned up)
* Add ability to clear per account level across following caches (`account-record-id`, `account-immutable`, `account-bcd`). 
* Add ability to clear per tenant level across following caches (`tenant-payment-state-machine-config`, `tenant`, `tenant-kv`, `tenant-config`, `tenant-overdue-config`, `tenant-catalog`, `tenant-record-id`)

sbrossie:
@javier-gomez-hs For per account/tenant level cache i would indeed provide as a key either the `accountId` (per account cache) or for the tenant specifying the tenant headers should be enough (`X-Killbill-Apikey`, `X-Killbill-Apisecret`).

javier-gomez-hs:
@sbrossie @pierre 
I've moved the endpoints to the AdminResource.
The invalidate cache per account is done and with integration tests.
The invalidate cache per tenant is still in progress, I'm still trying to understand the information that is stored in the tenant caches. They are totally different. 
The information that the caches are saving is like:

- tenant-record-id
`Tenant's UUID`

- tenantPaymentStateMachineConfigCache
`"PLUGIN_PAYMENT_STATE_MACHINE_noop::1"`

- tenantCache
Some UUID different than the **tenantId**, not sure what is saving here.

- tenantKvCache
`"PUSH_NOTIFICATION_CB::1"`

- tenantConfigCache
`"1"`

- tenantOverdueConfigCache
`"1"`

- tenantCatalogCache
`"1"`

The last 3 values is the tenant's `recordId`. 
 
What do you guys recommend me to do with the keys that are not related to the tenantId?

Thanks!

pierre:
> What do you guys recommend me to do with the keys that are not related to the tenantId?

The keys are always (indirectly) related to a tenant:

* The `RecordIdApi` can help you retrieve the tenant record_id from the tenantId
* Keys like `XX::Y` have the tenant record_id as the suffix (`Y`) -- check `DefaultTenantUserApi` for the implementation details
* `tenantCache` has the tenant api key as the cache key (see `TenantCacheLoader` for the implementation details)

In general, checking the cache loaders (subclasses of `BaseCacheLoader`) should help you understand what kind of key is used.

Does that make sense?

javier-gomez-hs:
@pierre 
Yes it does! By the time I read your comment I had already realised what is being stored in the tenant caches. Anyway, thanks for that information! 
Now I've updated the code with the latest changes.

Would you like to take a look?

Let me know if I need to change anything else.

Thanks!

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/TestResource.java
```diff
@@ -68,6 +68,8 @@
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
+import net.sf.ehcache.CacheManager;
+import net.sf.ehcache.Ehcache;
```
pierre: Extra imports.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AdminResource.java
```diff
@@ -104,4 +121,99 @@ public boolean apply(final PaymentTransaction input) {
...
+        // getting Tenant Record Id
+        Long tenantRecordId = recordIdApi.getRecordId(tenantContext.getTenantId(), ObjectType.TENANT, tenantContext);
+
+        // clear tenant-record-id cache by tenantId
+        final Ehcache tenantRecordIdCache = cacheManager.getEhcache(CacheType.TENANT_RECORD_ID.getCacheName());
+        tenantRecordIdCache.remove(currentTenant.getId().toString());
+
+        // clear tenant-payment-state-machine-config cache by tenantRecordId
+        final Ehcache tenantPaymentStateMachineConfigCache = cacheManager.getEhcache(CacheType.TENANT_PAYMENT_STATE_MACHINE_CONFIG.getCacheName());
+        String tenantPaymentStateMachineConfigCacheKey = "PLUGIN_PAYMENT_STATE_MACHINE_noop::" + tenantRecordId.toString();
```
pierre: That's not enough: the tenant can have multiple entries (one per payment plugin registered). You'll need to iterate through all the keys.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AdminResource.java
```diff
@@ -104,4 +121,99 @@ public boolean apply(final PaymentTransaction input) {
...
+        tenantPaymentStateMachineConfigCache.remove(tenantPaymentStateMachineConfigCacheKey);
+
+        // clear tenant cache by tenantApiKey
+        final Ehcache tenantCache = cacheManager.getEhcache(CacheType.TENANT.getCacheName());
+        tenantCache.remove(currentTenant.getApiKey());
+
+        // clear tenant-kv cache by tenantRecordId
+        final Ehcache tenantKvCache = cacheManager.getEhcache(CacheType.TENANT_KV.getCacheName());
+        String tenantKvCacheKey = "PUSH_NOTIFICATION_CB::" + tenantRecordId.toString();
```
pierre: Same here, there's more (catalog, overdue, templates, etc.).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AdminResource.java
```diff
@@ -104,4 +121,99 @@ public boolean apply(final PaymentTransaction input) {
         return Response.status(Status.OK).build();
     }
 
+    @POST
```
pierre: Question for @sbrossie: should we do `@DELETE` instead?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AdminResource.java
```diff
@@ -55,6 +64,8 @@
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
+import net.sf.ehcache.CacheManager;
```
pierre: It works today because ehcache is a transitive dependency of killbill-util, already in the jaxrs pom.xml. But better be explicit about it and add the `net.sf.ehcache:ehcache` dependency.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestCache.java
```diff
@@ -0,0 +1,131 @@
...
+        // verify that they are not null and have the expected tenant information
+        Assert.assertNotNull(tenantRecordIdCache);
+        Assert.assertNotNull(tenantRecordIdCache.get(currentTenant.getId().toString()));
+        Assert.assertNotNull(tenantPaymentStateMachineConfigCache);
+        String tenantPaymentStateMachineConfigCacheKey = "PLUGIN_PAYMENT_STATE_MACHINE_noop::1";
+        Assert.assertNotNull(tenantPaymentStateMachineConfigCache.get(tenantPaymentStateMachineConfigCacheKey));
+        Assert.assertNotNull(tenantCache);
+        Assert.assertNotNull(tenantCache.get(DEFAULT_API_KEY));
+        Assert.assertNotNull(tenantKvCache);
+        String tenantKvCacheKey = "PUSH_NOTIFICATION_CB::1";
```
pierre: It doesn't look like you registered an URL specifically, so I'm wondering what this value is? Does the test work if run by itself (I want to make sure it's not a value left over from another test)?

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestCache.java
```diff
@@ -0,0 +1,131 @@
...
+        final Ehcache tenantOverdueConfigCache = cacheManager.getEhcache(CacheType.TENANT_OVERDUE_CONFIG.getCacheName());
+        final Ehcache tenantCatalogCache = cacheManager.getEhcache(CacheType.TENANT_CATALOG.getCacheName());
+
+        DefaultTenant currentTenant = (DefaultTenant) tenantCache.get(tenantCache.getKeys().get(0)).getObjectValue();
+
+        // verify that they are not null and have the expected tenant information
+        Assert.assertNotNull(tenantRecordIdCache);
+        Assert.assertNotNull(tenantRecordIdCache.get(currentTenant.getId().toString()));
+        Assert.assertNotNull(tenantPaymentStateMachineConfigCache);
+        String tenantPaymentStateMachineConfigCacheKey = "PLUGIN_PAYMENT_STATE_MACHINE_noop::1";
```
pierre: Each test regenerates its own tenant - I don't think you can expect the tenant record id to be `1` (try running it on H2 for example to see if it fails: `mvn test -Ph2`).

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestCache.java
```diff
@@ -0,0 +1,131 @@
...
+
+        // invalidate caches per tenant level by tenantId
+        killBillClient.invalidateCacheByTenant(requestOptions);
+
+        // verify that now the caches don't have the previous values
+        Assert.assertNull(tenantRecordIdCache.get(currentTenant.getId().toString()));
+        Assert.assertNull(tenantPaymentStateMachineConfigCache.get(tenantPaymentStateMachineConfigCacheKey));
+        Assert.assertNull(tenantCache.get(DEFAULT_API_KEY));
+        Assert.assertNull(tenantKvCache.get(tenantKvCacheKey));
+        Assert.assertNull(tenantConfigCache.get(1L));
```
pierre: Same remark regarding `1`.

javier-gomez-hs:
@pierre 
I've updated the code with your suggestions and comments.
Please feel free to review it again and let me know if I have to update anything else.

Thanks!

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AdminResource.java
```diff
@@ -104,4 +123,106 @@ public boolean apply(final PaymentTransaction input) {
...
+
+        // clear tenant-catalog cache by tenantRecordId
+        final Ehcache tenantCatalogCache = cacheManager.getEhcache(CacheType.TENANT_CATALOG.getCacheName());
+        tenantCatalogCache.remove(tenantRecordId);
+
+        return Response.status(Status.OK).build();
+    }
+
+    private void removeCacheByKey(final Ehcache tenantCache, final String tenantRecordId) {
+        for (String key : (List<String>) tenantCache.getKeys()) {
```
pierre: Nit: such casts don't always behave well at runtime and can hide some cryptic bugs. Although it's probably not an issue here, I would suggest rewriting it:

```
for (final Object key : tenantCache.getKeys()) {
  if (key != null && key.toString().endsWith("::" + tenantRecordId)) {
    ...
```

pierre:
LGTM! Before merging, I'll have @sbrossie take a look as well when he's back.

Thanks!

javier-gomez-hs:
@pierre 
Thanks for the suggestion! Good catch.
I've updated the code. Let's wait for @sbrossie so he can take a look when he's back!
:shipit: 

Thanks.

sbrossie:
@javier-gomez-hs That looks good to me.

javier-gomez-hs:
Thanks @sbrossie 
:shipit:

Issue #72, opened 2013-09-06
Notes for account


sbrossie:
It would be useful to add a column to accounts where a note can be stored about the account. This would be available in the UI for advo/etc to add details about the account for internal use. Custom fields don't seem to be suited for this as length is limited to 255 characters and I think we would want a higher limit, such as 2k or 4k characters.

javier-gomez-hs:
@sbrossie @pierre 
I've started working with this issue and just made my initial commit.
I still have to add new Integration Tests but I just wanted you to start taking a look at the changes (they are a lot) and tell me if you see anything wrong.

These are the PR:
- killbill-core: https://github.com/killbill/killbill/pull/603
- killbill-api: https://github.com/killbill/killbill-api/pull/20
- killbill-client-java: https://github.com/killbill/killbill-client-java/pull/46
- killbill-oss-parent: https://github.com/killbill/killbill-oss-parent/pull/12

Thanks!

PR #603, opened 2016-08-19
Issue #72 - Notes for account

javier-gomez-hs:
- Added new attribute "notes" to Account object, Json, Dao, etc.
- Updated SQL files and DDL
- Updated Tests with new attribute


account/src/main/resources/org/killbill/billing/account/ddl.sql
```diff
@@ -23,6 +23,7 @@ CREATE TABLE accounts (
     country varchar(50) DEFAULT NULL,
     postal_code varchar(16) DEFAULT NULL,
     phone varchar(25) DEFAULT NULL,
+    notes varchar(4096) DEFAULT NULL,
```
pierre: Could you also provide a migration file?

See http://docs.killbill.io/0.16/database_migrations.html (see https://github.com/killbill/killbill/blob/work-for-release-0.17.x/invoice/src/main/resources/org/killbill/billing/invoice/migration/V20163917173937__invoice_status_459.sql for example).

javier-gomez-hs: Sure @pierre, do I have to create the SQL file manually with the alter table command?
I've never used the **flyway** commands.

Thanks!

pierre: > Sure @pierre, do I have to create the SQL file manually with the alter table command?

Yes, that's it. Just make sure to follow the file name pattern from http://docs.killbill.io/0.16/database_migrations.html#_kill_bill.

javier-gomez-hs: Thanks @pierre.
I've just added the migration file.

Thanks

Issue #255, opened 2015-01-23
Add support for external refunds


pierre:
Scenario: the original credit card has been deleted and the merchant wants to issue a refund by check.

Today, this can be supported by:

* Adding an external payment method to the account and issuing a **payment credit** (to record the payment, i.e. the check)
* Adjusting the balance of the invoice by issuing an **invoice credit** (i.e. CBA)

The main drawback of this approach is that the **payment credit** won't be linked to the invoice, so the **invoice credit** won't be marked as a recognizable credit in the financial reports (i.e. real money associated to the credit vs non-recognizable credit, which is credit given to nice customers for example).

One fix would be to change `PaymentAutomatonRunner`: when retrieving the `effectivePaymentMethodId`, verify if the payment method was deleted. If it has been deleted, fail the refund. The exception can be caught at the JAX-RS level, and if the right query parameter is passed, add the external payment method if needed, and trigger the **payment credit** against that payment method. We still need to link the payment to that invoice somehow.

For testing, we need to write an integration test for the scenario above using the test payment plugin.

matias-aguero-hs:
@pierre I started working on this issue and I have some issues implementing the solution you posted in previous comment.

> if the right query parameter is passed, add the external payment method if needed, and trigger the payment credit against that payment method

Can you please check if I'm in the correct way with **[this code](https://gist.github.com/matias-aguero-hs/7ae4d42f71caad2be320f57e62c35edd#file-paymentresource-java-L510-L542)**?

I'm getting an error when the credit is created over the external payment method:
```Log
INFO  o.k.b.p.api.DefaultPaymentApi  - ENTERING PaymentApi: transactionType='PURCHASE', accountId='7ef6f244-398d-489c-ba76-fd23669fe9c2', paymentMethodId='a2bd323f-4e62-4676-b97f-2af17addaa11', paymentTransactionExternalKey='f1b64a5b-02c0-4066-af5e-1c1737d8e18a', amount='10', currency='USD' 
[...]

PR #606, opened 2016-08-24
#255 support external refunds

matias-aguero-hs:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoicePaymentResource.java
```diff
@@ -177,8 +182,30 @@ public Response createRefundWithAdjustments(final InvoicePaymentTransactionJson
...
         }
 
-        final Payment result = paymentApi.createRefundWithPaymentControl(account, payment.getId(), json.getAmount(), account.getCurrency(), transactionExternalKey,
-                                                                               pluginProperties, createInvoicePaymentControlPluginApiPaymentOptions(false), callContext);
+        final Payment result;
+        PaymentMethod paymentMethod = null;
+        try {
+            paymentMethod = paymentApi.getPaymentMethodById(payment.getPaymentMethodId(), false, false, pluginProperties, callContext);
+        } catch (PaymentApiException e) {
+            log.warn("Payment method {} does not found", payment.getPaymentMethodId());
```
pierre: Because we now have the extra query parameter `externalPayment`, I don't even think we should do this check anymore - just retrieve the payment method when `externalPayment` is false and don't catch the exception.

payment/src/main/java/org/killbill/billing/payment/api/DefaultPaymentApi.java
```diff
@@ -513,7 +516,11 @@ public Payment createCredit(final Account account, final UUID paymentMethodId, @
             logEnterAPICall(log, transactionType, account, paymentMethodId, paymentId, null, amount, currency, paymentExternalKey, paymentTransactionExternalKey, null, null);
 
             final InternalCallContext internalCallContext = internalCallContextFactory.createInternalCallContext(account.getId(), callContext);
-            payment = paymentProcessor.createCredit(IS_API_PAYMENT, NULL_ATTEMPT_ID, account, paymentMethodId, paymentId, amount, currency, paymentExternalKey, paymentTransactionExternalKey,
+            final UUID nonNulPaymentMethodId = (paymentMethodId != null) ?
```
pierre: Nit: typo (`Null`).

payment/src/main/java/org/killbill/billing/payment/core/sm/PaymentAutomatonRunner.java
```diff
@@ -128,8 +129,10 @@ public PaymentStateContext buildPaymentStateContext(final boolean isApiPayment,
                                                         final Iterable<PluginProperty> properties,
                                                         final CallContext callContext,
                                                         final InternalCallContext internalCallContext) throws PaymentApiException {
+
         // Retrieve the payment id from the payment external key if needed
         final UUID effectivePaymentId = paymentId != null ? paymentId : retrievePaymentId(paymentExternalKey, paymentTransactionExternalKey, internalCallContext);
+        verifyPaymentMethodExistsForRefund(effectivePaymentId, transactionType, internalCallContext);
```
pierre: Why do you need this here? Shouldn't it be done already when the payment system retrieves the plugin name (in `PaymentAutomatonDAOHelper` and/or in `ProcessorBase`)?

matias-aguero-hs: I commented this line, run my test again and I got the refund despite that the payment method was deleted. It's not throwing an exception. `PaymentAutomatonDAOHelper#getPaymentProviderPluginName` loads the payment method by `paymentDao.getPaymentMethodIncludedDeleted(paymentMethodId, internalCallContext)`. I think this is the reason why is returning the payment method deleted anyway.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoicePaymentResource.java
matias-aguero-hs: So, do you mean that if `externalPayment=true` we assume that it's an external refund?

payment/src/main/java/org/killbill/billing/payment/core/sm/PaymentAutomatonRunner.java
pierre: You are probably right on the root cause - I don't understand why we call `getPaymentMethodIncludedDeleted` though (and not `getPaymentMethod`). I would try to change this and verify all tests still pass.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoicePaymentResource.java
pierre: Yes, if set to true, assume we are in scenario 2a.

payment/src/main/java/org/killbill/billing/payment/core/sm/PaymentAutomatonRunner.java
sbrossie: I suspect the reason for `getPaymentMethodIncludedDeleted` is that when following payment operations we want(ed) to make sure we go through the same plugin -> gateway. By retrieving the original payment method (even though it could now be deleted), we are able to extract the `pluginName` and route the following operation towards same plugin -> payment gateway. 

In situation where we have payment control plugin this is probably not mandatory, but in other situations this *may* be needed.

matias-aguero-hs: @pierre @sbrossie Did you guys have the chance to discuss it?

sbrossie: @matias-aguero-hs  At a high level, I think the use case where we try to gather the `pluginName` based on a `paymentMethodId` (whether that payment method was deleted or not) makes sense. So the fact that `getPaymentProviderPluginName` uses the `getPaymentMethodIncludedDeleted` seems correct to me.

Now regarding use cases, i can see two different use cases:
1. One use case is the 'GET' path (e.g retrieve detail about a previous payment, including payment plugin properties). Whether payment method associated with that payment was deleted or not, those properties should be returned.
2. Other use case is the creation of new payment transaction. (`REFUND` case being one of them). In all those cases, we should not allow creating new transaction with a deleted payment method.

So probably my thinking is that depending on context (maybe a new boolean) we rely or not on the deleted payment method. Does that make sense?

matias-aguero-hs: @sbrossie yes, does make sense.

Adding a new boolean parameter and running all tests I found this particular case: [Can add a chargeback for deleted payment methods](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestChargeback.java#L89-L89). Is that correct? Can a payment transaction be created over a deleted payment method?

pierre: > Is that correct? Can a payment transaction be created over a deleted payment method?

Chargebacks are special... 😼 

One scenario would be the following:

* Shopper signs-up for free trial
* 30 days later, the shopper is auto-enrolled into a paying plan. CC is charged.
* When the shopper looks at his statement, he forgot signing-up for the service and canceling the free trial. He goes on deleting his account including the payment method.
* Instead of trying to get a refund with customer support, he issues a chargeback with the CC company. Kill Bill might get notified of the chargeback [months](https://chargebacks911.com/chargeback-time-limit/) after the shopper has closed his account. At this point, the payment method has been deleted but we still need to record that chargeback for accounting purposes.

All other operations (auth, capture, void, refund, credit, purchase) do require a valid payment method on file though.

matias-aguero-hs: Roger!

payment/src/main/java/org/killbill/billing/payment/core/sm/PaymentAutomatonDAOHelper.java
```diff
@@ -164,7 +165,7 @@ public String getPaymentProviderPluginName() throws PaymentApiException {
     }
 
     public PaymentPluginApi getPaymentPluginApi() throws PaymentApiException {
-        final String pluginName = getPaymentProviderPluginName();
+        final String pluginName = getPaymentProviderPluginName(true);
         return getPaymentPluginApi(pluginName);
```
sbrossie: I would have expected the opposite: Since this is called when [running payment operations](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java#L74), and that we *only* want to fetch deleted payment method for `GET` operations, then shouldn't this be `false`?

payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java
```diff
@@ -70,7 +70,7 @@ protected PaymentOperation(final GlobalLocker locker,
     public OperationResult doOperationCallback() throws OperationException {
         final String pluginName;
         try {
-            pluginName = daoHelper.getPaymentProviderPluginName();
+            pluginName = daoHelper.getPaymentProviderPluginName(true);
             this.plugin = daoHelper.getPaymentPluginApi();
```
sbrossie: Same remark as before: we only want to fetch deleted payment method for GET operations, then shouldn't this be false? By the way interestingly enough we end up doing twice the call to `getPaymentProviderPluginName`  (once directly and once through `getPaymentPluginApi`). Is there a possibility simplify that (if it's too complicated drop it).

payment/src/main/java/org/killbill/billing/payment/core/sm/PaymentAutomatonDAOHelper.java
matias-aguero-hs: done

payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java
matias-aguero-hs: done

matias-aguero-hs:
> I also realize this goes beyond the original task so i'll leave it to you to decide if this is something you want to tackle or not.

I prefer to open a new issue to  track that refactor separately.

payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java
```diff
@@ -68,17 +68,15 @@ protected PaymentOperation(final GlobalLocker locker,
...
-            pluginName = daoHelper.getPaymentProviderPluginName(true);
             this.plugin = daoHelper.getPaymentPluginApi();
         } catch (final PaymentApiException e) {
             throw convertToUnknownTransactionStatusAndErroredPaymentState(e);
         }
 
         if (paymentStateContext.shouldLockAccountAndDispatch()) {
             // This will already call unwrapExceptionFromDispatchedTask
-            return doOperationCallbackWithDispatchAndAccountLock(pluginName);
+            return doOperationCallbackWithDispatchAndAccountLock(daoHelper.getPluginName());
```
sbrossie: Oh i see. i was getting confused because of the naming: `plugin` is really a `PaymentPluginApi` type (which is unrelated with the `pluginName`). 

Nit: we could rename  `plugin` to `paymentPluginApi` to make it ovbvious

payment/src/main/java/org/killbill/billing/payment/core/sm/PaymentAutomatonDAOHelper.java
```diff
@@ -165,7 +165,7 @@ public String getPaymentProviderPluginName(final boolean includeDeteled) throws
     }
 
     public PaymentPluginApi getPaymentPluginApi() throws PaymentApiException {
-        final String pluginName = getPaymentProviderPluginName(true);
+        final String pluginName = getPaymentProviderPluginName(false);
         return getPaymentPluginApi(pluginName);
```
sbrossie: I suppose changing that value does not have any impact on our tests (they work either way?)

matias-aguero-hs: I ran `mvn clean install -P mysql` and everything passed.

payment/src/main/java/org/killbill/billing/payment/core/sm/payments/PaymentOperation.java
matias-aguero-hs: done

matias-aguero-hs:
> Looks good to me assuming tests pass.

Yes, all tests passed.

> I think it only makes sense to create the issue if you fix it now (since you have some context). If not, we drop it.

I'm working now on Dwolla integration, but you can create it anyway and we can tackle it sometime next week.

sbrossie:
@matias-aguero-hs Created #618

Issue #545, opened 2016-05-18
Catalog defines a `plansAllowedInBundle` but does not seem to be used


sbrossie:
I am also not quite sure about the semantics. I believe that would define how many subscription of a particular `Plan` could exist within the same bundle.

sbrossie:
The XML model already supports limiting how many subscriptions of a particular `Plan` (and for a given bundle) can be created, but this has not been implemented.

javier-gomez-hs:
@sbrossie @pierre 
Hi guys! I've started working on this issue and I was wondering if this is expected to happen only with **ADD_ON** products / plans?
I think so, because when a **BASE** product is used to create an entitlement, the `bundleId` is not being taken into account. Am I right?

Thanks!!

sbrossie:
@javier-gomez-hs  Yes you are correct, The `plansAllowedInBundle` basically restricts the number of ADD_ON subscriptions of a specific kind that can be created within a given bundle.

javier-gomez-hs:
@sbrossie 
Thanks for the info!
Just to confirm: I should add the restriction not only when a new ADD_ON subscription is being created, but also when an existing ADD_ON subscription changes its Plan / Product, right?
For example:
**ACCOUNT** has a base **SUBSCRIPTION** "SA" with 3 **ADD_ONs**: A1, A2 and A3.
A1 and A2 are **ADD_ONs** of the same plan, and A3 has a different plan.
The limit (`plansAllowedInBundle`) is **2**.
Now the user tries to change the A3 **ADD_ON**, to match the same plan of A1 and A2.
Due to the limitation of 2, it should **reject the change** with an exception.
Is that right?

Thanks!

sbrossie:
@javier-gomez-hs Yes this is correct.

javier-gomez-hs:
I have created the first approach:
- killbill-core: https://github.com/killbill/killbill/pull/609
- killbill-api: https://github.com/killbill/killbill-api/pull/21
- killbill-oss-parent: https://github.com/killbill/killbill-oss-parent/pull/13

Please feel free to take a look and let me know if there is something wrong.

Thanks!

javier-gomez-hs:
@sbrossie 
[...]

PR #609, opened 2016-08-30
Issue #545 - PlansAllowedInBundle

javier-gomez-hs:
- Added new control of the attribute plansAllowedInBundle, to limit the number of plans allowed per bundle (in progress)

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
@@ -224,6 +224,8 @@ public class DefaultSubscriptionInternalApi extends SubscriptionApiBase implemen
             }
 
             boolean first = true;
+            final List<SubscriptionBase> subscriptionsForBundle = getSubscriptionsForBundle(bundleId, null, context);
+
             for (EntitlementSpecifier entitlement : entitlements) {
 
                 final PlanPhaseSpecifier spec = entitlement.getPlanPhaseSpecifier();
@@ -247,7 +249,7 @@ public class DefaultSubscriptionInternalApi extends SubscriptionApiBase implemen
                 // verify the number of subscriptions (of the same kind) allowed per bundle and the existing ones
                 if (ProductCategory.ADD_ON.toString().equalsIgnoreCase(plan.getProduct().getCategory().toString())) {
                     if (plan.getPlansAllowedInBundle() != -1 && plan.getPlansAllowedInBundle() > 0) {
-                        int existingAddOnsWithSamePlanName = countExistingAddOnsWithSamePlanName(getSubscriptionsForBundle(bundleId, null, context), plan.getName());
+                        int existingAddOnsWithSamePlanName = countExistingAddOnsWithSamePlanName(subscriptionsForBundle, plan.getName());
                         int currentAddOnsWithSamePlanName = countCurrentAddOnsWithSamePlanName(entitlements, catalog, plan.getName(), effectiveDate, callContext);
                         if ((existingAddOnsWithSamePlanName + currentAddOnsWithSamePlanName) > plan.getPlansAllowedInBundle()) {
                             // a new ADD_ON subscription of the same plan can't be added because it has reached its limit by bundle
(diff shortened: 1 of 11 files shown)


subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
```diff
@@ -221,6 +244,18 @@ public SubscriptionBase createSubscription(final UUID bundleId, final PlanPhaseS
                     }
                 }
 
+                // verify the number of subscriptions (of the same kind) allowed per bundle and the existing ones
+                if (ProductCategory.ADD_ON.toString().equalsIgnoreCase(plan.getProduct().getCategory().toString())) {
+                    if (plan.getPlansAllowedInBundle() != -1 && plan.getPlansAllowedInBundle() > 0) {
+                        int existingAddOnsWithSamePlanName = countExistingAddOnsWithSamePlanName(getSubscriptionsForBundle(bundleId, null, context), plan.getName());
```
sbrossie: Make the `getSubscriptionsForBundle` outside of the `for` loop as it is quite expensive and should only needed once.

sbrossie:
Overall logic + tests look good.

However:
* I am not seeing the logic (not the test) for the  `changePlan` use case?
* I had one comment about performance in the `createBaseSubscriptionWithAddOns` with regard to `getSubscriptionForBundle`.
* Optional (but something to think of): Regarding use of `getSubscriptionForBundle`, I am wondering if there is a way to avoid making that call and use a less expensive DAO call to just retrieve what we need (or maybe additional info when retrieving the bundle

javier-gomez-hs:
@sbrossie 
Yes, you are right I forgot to commit the `changePlan` logic / test.
I'll work on that and the possibility to replace the `getSubscriptionForBundle` with other similar call.

Thanks!

javier-gomez-hs:
@sbrossie 
I've updated the Pull Request with the `changePlan` logic and its test. 

Thanks.

sbrossie:
@javier-gomez-hs The main remark i have is for the `changePlan`: You implemented the logic in the `getDryRunChangePlanEffectiveDate` logic (which works because `entitlement` module calls `dryRun` prior making the change plan). This has the benefit to also implement that logic for dryRun use case, which is nice. A slightly better approach would have been to isolate the logic and call it both from `getDryRunChangePlanEffectiveDate` and `changePlan` (so the two are independent, and if we were to not do dryRun call prior calling `changePlan` that would still work.


If this is easy to fix (< 2 hours of work), in would suggest doing it, if not, this is fine and we can merge it "as is". Let me know whet you think.

sbrossie:
@javier-gomez-hs The main remark i have is for the `changePlan`: You implemented the logic in the `getDryRunChangePlanEffectiveDate` logic (which works because `entitlement` module calls `dryRun` prior making the change plan). This has the benefit to also implement that logic for dryRun use case, which is nice. A slightly better approach would have been to isolate the logic and call it both from `getDryRunChangePlanEffectiveDate` and `changePlan` (so the two are independent, and if we were to not do dryRun call prior calling `changePlan` that would still work.


If this is easy to fix (< 2 hours of work), in would suggest doing it, if not, this is fine and we can merge it "as is". Let me know whet you think.

sbrossie:
@javier-gomez-hs The main remark i have is for the `changePlan`: You implemented the logic in the `getDryRunChangePlanEffectiveDate` logic (which works because `entitlement` module calls `dryRun` prior making the change plan). This has the benefit to also implement that logic for dryRun use case, which is nice. A slightly better approach would have been to isolate the logic and call it both from `getDryRunChangePlanEffectiveDate` and `changePlan` (so the two are independent, and if we were to not do dryRun call prior calling `changePlan` that would still work.


If this is easy to fix (< 2 hours of work), in would suggest doing it, if not, this is fine and we can merge it "as is". Let me know whet you think.

javier-gomez-hs:
@sbrossie 
In order to isolate the logic and call it from both `getDryRunChangePlanEffectiveDate` and `changePlan`, I'll have to do it inside [this method](https://github.com/javier-gomez-hs/killbill/blob/a03d639c1d1c410c3acaa520218b0e814eee266f/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java#L531-L531) in the `DefaultEntitlement.java` class.
I'll have to inject the CatalogUserApi so I can get the current Catalog and its Plans.
Does that sounds ok for you guys?

Thanks!

javier-gomez-hs:
@sbrossie 
Nevermind my last comment.
I've added the new logic in the [DefaultSubscriptionBaseApiService](https://github.com/javier-gomez-hs/killbill/blob/365ce499f9091bd76576bd4219a1efa948bb42ef/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java#L409-L409)

Please review it and let me know if I have to change anything else.

Thanks!

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
```diff
@@ -394,6 +406,16 @@ private void doChangePlan(final DefaultSubscriptionBase subscription,
...
+        if (ProductCategory.ADD_ON.toString().equalsIgnoreCase(newPlan.getProduct().getCategory().toString())) {
+            if (newPlan.getPlansAllowedInBundle() != -1
+                && newPlan.getPlansAllowedInBundle() > 0
+                && countExistingAddOnsWithSamePlanName(dao.getSubscriptions(subscription.getBundleId(), null, internalCallContext), newPlan.getName())
+                   >= newPlan.getPlansAllowedInBundle()) {
+                // the plan can be changed to the new value, because it has reached its limit by bundle
+                throw new SubscriptionBaseApiException(ErrorCode.SUB_CHANGE_AO_MAX_PLAN_ALLOWED_BY_BUNDLE, newPlan.getName());
+            }
+        }
+
```
sbrossie: We are getting there. One more comment: Is `countExistingAddOnsWithSamePlanName` not the same thing as `countExistingAddOnsWithSamePlanName` in `DefaultSubscriptionInternalApi` ? Can we share the code (maybe a static method that can be used in both classes?)

javier-gomez-hs:
@sbrossie 
I've moved the method to the `AddonUtils`, since both classes are using the same method.
Does that sounds good for you?

Thanks!

PR #615, opened 2016-09-14
Issue 611 - Implement missing overdue conditions

javier-gomez-hs:


sbrossie:
@javier-gomez-hs Looks good. I think a slightly better place to write the tests would be under beatrix module --maybe another version of [TestOverdueIntegration](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/beatrix/src/test/java/org/killbill/billing/beatrix/integration/overdue/TestOverdueIntegration.java). The advantage is that instead of use `crappyWaitForLackOfProperSynchonization` (which is either flaky or very slow) you can rely on events to make the synchronization.

Do you think you move those tests?

javier-gomez-hs:
@sbrossie 
Yes, I can move the tests to the beatrix module. I'll have to rewrite them because it uses the APIs instead of the java client.
I'll let you know when I'm done.

Thanks!

javier-gomez-hs:
@sbrossie 
I've pushed my changes to move the Integration tests to the beatrix module.
Sorry for the delay!

Feel free to take a look and let me know if you find something else to change.

Thanks!

Issue #599, opened 2016-08-09
Fix payment api to allow retrieving payment transaction tags


sbrossie:
See [current code](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/TransactionResource.java#L180) where we can't specify `accountId` based on input transactionId;

javier-gomez-hs:
@sbrossie 
I've started with this one and I wonder if you think is ok to add a new parameter to the method in the [Resource](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/TransactionResource.java#L174), because in order to get the Tags we need the `accountId`.
If you agree, I can add it and write a test to verify the behaviour.

Thanks!

sbrossie:
The other option is to enhance our PaymentApi to allow retrieving a `PaymentTransaction` from its ID, and then do a similar thing in the resource than we do [here](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/PaymentResource.java#L886).

@pierre Any thoughts?

javier-gomez-hs:
@sbrossie 
Yes, I agree.
In fact, the `accountId` is needed to get the AccountAuditLogs [here](https://github.com/javier-gomez-hs/killbill/blob/d4ddcc22b75b1e5252e76996f50ee1021a5d6bf2/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java#L196-L196) and add the audit logs to the Tag and build the response.

pierre:
@sbrossie @javier-gomez-hs I agree about adding the new API, something like:

```
Payment getPaymentByTransactionId(paymentTransactionId, bool, bool, properties, context);
```

javier-gomez-github:
@pierre @sbrossie  
Yes, I've been working on that and added a very similar method.
I had some issues with my other account (that's why I'm commenting from this one) so as soon as I can get my account back, I'll push my changes.

Thanks!

javier-gomez-github:
@sbrossie @pierre 
Guys, I wasn't able to recover my old account so I'll start pushing my changes using this account.
The PRs created for this issue are:
- killbill-client-java: https://github.com/killbill/killbill-client-java/pull/48
[...]

PR #620, opened 2016-09-26
Issue #599 - Fix payment api to allow retrieving payment transaction tags

javier-gomez-github:
- Added method to get a Payment Transaction by Id
- Updated method in Resource to get the Payment Transaction Tags by its Id
- Added Integration Tests

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/TransactionResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/TransactionResource.java
@@ -173,12 +173,15 @@ public class TransactionResource extends JaxRsResourceBase {
     @ApiResponses(value = {@ApiResponse(code = 400, message = "Invalid transaction id supplied"),
                            @ApiResponse(code = 404, message = "Invoice not found")})
     public Response getTags(@PathParam(ID_PARAM_NAME) final String id,
+                            @QueryParam(QUERY_WITH_PLUGIN_INFO) @DefaultValue("false") final Boolean withPluginInfo,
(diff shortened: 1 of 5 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/TransactionResource.java
```diff
@@ -172,12 +173,16 @@ public Response deleteCustomFields(@PathParam(ID_PARAM_NAME) final String id,
     @ApiResponses(value = {@ApiResponse(code = 400, message = "Invalid transaction id supplied"),
                            @ApiResponse(code = 404, message = "Invoice not found")})
     public Response getTags(@PathParam(ID_PARAM_NAME) final String id,
+                            @QueryParam(QUERY_WITH_PLUGIN_INFO) @DefaultValue("false") final Boolean withPluginInfo,
+                            @QueryParam(QUERY_WITH_ATTEMPTS) @DefaultValue("false") final Boolean withAttempts,
+                            @QueryParam(QUERY_PLUGIN_PROPERTY) final List<String> pluginPropertiesString,
                             @QueryParam(QUERY_AUDIT) @DefaultValue("NONE") final AuditMode auditMode,
```
sbrossie: I don't think we need any of these properties: Those are for retrieving payment  and here we only care about retrieving tags (fetching the payment is just there to retrieve the `paymentId`).

sbrossie:
@javier-gomez-github There are a few things that need to be fixed but i don't think it should be hard (i'' do it as i merge)

PR #623, opened 2016-10-04
Issue 474 - Mapping currency -> locale is not unique

javier-gomez-github:
- Changed code in DefaultInvoiceFormatter to match the new Currencies and Symbols defined in killbill-api
- Fixed Integration Tests to match the new symbols from killbill-api

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
@@ -26,7 +26,6 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
-import java.util.Map;
 import java.util.UUID;
 
 import org.joda.money.CurrencyUnit;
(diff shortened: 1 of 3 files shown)


invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatter.java
```diff
@@ -240,8 +240,8 @@ private String getFormattedAmountByLocaleAndInvoiceCurrency(final BigDecimal amo
         dfs.setInternationalCurrencySymbol(currencyUnit.getCurrencyCode());
 
         try {
-            final java.util.Currency currency = java.util.Currency.getInstance(invoiceCurrencyCode);
-            dfs.setCurrencySymbol(currency.getSymbol(currencyLocaleMap.get(currency)));
```
pierre: We should now be able to get rid of the `currencyLocaleMap`.

javier-gomez-github: Done!

sbrossie:
@javier-gomez-github After merging i am seeing a compilation error in `DefaultInvoiceFormatterFactory`: Should we get rid of the map [here](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatterFactory.java#L52) as well?

pierre:
> Should we get rid of the map [here](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/invoice/src/main/java/org/killbill/billing/invoice/template/formatters/DefaultInvoiceFormatterFactory.java#L52) as well?

Yup, that map was a hack that this issue fixes.

PR #628, opened 2016-10-15
Jaxrs payment completion fix

sbrossie:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -949,10 +950,15 @@ private Response processPayment(final PaymentTransactionJson json,
...
         if (paymentId != null) {
             final Payment initialPayment = paymentApi.getPayment(paymentId, false, false, pluginProperties, callContext);
-            final PaymentTransaction pendingTransaction = lookupPendingTransaction(initialPayment,
+            final PaymentTransaction pendingOrSuccessTransaction = lookupPendingOrSuccessTransaction(initialPayment,
                                                                                    json != null ? json.getTransactionId() : null,
                                                                                    json != null ? json.getTransactionExternalKey() : null,
                                                                                    json != null ? json.getTransactionType() : null);
+            // If transaction was already completed, return early (See #626)
+            if (pendingOrSuccessTransaction.getTransactionStatus() == TransactionStatus.SUCCESS) {
+                return uriBuilder.buildResponse(uriInfo, PaymentResource.class, "getPayment", pendingOrSuccessTransaction.getPaymentId());
```
andrenpaes: Not sure if it makes a difference, but maybe we should use the `createPaymentResponse` here?

In theory, if the transaction status is SUCCESS, the payment should have also be transitioned correctly.

sbrossie: I considered that, but that would oblige me to fetch payment first because it is needed in `createPaymentResponse`. Is there any advantage doing that?

andrenpaes: You could use the `initialPayment`, but TBH, it doesn't matter.
This is what we want from that method:
https://github.com/killbill/killbill/blob/work-for-release-0.17.x/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java#L566

And it's exactly what you used, so we're good.

PR #629, opened 2016-10-15
jaxrs: Fixes #627

### LINKED ISSUE #627: Enhance completion payment call to allow plugin properties to passed in the body.
The completion endpoint `POST:/1.0/kb/payments` is implemented by looking at the plugin properties in the query parameters. However, the call also takes a body `PaymentTransactionJson` which includes a list `properties`.  Could we extend that call to use those instead (or in addition) ?

The main result for such a change is that in some situation the URL can become quite large (3DS `PaRes`) and web server can deny the request (on tomcat default limit is 8k).

sbrossie:
/cc @andrenpaes @pierre

### DIFF
--- a/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPaymentPluginProperties.java
+++ b/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPaymentPluginProperties.java
@@ -1,0 +1,262 @@
+/*
+ * Copyright 2014-2016 Groupon, Inc
+ * Copyright 2014-2016 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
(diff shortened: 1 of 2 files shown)


profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPaymentPluginProperties.java
```diff
@@ -0,0 +1,262 @@
...
+
+        final List<PluginProperty> bodyProperties = new ArrayList<PluginProperty>();
+        addProperty("keyXXX1", "valXXXX1", bodyProperties, expectProperties);
+        addProperty("keyXXX2", "valXXXX2", bodyProperties, expectProperties);
+        addProperty("keyXXX3", "valXXXX3", bodyProperties, expectProperties);
+
+        testInternal(queryProperties, bodyProperties, expectProperties);
+    }
+
+    @Test(groups = "slow")
```
pierre: Should that method be annotated? Isn't it a utility method?

pierre:
👍

Issue #219, opened 2014-07-31
Create shopping cart API


pierre:
We want to be able to create subscriptions (e.g. a base plan with a few add-ons) in one go (transactional), and potentially create external charges too, but still generate a single invoice.

See https://github.com/killbill/killbill/issues/73.

pierre:
See also https://github.com/killbill/killbill/issues/199.

javier-gomez-hs:
@pierre 
Would you be able to provide more details to this ticket?
The idea of adding a base subscription with a few add-ons in one go already exists, but now we need this to happen in one single transaction? And generating one single invoice?
Am I right?

Thanks!

pierre:
> The idea of adding a base subscription with a few add-ons in one go already exists

Yes: this ticket is to be able to also create one-off charges (i.e. external charges), all in the same invoice.

The difficulty is that we don't have a way to create a single transaction across multiple Kill Bill modules (entitlement/subscription and invoice), without playing with tags (error prone).

Are you starting on this? If so, I'll have to chat with @sbrossie to see what our options are regarding implementation.

pierre:
@javier-gomez-hs: We have decided not to add the ability to create external charges in that API. Instead, we just want to be able to create multiple entitlements (across bundles) atomically.

Work is similar to #199 and the new API would look like:

```
List<Entitlement> createBaseEntitlementsWithAddOns(UUID accountId, Iterable<BaseEntitlementSpecifierWithAddOns> specifiers, Iterable<PluginProperty> properties, CallContext context) throws EntitlementApiException;

public interface BaseEntitlementSpecifierWithAddOns {
  String externalKey;
  Iterable<EntitlementSpecifier> entitlementSpecifier;
  LocalDate entitlementEffectiveDate;
  LocalDate billingEffectiveDate;
  boolean isMigrated;
}
```

The implementation will probably need to create bundles serially (i.e. not atomically) which is fine, since there is no side effect. However, all subscription events across all bundles need to be created atomically (to generate a single invoice).

A new corresponding API in JAX-RS needs to be added as well (and corresponding client changes).

Does that make sense?

sbrossie:
@javier-gomez-hs It seems like you are done with most your tasks. Is this one your next piece? Does the description we provided make sense?

javier-gomez-github:
@sbrossie @pierre 
Sorry I didn't see these comments. Remember that I had to change my account to this one.
The new description does make sense, so I'll start working on this one!

Thanks!

[...]

PR #632, opened 2016-10-20
Issue 219 - Create shopping cart API

javier-gomez-github:
- Added new method to SubscriptionResource to allow the creation of Bulks of Subscriptions and AddOns in one single transaction.
- Added new method to EntitlementApi to allow the creation of Bulks of Subscriptions and AddOns in one single transaction.
- Added new method to DefaultSubscriptionInternalApi to allow the creation of Bulks of Subscriptions and AddOns in one single transaction.
- Added new method to DefaultSubscriptionBaseApiService to allow the creation of Bulks of Subscriptions and AddOns in one single transaction.
- Added new method to DefaultSubscriptionDao to allow the creation of Bulks of Subscriptions and AddOns in one single transaction.


subscription/src/main/java/org/killbill/billing/subscription/engine/dao/SubscriptionDao.java
```diff
@@ -83,6 +83,8 @@
 
     public void createSubscriptionWithAddOns(List<DefaultSubscriptionBase> subscriptions, Map<UUID, List<SubscriptionBaseEvent>> initialEventsMap, InternalCallContext context);
```
pierre: At the DAO level, maybe we can remove this redundant method (`createSubscriptionsWithAddOns` is a superset)?

subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
```diff
@@ -50,6 +51,9 @@ public DefaultSubscriptionBase createPlan(SubscriptionBuilder builder, Plan plan
     public List<DefaultSubscriptionBase> createPlans(Iterable<SubscriptionSpecifier> subscriptions, CallContext context)
```
pierre: Since it's an internal API, can we remove this in favor of `createPlansWithAddOns`?

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
```diff
@@ -513,34 +513,57 @@ public Void inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFa
     }
 
     @Override
-    public void createSubscriptionWithAddOns(final List<DefaultSubscriptionBase> subscriptions, final Map<UUID, List<SubscriptionBaseEvent>> initialEventsMap, final InternalCallContext context) {
+    public void createSubscriptionsWithAddOns(final List<List<DefaultSubscriptionBase>> subscriptionsWithAddOns,
```
pierre: Do we really need to group these as a `List<List<>>`? Or could we simplify it through the entire call chain and flatten it as a simple list?

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
```diff
@@ -275,6 +277,99 @@ public SubscriptionBase apply(final DefaultSubscriptionBase input) {
...
+                final DateTime now = clock.getUTCNow();
+                final DateTime effectiveDate = (entitlementWithAddOnsSpecifier.getEntitlementEffectiveDate() != null) ?
+                                               DefaultClock.truncateMs(entitlementWithAddOnsSpecifier.getEntitlementEffectiveDate().toDateTimeAtCurrentTime()) : now;
+
+                final SubscriptionBaseBundle bundle = createBundleForAccount(accountId, entitlementWithAddOnsSpecifier.getExternalKey(), context);
+                final List<SubscriptionBase> subscriptionWithAddOns = createBaseSubscriptionWithAddOns(bundle.getId(), entitlementWithAddOnsSpecifier.getEntitlementSpecifier(),
+                                                                                                       entitlementWithAddOnsSpecifier.getEntitlementEffectiveDate().toDateTimeAtCurrentTime(),
+                                                                                                       entitlementWithAddOnsSpecifier.isMigrated(), context);
+                boolean first = true;
+                for (EntitlementSpecifier entitlement : entitlementWithAddOnsSpecifier.getEntitlementSpecifier()) {
```
pierre: Do you need these checks below? Aren't they done by `createBaseSubscriptionWithAddOns` above? (I'm just trying to find a way to minimize duplicated code)

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -357,6 +359,147 @@ public Response doResponseOk(final Entitlement entitlement) {
     }
 
     @TimedResource
+    @POST
+    @Path("/createEntitlementsWithAddOns")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Create multiple entitlements with addOn products")
+    @ApiResponses(value = {@ApiResponse(code = 400, message = "Invalid entitlements supplied")})
+    public Response createEntitlementsWithAddOns(final List<BulkBaseSubscriptionAndAddOnsJson> entitlementsWithAddOns,
```
pierre: Could `createEntitlementWithAddOns` above calls this or share some of the code and validations?

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
sbrossie: In principle i agree with @pierre .

However we should also probably do the checks first and then start creating state; if not we can end up creating part way the subscriptions and suddenly throwing an exceptions: Although it is clear there is no contract that this call will create the whole state atomically, we should do everything we can to end up in a half finished situation. So in the end, i would all the checks should happen before.

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
```diff
@@ -275,6 +277,99 @@ public SubscriptionBase apply(final DefaultSubscriptionBase input) {
...
+            final List<List<DefaultSubscriptionBase>> result = apiService.createPlansWithAddOns(accountId, subscriptionAndAddOns, callContext);
+            return buildSubscriptionBaseList(result);
+
+        } catch (CatalogApiException e) {
+            throw new SubscriptionBaseApiException(e);
+        }
+    }
+
+    private List<SubscriptionBase> buildSubscriptionBaseList(final List<List<DefaultSubscriptionBase>> defaultSubscriptionBaseList) {
+        List<SubscriptionBase> result = new ArrayList<SubscriptionBase>();
```
sbrossie: It sucks that we need that :-( It's difficult to say by just reading the diff, if it could be avoided or not.

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
```diff
@@ -118,6 +118,57 @@ public DefaultSubscriptionBase createPlan(final SubscriptionBuilder builder, fin
...
+        List<List<DefaultSubscriptionBase>> subscriptionBaseAndAddOnsList = new ArrayList<List<DefaultSubscriptionBase>>();
+
+        for (SubscriptionAndAddOnsSpecifier subscriptionAndAddOns : subscriptionsAndAddOns) {
+            List<DefaultSubscriptionBase> subscriptionBaseList = new ArrayList<DefaultSubscriptionBase>();
+            for (SubscriptionSpecifier subscription : subscriptionAndAddOns.getSubscriptionSpecifiers()) {
+
+                try {
+                    final DefaultSubscriptionBase subscriptionBase = new DefaultSubscriptionBase(subscription.getBuilder(), this, clock);
+                    final InternalCallContext internalCallContext = createCallContextFromBundleId(subscriptionBase.getBundleId(), context);
+                    final List<SubscriptionBaseEvent> events = getEventsOnCreation(subscriptionBase.getBundleId(), subscriptionBase.getId(), subscriptionBase.getAlignStartDate(),
```
sbrossie: Do we need to create a new `internalCallContext ` at each iteration of the loop (since it should be for the same `tenant` and `account` ?

sbrossie:
@javier-gomez-github Great job on the first iteration (putting all this code together).

I think next stage is to see if you can simplify and reuse as much logic as possible all throughout the stack, and then think of performance (are there any optimization possible ?).

Finally, we should think of a good test plan (probably beatrix + profile for JAXRS).

At that stage, i think it will make sense for us to actually get your code and read through it linearly (just looking at the diff is difficult sometimes).

javier-gomez-github:
@sbrossie @pierre 
I've updated the PR with new changes to reuse code, improve performance and simplify the logic, according to your suggestions in this PR.
I've tested this new method with a real call through the API and worked as expected: multiple bundles with base + addOns, one single transaction and one single invoice generated with all the new items.
Please feel free to review it again! 
In the meantime I'll be writing the Integration tests.

Thanks!

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -357,6 +359,148 @@ public Response doResponseOk(final Entitlement entitlement) {
...
+        final List<BaseEntitlementWithAddOnsSpecifier> baseEntitlementWithAddOnsSpecifierList = new ArrayList<BaseEntitlementWithAddOnsSpecifier>();
+        for (BulkBaseSubscriptionAndAddOnsJson bulkBaseEntitlementWithAddOns : entitlementsWithAddOns) {
+
+            final SubscriptionJson baseEntitlement = Iterables.tryFind(bulkBaseEntitlementWithAddOns.getBaseEntitlementAndAddOns(), new Predicate<SubscriptionJson>() {
+                @Override
+                public boolean apply(final SubscriptionJson subscription) {
+                    return ProductCategory.BASE.toString().equalsIgnoreCase(subscription.getProductCategory());
+                }
+            }).orNull();
+            verifyNonNull(baseEntitlement.getAccountId(), "SubscriptionJson accountId needs to be set for BASE product.");
```
sbrossie: We should also check that `baseEntitlement` is not null

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -357,6 +359,148 @@ public Response doResponseOk(final Entitlement entitlement) {
...
+            final List<EntitlementSpecifier> entitlementSpecifierList = new ArrayList<EntitlementSpecifier>();
+
+            // verify the number of BASE subscriptions
+            final int baseSubscriptionsSize = Iterables.size(Iterables.filter(bulkBaseEntitlementWithAddOns.getBaseEntitlementAndAddOns(), new Predicate<SubscriptionJson>() {
+                @Override
+                public boolean apply(final SubscriptionJson subscription) {
+                    return subscription.getProductCategory().equals(ProductCategory.BASE.toString());
+                }
+            }));
+            verifyNumberOfElements(baseSubscriptionsSize, 1, "Only one BASE product is allowed.");
```
sbrossie: An optimization here would be to use the previous filtering and return a list of `BASE` and check that size exactly equals to 1 (would take care of both validations)

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
sbrossie: Could the previous call `createEntitlementWithAddOns` be implemented by calling this new one `createEntitlementSWithAddOns` by just passing a `List<BulkBaseSubscriptionAndAddOnsJson>` of size one, that is made from the input `final List<SubscriptionJson> entitlements` ?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -245,6 +258,58 @@ public Entitlement doCall(final EntitlementApi entitlementApi, final Entitlement
     }
 
     @Override
+    public List<Entitlement> createBaseEntitlementsWithAddOns(final UUID accountId, final Iterable<BaseEntitlementWithAddOnsSpecifier> baseEntitlementSpecifiersWithAddOns, final Iterable<PluginProperty> properties, final CallContext callContext) throws EntitlementApiException {
+
+        logCreateEntitlementsWithAOs(log, baseEntitlementSpecifiersWithAddOns);
+
+        final List<BaseEntitlementWithAddOnsSpecifier> baseEntitlementSpecifiersWithAddOnsList = new ArrayList<BaseEntitlementWithAddOnsSpecifier>();
+        Iterables.addAll(baseEntitlementSpecifiersWithAddOnsList, baseEntitlementSpecifiersWithAddOns);
```
sbrossie: Do we really need a `List` in the `DefaultEntitlementContext`? If not, that would avoid the copy which seems only needed for that purpose.

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
```diff
@@ -211,66 +212,120 @@ public SubscriptionBase createSubscription(final UUID bundleId, final PlanPhaseS
...
+
+            final List<SubscriptionAndAddOnsSpecifier> subscriptionAndAddOns = new ArrayList<SubscriptionAndAddOnsSpecifier>();
+            for (BaseEntitlementWithAddOnsSpecifier entitlementWithAddOnsSpecifier : baseEntitlementWithAddOnsSpecifier) {
+                final DateTime now = clock.getUTCNow();
+                final DateTime effectiveDate = (entitlementWithAddOnsSpecifier.getEntitlementEffectiveDate() != null) ?
+                                               DefaultClock.truncateMs(entitlementWithAddOnsSpecifier.getEntitlementEffectiveDate().toDateTimeAtCurrentTime()) : now;
+
+                final SubscriptionBaseBundle bundle = createBundleForAccount(accountId, entitlementWithAddOnsSpecifier.getExternalKey(), context);
+
+                final List<SubscriptionSpecifier> subscriptions = verifyAndBuildSubscriptionSpecifiers(bundle.getId(),
```
sbrossie: The method `verifyAndBuildSubscriptionSpecifiers` seems to do too much for what we need (and therefore is costly): Because we know that we are creating new bundles with new subscriptions there is no need to fetch any state associated with these new bundles/subscriptions (hence call `getSubscriptionsForBundle` in `verifyAndBuildSubscriptionSpecifiers` is not required). The ADD_ON check for `SUB_CREATE_AO_MAX_PLAN_ALLOWED_BY_BUNDLE` can be done simply by looking at the arguments for each `BaseEntitlementWithAddOnsSpecifier`.

You could cut the `verifyAndBuildSubscriptionSpecifiers ` in several pieces and only call the lower piece that does the final check.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
javier-gomez-github: You mean to use it as an `Iterable<BaseEntitlementWithAddOnsSpecifier>` instead of a `List<BaseEntitlementWithAddOnsSpecifier>`?

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
javier-gomez-github: I can change that! 
The thing is that what I did was to extract all the content to this new method. But that was already part of the older method `createBaseSubscriptionWithAddOns()`.
So I just extracted it and reused it from both methods.

javier-gomez-github: >
call getSubscriptionsForBundle in verifyAndBuildSubscriptionSpecifiers is not required). The ADD_ON check for SUB_CREATE_AO_MAX_PLAN_ALLOWED_BY_BUNDLE can be done simply by looking at the arguments for each BaseEntitlementWithAddOnsSpecifier.
>

I thinkg that's not enough, since the `getSubscriptionsForBundle()` looks for the existing Subscriptions (already created in the DB) and then it is being verified against the ones that are about to be added.

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/BulkBaseSubscriptionAndAddOnsJson.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModelProperty;
+
+public class BulkBaseSubscriptionAndAddOnsJson {
+
+    @ApiModelProperty(required = true)
+    private final List<SubscriptionJson> baseEntitlementAndAddOns;
+
```
sbrossie: @javier-gomez-github I would suggest to return a `201` and a `Location` header that points to the [endpoint retrieving all bundles for the account](https://github.com/killbill/killbill/blob/killbill-0.17.6/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java#L266)

Note that if we wanted to be fancy, we could add a filter query parameter for that endpoint and return that filtering in the `Location` header, so that we only return the new stuff that was created in the shopping card api. Probably not very difficult to add.

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
sbrossie: > I think that's not enough, since the getSubscriptionsForBundle() looks for the existing Subscriptions (already created in the DB) and then it is being verified against the ones that are about to be added.

In that case i think this is enough because you know there is nothing on disk (there are no existing subscriptions, since all the bundles created in the shopping card api were new? Or is that assumption wrong? Do we support to both create new bundles/subscriptions and also add subscriptions (AO) to existing bundles?

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/BulkBaseSubscriptionAndAddOnsJson.java
sbrossie: I would think the filter is the list of bundles (bundle ids) that should be returned. If such filter  new query param `QUERY_BUNDLES`does not exists, default behavior, if not bundles specified.

When you construct your `Location` you can then set that query params to be the list you just created. Makes sense?

javier-gomez-github: Yes, I've thought about that solution! I'll work on that. Thanks Stephane!

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
javier-gomez-github: >
Do we support to both create new bundles/subscriptions and also add subscriptions (AO) to existing bundles?
>

I don't think so!
I'll try to simplify the method then!

Thanks

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/BulkBaseSubscriptionAndAddOnsJson.java
javier-gomez-github: @sbrossie 
So, I've been working on this part of the issue, modifying the `getAccountBundles()` Resource to accept the list of UUIDs as a `List<String>` called `bundlesFilter`.
The thing is that I had to modify the `buildResponse()` method at `JaxrsUriBuilder` in order to accept a `Map<String, Object>` of queryParams, and the problem is that at the end I'll have to call the endpoint like this:
```
http://127.0.0.1:8080/1.0/kb/accounts/ff5025cc-1026-4992-bc20-72b76d617e48/bundles?bundlesFilter=11c0ce4d-600d-4c06-9172-6ac36760679a&bundlesFilter=2a7d7916-fec2-4366-93ff-1c6627cfcb6e
```

As you can see, the `bundlesFilter` query parameter needs to be repeated so it can be read as a `List<String>` from the Resource.
That works, but I'm having some troubles when I try to "build" the query params Map that will be sent as part of the Location in the Response, because the Key is the same for all of them. I mean, I can do that but I'll have to create a new method in `JaxrsUriBuilder` ONLY for this case. 
Is that what you are expecting? Or should I just return the list of bundles for that Account?

Thanks!

sbrossie: @javier-gomez-github I would suggest to create a comma separated list of values which can easily be extracted as a list, similarly to what happens [here](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java#L636)

```
http://127.0.0.1:8080/1.0/kb/accounts/ff5025cc-1026-4992-bc20-72b76d617e48/bundles?bundlesFilter=11c0ce4d-600d-4c06-9172-6ac36760679a,2a7d7916-fec2-4366-93ff-1c6627cfcb6e
```

Does that make sense?

javier-gomez-github: @sbrossie 
Yes, I've seen that case of the Plugin Properties but I didn't know if you wanted to write a new method to "extract" the `bundleId` (in our case).
I'm working on that! 

Thanks!

sbrossie: @javier-gomez-github A better example would be [here](https://github.com/killbill/killbill/blob/d906e0d33d0a5c3d7900613dc9789b17f8711cd0/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/PaymentResource.java#L253). Plugin properties are more complex because they are URL encoded and such. In your case, you should have almost nothing to do (i think jersey will automatically deserialize the string separated list into the `List`)

javier-gomez-github:
@sbrossie 
I've updated the code with some of the suggestions you proposed and other fixes.
I still have some more to do, but you can start taking a look at the new changes.

Thanks and have a nice weekend!

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -265,15 +266,19 @@ public AccountJson apply(final Account account) {
...
 
+        boolean filter = (null != bundlesFilter && !bundlesFilter.isEmpty());
         final List<SubscriptionBundle> bundles = (externalKey != null) ?
-                                                 subscriptionApi.getSubscriptionBundlesForAccountIdAndExternalKey(uuid, externalKey, tenantContext) :
-                                                 subscriptionApi.getSubscriptionBundlesForAccountId(uuid, tenantContext);
+                                                 ((filter) ? filterBundles(subscriptionApi.getSubscriptionBundlesForAccountIdAndExternalKey(uuid, externalKey, tenantContext), Arrays.asList(bundlesFilter.split(","))) :
+                                                  subscriptionApi.getSubscriptionBundlesForAccountIdAndExternalKey(uuid, externalKey, tenantContext)):
+                                                 ((filter) ? filterBundles(subscriptionApi.getSubscriptionBundlesForAccountId(uuid, tenantContext), Arrays.asList(bundlesFilter.split(","))) :
+                                                  subscriptionApi.getSubscriptionBundlesForAccountId(uuid, tenantContext));
```
sbrossie: Might be more readable to first fetch the bundles (what we had before) and then apply filtering if needed.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -265,15 +266,19 @@ public AccountJson apply(final Account account) {
                            @ApiResponse(code = 404, message = "Account not found")})
     public Response getAccountBundles(@PathParam("accountId") final String accountId,
                                       @QueryParam(QUERY_EXTERNAL_KEY) final String externalKey,
+                                      @QueryParam(QUERY_BUNDLES_FILTER) final String bundlesFilter,
```
sbrossie: I think you could do better than that by directly having a `List<String>` here in the query params-- that way there is no deserialization to be done on your end.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -289,6 +294,16 @@ public BundleJson apply(final SubscriptionBundle input) {
         return Response.status(Status.OK).entity(result).build();
     }
 
+    private List<SubscriptionBundle> filterBundles(final List<SubscriptionBundle> subscriptionBundlesForAccountId, final List<String> bundlesFilter) {
+        List<SubscriptionBundle> result = new ArrayList<SubscriptionBundle>();
+        for (SubscriptionBundle subscription : subscriptionBundlesForAccountId) {
+            if (bundlesFilter.contains(subscription.getId().toString())) {
```
sbrossie: Variable name should probably be change to `bundle` or `subscriptionBundle` since this is not a subscription object.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java
```diff
@@ -248,6 +246,153 @@ public void testCreateSubscriptionWithAddOns() throws Exception {
...
+                                      NextEvent.CREATE, NextEvent.BLOCK,
+                                      NextEvent.CREATE, NextEvent.BLOCK,
+                                      NextEvent.CREATE, NextEvent.BLOCK,
+                                      NextEvent.NULL_INVOICE, NextEvent.NULL_INVOICE,
+                                      NextEvent.NULL_INVOICE, NextEvent.NULL_INVOICE,
+                                      NextEvent.NULL_INVOICE,
+                                      NextEvent.INVOICE,
+                                      NextEvent.PAYMENT, NextEvent.INVOICE_PAYMENT
+                                     );
+        final List<Entitlement> allEntitlements = entitlementApi.createBaseEntitlementsWithAddOns(account.getId(), entitlementWithAddOnsSpecifierList, ImmutableList.<PluginProperty>of(), callContext);
```
sbrossie: Oh boy, that's quite a few events...  ;-)

javier-gomez-github:
@sbrossie 

>
Tests look good.
I am getting confused on what is left at this point.
>

I think this is the list of remaining items:
* Reuse code or reimplement the existing method https://github.com/killbill/killbill/pull/632#discussion_r84964470 
-> I believe I can't call to the new method by sending the list of size one, because from the new method I'm getting a `Response` Object, that is not equal to the one I'm expecting in the old method. The old method expects an `Entitlement` object and the new one uses a `List<Entitlement>` object. Any thoughts?
* Optimize the method https://github.com/killbill/killbill/pull/632#discussion_r84963637
-> Working on that right now
* Simplify the existing methods https://github.com/killbill/killbill/pull/632#discussion_r84969632
-> Will continue with this one
* `QUERY_BUNDLES_FILTER` parameter in `AccountResource` https://github.com/killbill/killbill/pull/632#discussion_r85624924
-> I've tested it and it is not deserializing the comma-separated values into a List of Strings. I'm just receiving a List of size 1, with the comma-separated values in one single String. That is why I had to deserialize it manually.
* Simplify code in `AccountResource` https://github.com/killbill/killbill/pull/632#discussion_r85624814
-> Will continue with this one
* Variable name in `AccountResource` https://github.com/killbill/killbill/pull/632#discussion_r85625013
-> Will fix that

If you think there is anything else left to do, please let me know.

Thanks!

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -245,6 +258,57 @@ public Entitlement doCall(final EntitlementApi entitlementApi, final Entitlement
...
+
+        final WithEntitlementPlugin<List<Entitlement>> createBaseEntitlementsWithAddOns = new WithEntitlementPlugin<List<Entitlement>>() {
+            @Override
+            public List<Entitlement> doCall(final EntitlementApi entitlementApi, final EntitlementContext updatedPluginContext) throws EntitlementApiException {
+                final InternalCallContext contextWithValidAccountRecordId = internalCallContextFactory.createInternalCallContext(accountId, callContext);
+
+                try {
+                    final List<SubscriptionBase> subscriptionsWithAddOns = subscriptionBaseInternalApi.createBaseSubscriptionsWithAddOns(accountId, baseEntitlementSpecifiersWithAddOns, contextWithValidAccountRecordId);
+
+                    final DateTime entitlementRequestedDate = dateHelper.fromLocalDateAndReferenceTime(
```
pierre: Question: why is this date shared? Couldn't it be different depending on the subscription (i.e. shouldn't each blocking even have its own date)?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -94,6 +99,7 @@
 import org.killbill.commons.metrics.TimedResource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.weakref.jmx.internal.guava.collect.Iterators;
```
pierre: ⚠️ Wrong import.

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
```diff
@@ -211,66 +212,120 @@ public SubscriptionBase createSubscription(final UUID bundleId, final PlanPhaseS
...
+                                                                                                   effectiveDate,
+                                                                                                   catalog,
+                                                                                                   callContext);
 
-            for (EntitlementSpecifier entitlement : entitlements) {
+            final List<DefaultSubscriptionBase> result = apiService.createPlans(subscriptions, callContext);
+            return ImmutableList.copyOf(Iterables.transform(result, new Function<DefaultSubscriptionBase, SubscriptionBase>() {
+                @Override
+                public SubscriptionBase apply(final DefaultSubscriptionBase input) {
+                    return (SubscriptionBase) input;
```
pierre: Nit: I don't think you need the cast?

api/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseInternalApi.java
```diff
@@ -52,6 +52,9 @@ public SubscriptionBase createSubscription(UUID bundleId, PlanPhaseSpecifier spe
     public List<SubscriptionBase> createBaseSubscriptionWithAddOns(UUID bundleId, Iterable<EntitlementSpecifier> entitlements, DateTime requestedDateWithMs,
```
pierre: This extra internal method can probably go away (you might already be working on it).

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
javier-gomez-github: @pierre 
I'm not sure about that. The same behavior is happening in the old `createBaseEntitlementWithAddOns()`.
Isn't it one date per each entitlement?
Should I change both of them?

Thanks!

pierre: Yup, I think that's a bug in both methods (cc @sbrossie to confirm)

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -300,60 +298,202 @@ public boolean apply(final SubscriptionJson subscription) {
...
+        final List<BaseEntitlementWithAddOnsSpecifier> baseEntitlementWithAddOnsSpecifierList = new ArrayList<BaseEntitlementWithAddOnsSpecifier>();
+        for (BulkBaseSubscriptionAndAddOnsJson bulkBaseEntitlementWithAddOns : entitlementsWithAddOns) {
+            final Iterable<SubscriptionJson> baseEntitlements = Iterables.filter(
+                    bulkBaseEntitlementWithAddOns.getBaseEntitlementAndAddOns(), new Predicate<SubscriptionJson>() {
+                @Override
+                public boolean apply(final SubscriptionJson subscription) {
+                    return ProductCategory.BASE.toString().equalsIgnoreCase(subscription.getProductCategory());
+                }
+            });
+            Preconditions.checkArgument(Iterators.size(baseEntitlements.iterator()) > 0, "SubscriptionJson Base Entitlement needs to be provided");
```
pierre: > Question: do we need this check? If the user specifies just a list of add-ons to create for an existing bundle, shouldn't the rest of the code already work as-is?

Follow-up on my own question: I think it depends on what we want to support.

Either we assume that this will always be called for a list of [(BASE1, AO1, AO2, ...), (BASE2, AO1', AO2', ...)] subscriptions (in which case the bundleId can never be specified) or we also want to support a list of [(AO3, AO4, ...),(AO3', AO4', ...)] where the associated base subscriptions already exist (in which case the validations need to be relaxed).

It looks like we just want case 1), according to the `DefaultSubscriptionInternalApi` implementation? If that's the case, `bundleId` should never be specified (line 391).

matias-aguero-hs:
> - I think we're still missing the Java client PR

Javier attached all PR [here](https://github.com/killbill/killbill/issues/219#issuecomment-258171867).

> - Do you have an idea on how the endpoint performs? Just wondering if the default timeout of 3s is still a good value.

Let's us try with a bigger bulk of subscriptions and verify if 3s is enough.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
javier-gomez-github: @pierre 
Should I use the current SubscriptionBase's `getCreatedDate()`?
Inside [this](https://github.com/javier-gomez-github/killbill/blob/d4ce9b37a52cd8000a5a6ca1ffa321f777ac1e5b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java#L240) for method.
Or which date should I use to create the Blocking events from the subscription object?

Thanks!

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
javier-gomez-github: As far as I know, we just want case 1 (Base + AddOns). 
So, in summary, you think the check in line 372 is no longer necessary?

pierre: I'm more concerned about line 391 actually - the `bundleId` should never be specified (since we assume the user wants to create a new one). Maybe we should instead write a check for that?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
pierre: I think the date is correctly computed (effective date of the subscription), but it should be in the `for` loop instead (so that the blocking state effective date matches the effective date of the subscription).

Note that we never really use `createdDate` / `updatedDate` in the code, these are more for debugging purposes. Entitlement is all driven by the effective date of the subscription events.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -273,12 +273,14 @@ public Response getAccountBundles(@PathParam("accountId") final String accountId
         final UUID uuid = UUID.fromString(accountId);
         final Account account = accountUserApi.getAccountById(uuid, tenantContext);
 
+        List<SubscriptionBundle> bundles = (externalKey != null) ?
+                                                 subscriptionApi.getSubscriptionBundlesForAccountIdAndExternalKey(uuid, externalKey, tenantContext) :
+                                                 subscriptionApi.getSubscriptionBundlesForAccountId(uuid, tenantContext);
+
         boolean filter = (null != bundlesFilter && !bundlesFilter.isEmpty());
```
sbrossie: Nit: Use final

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -305,59 +298,35 @@ public boolean apply(final SubscriptionJson subscription) {
...
+                final List<BaseEntitlementWithAddOnsSpecifier> baseEntitlementWithAddOnsSpecifierList = new ArrayList<BaseEntitlementWithAddOnsSpecifier>();
+                baseEntitlementWithAddOnsSpecifierList.add(baseEntitlementSpecifierWithAddOns);
 
-                final LocalDate resolvedEntitlementDate = requestedDate != null ? toLocalDate(requestedDate) : toLocalDate(entitlementDate);
-                final LocalDate resolvedBillingDate = requestedDate != null ? toLocalDate(requestedDate) : toLocalDate(billingDate);
-                return entitlementApi.createBaseEntitlementWithAddOns(account.getId(), baseEntitlement.getExternalKey(), entitlementSpecifierList,
-                                                                      resolvedEntitlementDate, resolvedBillingDate, isMigrated, pluginProperties, callContext);
+                return entitlementApi.createBaseEntitlementsWithAddOns(account.getId(), baseEntitlementWithAddOnsSpecifierList, pluginProperties, callContext);
```
sbrossie: That means we probably don't need api `createBaseEntitlementWithAddOns`. @pierre any objection if we remove it?

pierre: 👍

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
javier-gomez-github: @pierre 
I understand that the effectiveDate is the correct one. The thing is that if you see [this line](https://github.com/javier-gomez-github/killbill/blob/d4ce9b37a52cd8000a5a6ca1ffa321f777ac1e5b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java#L282), I'm getting the effectiveDate from the `updatedPluginContext`, not from the `SubscriptionBase` object.
If I move it to inside the `for` statement, the items I have are `SubscriptionBase` which are the recently created Subscriptions and AddOns. And basically, they don't have the `effectiveDate`, but the `createdDate`, `updatedDate`, etc.
In summary, should I iterate over the `updatedPluginContext.getBaseEntitlementWithAddOnsSpecifiers()` until I find the one that matches the `SubscriptionBase` and then use its `effectiveDate`?
I don't see it right.

Thanks!

pierre: If you look at `DefaultSubscriptionInternalApi` and `DefaultSubscriptionBaseApiService`, this effective date is used as the first (creation) event effective date, which should match the `effectiveTransitionTime` of the first `SubscriptionBaseTransition`.

That way, you don't even need to transform the `LocalDate` into a `DateTime`.

javier-gomez-github: @pierre 
Good catch! I forgot to check the Transitions!

Thanks!

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
javier-gomez-github: @pierre 
You are right, the `bundleId` must not be specified.
Should I iterate over all the entitlements to check that the `bundleId` is not present in any of them?
Something like this:

```
final Iterable<SubscriptionJson> entitlementsWithBundleSpecified = Iterables.filter(
    bulkBaseEntitlementWithAddOns.getBaseEntitlementAndAddOns(), new Predicate<SubscriptionJson>() {
        @Override
        public boolean apply(final SubscriptionJson subscription) {
            return subscription.getBundleId() != null;
        }
    }
);
Preconditions.checkArgument(Iterators.size(entitlementsWithBundleSpecified.iterator()) > 0, "BundleId must not be specified when creating new bulks");
```

Thanks.

javier-gomez-github:
@pierre @sbrossie 
I've updated the PR with the new changes suggested.
Could you take a look again?
I still have to fix [this test](https://github.com/javier-gomez-github/killbill/blob/e0c59ec3dabcec77672c8b5bb27c2a94c83f6073/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestMigrationSubscriptions.java#L215-L215) that was using the old method.
The thing is that I'm getting a different set of events when I call my new method, and apparently I'm having some issues with the `entitlementEffectiveDate` and the `billingEffectiveDate`.
Basically in the first iteration I'm getting all the CREATE / INVOICE / PAYMENT / INVOICE_PAYMENT events and that should happen after moving 1 month to the future.
If you have any comments to add to this test that could help me to clarify the expected behavior, that would be very helpful!!

Thanks!!

pierre:
@javier-gomez-github It looks like there are a few conflicts - can you take a look?

javier-gomez-github:
@sbrossie
Yes, those tests have been replaced with the new ones in 'TestSubscription', since both were using the same API.

sbrossie:
@javier-gomez-github I am unclear about the test `testSimpleMigrationBundle` failing: Are you saying that was you create the subscriptions (shopping card api), you now get immediately the notifications that should happen a [month ahead](https://github.com/javier-gomez-github/killbill/blob/e0c59ec3dabcec77672c8b5bb27c2a94c83f6073/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestMigrationSubscriptions.java#L280) ?

javier-gomez-github:
@sbrossie 
Nevermind, I've found the issue with the test and now it is working the same way it was working before I removed the old method.
Basically, if I use [the change suggested](https://github.com/killbill/killbill/pull/632#discussion_r86572921) by @pierre, the `entitlementEffectiveDate` is not being applied as expected.
I've rolled back that change to use the effectiveDate `dateHelper.fromLocalDateAndReferenceTime(updatedPluginContext.getBaseEntitlementWithAddOnsSpecifiers().iterator().next().getEntitlementEffectiveDate()` and now it is working as expected.
Maybe using the Dates of the Transitions is not the same as using the provided / generated effectiveDate. 
Makes sense?

sbrossie:
@javier-gomez-github I 'll sync up with @pierre on the date issue (i am lacking the context right, and this PR is getting messy for me to follow things). Let's assume this is 👍

javier-gomez-github:
@sbrossie
Ok sounds good. I've pushed all my final changes so you can take a deep look.
It's hard for me too to understand the requested / done changes!!

Thanks!

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
sbrossie: @javier-gomez-github Ideally, the endpoint should allow to create a mix of either BASE + AOS, or simply AOs on an existing bundle. However, current implementation assumes we need a BASE (new bundle), so in order to keep things simple at this point, i would suggest to keep that behavior:
* Have a precondition that `bundleId` is NULL
*  Have a precondition that there is always a `BASE` provided.

/cc @pierre

pierre:
@javier-gomez-github Is it easy for you to get the effective date of the `BlockingState` when entering `createBaseEntitlementsWithAddOns` before and after my suggested change? This would help us understand why we mix the entitlement and billing dates (hence invoicing too early).

@sbrossie We might not have time before 0.18.x but we should eventually add another similar migration test with multiple specifiers, to make sure we respect the right dates (migration vs billing) when creating multiple entitlements (which was my initial comment - I think there is a bug and we use the wrong ones for the second and other specs).

javier-gomez-github:
@pierre 
>
Is it easy for you to get the effective date of the BlockingState when entering createBaseEntitlementsWithAddOns before and after my suggested change? This would help us understand why we mix the entitlement and billing dates (hence invoicing too early).
>

Basically we have the `entitlementEffectiveDate` and the `billingEffectiveDate`. The first one is being used **only** to create the Blocking State Events. The second one is our "real" effectiveDate and is being used during the whole process of the creation of the Entitlements and AddOns.
So, if we take into account the example of the migration Test, we have:
* clock = 1 - 1 - 2016
* entitlementMigrationDate = 12 - 20 - 2015
* billingMigrationDate = 2 - 1 - 2016

Now, when building the Blocking States inside [this for loop](https://github.com/javier-gomez-github/killbill/blob/54e695becbe8e97de53593d610917e593c8ee150/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java#L214) Pierre suggested to get the `effectiveTransitionTime` from the first Transition of each SubscriptionBase, but the problem is that this `effectiveTransitionTime` is equal to the `billingMigrationDate` which is not the expected one! The date assigned to each Blocking State event must be equal to the `entitlementMigrationDate`.
Am I right?

Hope this helps to clarify the situation.

Thanks!

pierre:
> Am I right?

Yup, you are! That's exactly the issue.

There are still 2 bugs though:

1. line 215: wrong date used as entitlement dates could be different for multiple bundles specifiers
2. line 218: wrong bundle id used as there could be multiple bundles created (`setBlockingStatesAndPostBlockingTransitionEvent` probably needs to be changed to support multiple bundles)

For 1., here is what I'm suggesting:

* Make `createBaseSubscriptionsWithAddOns` return a `List<SubscriptionBaseWithAddOns>` instead (similar structure to `BaseEntitlementWithAddOnsSpecifier`, where created subscriptions are grouped by bundle)
* Make sure that the ordering of `baseEntitlementSpecifiersWithAddOns` is respected when returning `List<SubscriptionBaseWithAddOns>` so that the matching between a `SubscriptionBaseWithAddOns` and a `SubscriptionBaseWithAddOns` can be done by position in the list
* Use the right effective date when creating the blocking state

Because of the two bugs above, I feel we should take the time to add a test (maybe similar to the migration one) where we have multiple specifiers (to create multiple bundles) with different entitlement dates and billing dates and verify that the blocking states effective date match the entitlement dates and that the subscription transition dates match the billing dates

Does that make sense?

javier-gomez-github:
@pierre @sbrossie 
I've worked on the suggested changes and just pushed my code. Please feel free to review it.
I'd like to write the test with multiple specifiers and different entitlement and billing dates so that will be my next commit.

Note: There is one Integration Test failing in **work-for-release-0.17.x** branch, inside the `TestOverdueIntegration` file.
In [this line](https://github.com/javier-gomez-github/killbill/blob/1b327db7accc0ba7078dbcb9529d735437c6d6b4/beatrix/src/test/java/org/killbill/billing/beatrix/integration/overdue/TestOverdueIntegration.java#L947) I had to remove one of the INVOICE_ADJUSTMENT events to get this test working again.
The thing is that I've seen that Pierre added that line on Monday so maybe it is necessary. I haven't pushed that change so you can decide whether or not that event should be there.

Thanks!

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -208,15 +209,20 @@ private EntitlementSpecifier getFirstEntitlementSpecifier(final Iterable<BaseEnt
...
+                    final List<SubscriptionBaseWithAddOns> subscriptionsWithAddOns = subscriptionBaseInternalApi.createBaseSubscriptionsWithAddOns(accountId, baseEntitlementWithAddOnsSpecifiers, contextWithValidAccountRecordId);
                     final List<BlockingState> blockingStates = new ArrayList<BlockingState>();
-                    for (final SubscriptionBase cur : subscriptionsWithAddOns) {
-                        final BlockingState blockingState = new DefaultBlockingState(cur.getId(), BlockingStateType.SUBSCRIPTION, DefaultEntitlementApi.ENT_STATE_START, EntitlementService.ENTITLEMENT_SERVICE_NAME, false, false, false,
-                                                                                     dateHelper.fromLocalDateAndReferenceTime(updatedPluginContext.getBaseEntitlementWithAddOnsSpecifiers().iterator().next().getEntitlementEffectiveDate(), contextWithValidAccountRecordId));
-                        blockingStates.add(blockingState);
+                    final List<BaseEntitlementWithAddOnsSpecifier> baseEntitlementWithAddOnsSpecifierList = Lists.newArrayList(baseEntitlementWithAddOnsSpecifiers.iterator());
```
pierre: Could you just do a `for` indexed loop to avoid copies?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -208,15 +209,20 @@ private EntitlementSpecifier getFirstEntitlementSpecifier(final Iterable<BaseEnt
...
+                        for (SubscriptionBase subscriptionBase : cur.getSubscriptionBaseList()) {
+                            final BlockingState blockingState = new DefaultBlockingState(subscriptionBase.getId(), BlockingStateType.SUBSCRIPTION,
+                                                                                         DefaultEntitlementApi.ENT_STATE_START,
+                                                                                         EntitlementService.ENTITLEMENT_SERVICE_NAME,
+                                                                                         false, false, false,
+                                                                                         dateHelper.fromLocalDateAndReferenceTime(baseEntitlementWithAddOnsSpecifierList.get(subscriptionsWithAddOns.indexOf(cur)).getEntitlementEffectiveDate(), contextWithValidAccountRecordId));
+                            blockingStates.add(blockingState);
+                        }
+                        entitlementUtils.setBlockingStatesAndPostBlockingTransitionEvent(blockingStates, cur.getBundleId(), contextWithValidAccountRecordId);
```
pierre: Should be called once (that's why I mentioned that the signature may need to be changed, to support multiple bundles).

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
```diff
@@ -267,24 +269,25 @@ public SubscriptionBase createSubscription(final UUID bundleId, final PlanPhaseS
...
+                                                             catalog,
+                                                             callContext)
+                );
                 subscriptionAndAddOns.add(subscriptionAndAddOnsSpecifier);
             }
 
-            final List<DefaultSubscriptionBase> result = apiService.createPlansWithAddOns(accountId, subscriptionAndAddOns, callContext);
-            return ImmutableList.copyOf(Iterables.transform(result, new Function<DefaultSubscriptionBase, SubscriptionBase>() {
+            final List<DefaultSubscriptionBaseWithAddOns> result = apiService.createPlansWithAddOns(accountId, subscriptionAndAddOns, callContext);
+            return ImmutableList.copyOf(Iterables.transform(result, new Function<DefaultSubscriptionBaseWithAddOns, SubscriptionBaseWithAddOns>() {
```
pierre: Nit: I see a lot of `ImmutableList.copyOf(Iterables.transform`. There is some overhead compared to doing a simple `for` loop (which would probably be easier to use in these cases).

javier-gomez-github:
@pierre 
I'm working on the fixes suggested.
Regarding the failing test, in fact it is failing in the **work-for-release-0.17.x** [branch](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/beatrix/src/test/java/org/killbill/billing/beatrix/integration/overdue/TestOverdueIntegration.java#L949), which does not have my changes yet 🤔

pierre:
> I'm working on the fixes suggested.
Regarding the failing test, in fact it is failing in the **work-for-release-0.17.x** [branch](https://github.com/killbill/killbill/blob/work-for-release-0.17.x/beatrix/src/test/java/org/killbill/billing/beatrix/integration/overdue/TestOverdueIntegration.java#L949), which does not have my changes yet 🤔

ack - let us know when you have pushed your latest changes and we'll merge it. We'll take care of fixing that test if need be.

Thanks!

javier-gomez-github:
@pierre 
I've pushed some changes.
I still have [one](https://github.com/killbill/killbill/pull/632#discussion_r87410500) left to push, if you agree, and that'd be all! 🙏

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
pierre: Not sure I follow the issue - it's probably because the diff is too hard to read at this point. I'll take a look at the code itself and fix it if needed, no worries.

pierre:
Last question: were you able to add a test with multiple specifiers, different entitlement and billing dates?

javier-gomez-github:
@pierre 
>
Last question: were you able to add a test with multiple specifiers, different entitlement and billing dates?
>

Working on that! 
🏃

subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java
javier-gomez-github: @pierre 
Sorry, I haven't pushed that change, just asked if you agree using the List definition without the Type, since the `apiService.createPlansWithAddOns()` [method](https://github.com/javier-gomez-github/killbill/blob/8462e60d485c009fa2ddb357ff19802f41ca2415/subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java#L287) returns a `List<DefaultSubscriptionBaseWithAddOns>` and I have to "convert" it into a `List<SubscriptionBaseWithAddOns>` (which is the interface of the received implementation).
If you agree, I can use this:
```
final List result = apiService.createPlansWithAddOns(accountId, subscriptionAndAddOns, callContext);
return result;
```
And that way I can skip the `copyOf()` method and return it as expected.

If you accept that, I'll push it.

pierre: I think we actually do want to keep the type.

My original comment was about rewriting the code to use a simple `for` loop instead of `ImmutableList.copyOf(Iterables.transform` in a few places (easier to read, no need for the `copyOf` as the items can be added in a list right away, etc.).

But now, I am confused why the conversion is even needed since `DefaultSubscriptionBaseWithAddOns` should implement `SubscriptionBaseWithAddOns` already.

I can take a look at it when we merge it - it's more important to verify the implementation with the multiple-dates test.

javier-gomez-github: @pierre 
>
My original comment was about rewriting the code to use a simple for loop instead of ImmutableList.copyOf(Iterables.transform in a few places (easier to read, no need for the copyOf as the items can be added in a list right away, etc.).
>

I've just pushed the code using the for loop.

>
But now, I am confused why the conversion is even needed since DefaultSubscriptionBaseWithAddOns should implement SubscriptionBaseWithAddOns already.
>

I too am confused about that. But if I don't convert it I'm getting the "Incompatible types" error in the compiler:
Required `List<SubscriptionBaseWithAddOns>`, found `List<DefaultSubscriptionBaseWithAddOns>` 🤔

javier-gomez-github:
@pierre 
I've just pushed the new Integration test for a migration with multiple specifiers and different entitlement and billing dates!

Issue #621, opened 2016-09-29
Catalog code performance


sbrossie:
The catalog code implementation  has never been optimized to scale with a large # of plans:
- Lots of `for` loop everywhere (linear searches, computation could become exponential if such looks are nested, to be checked)
- All collections are kept as arrays (in the case of mutable catalog, when adding entries), there is a need to re-alloc for each new entry

sbrossie:
Ran a new experiment with a catalog containing 100k entries. Worst time to find  an entry (with linear search is ~ 5 mSec:

```
    @Test(groups = "fast")
    public void testForIssue621() throws Exception {
        final CatalogUpdater catalogUpdater = new CatalogUpdater("dummy", BillingMode.IN_ADVANCE, clock.getUTCNow(), null);
        int MAX_PLANS = 100000;
        for (int i = 1; i <= MAX_PLANS; i++) {
            final SimplePlanDescriptor desc = new DefaultSimplePlanDescriptor("foo-monthly-" + i , "Foo", ProductCategory.BASE, Currency.EUR, BigDecimal.TEN, BillingPeriod.MONTHLY, 0, TimeUnit.UNLIMITED, ImmutableList.<String>of());
            catalogUpdater.addSimplePlanDescriptor(desc);
        }
        final StandaloneCatalog catalog = catalogUpdater.getCatalog();

        final long timeFirstPlan = findPlanWithTiming(catalog, "foo-monthly-1");
        final long timeLastPlan = findPlanWithTiming(catalog, "foo-monthly-" + MAX_PLANS);

        System.err.println("Time: firstPlan (nanoSec) = " + timeFirstPlan + ", timeLastPlan (nanoSec) = " + timeLastPlan);
    }

    private long findPlanWithTiming(final StandaloneCatalog catalog, final String planName) throws CatalogApiException {
        final long ini = System.nanoTime();
        final Plan plan = catalog.findCurrentPlan(planName);
        assertEquals(plan.getName(), planName);
        final long fini = System.nanoTime();
        return (fini - ini);
    }
```

Results:

```
Time: firstPlan (nanoSec) = 1673000, timeLastPlan (nanoSec) = 5264000
```

sbrossie:
I am re-opening the issue to also address performance in the `StandaloneCatalogMapper` when using catalog plugin api.

sbrossie:
[...]

PR #634, opened 2016-10-24
Use hash lookup in StandaloneCatalogMapper

marksimu:
Contains speed improvements for https://github.com/killbill/killbill/issues/621.


sbrossie:
@marksimu That makes sense. I am wondering if we should not apply the same changes for `tmpDefaultPlans` and then eliminate the functions `findOrIllegalState` ? Could you give it a try and see what new numbers are?

marksimu:
I have made the same change to tmpDefaultPlans.  I have also made an improvement to a for loop.  Improving this loop reduces the time by about 350ms cumulatively.

Issue #639, opened 2016-10-31
Second capture (same key) succeeds while first attempt is in UNKNOWN 


sbrossie:
When a given transaction is in `UNKNOWN`, any subsequent transaction operation with the same `transactionExternalKey` should go through Janitor logic and of the following should happen:
1. The Janitor (through the plugin call) allows to resolve the state, and then depending on resolution `SUCCESS` or `FAILED`, second call should return with success/failure.
2. The Janitor (through the plugin call) does NOT allows to resolve the state, and the second call should throw an exception.

unieagle:
The example test that replays this issue:
```java
    public void testDoubleCaptureWhileExistASuccessfulCapture() throws Exception {
        final String paymentExternalKey = UUID.randomUUID().toString();
        final BigDecimal requestedAmount = BigDecimal.TEN;

        final Payment authorization = createPayment(TransactionType.AUTHORIZE, null, paymentExternalKey, UUID.randomUUID().toString(), requestedAmount, PaymentPluginStatus.PROCESSED);
        assertNotNull(authorization);
        Assert.assertEquals(authorization.getTransactions().size(), 1);
        Assert.assertEquals(authorization.getTransactions().get(0).getTransactionStatus(), TransactionStatus.SUCCESS);

        final String paymentTransactionExternalKey = UUID.randomUUID().toString();

        // 1st capture with UNDEFINED
        Payment pendingPayment = createPayment(TransactionType.CAPTURE, authorization.getId(), paymentExternalKey, paymentTransactionExternalKey, requestedAmount, PaymentPluginStatus.UNDEFINED);
        Assert.assertEquals(pendingPayment.getTransactions().size(), 2);
        Assert.assertEquals(pendingPayment.getTransactions().get(0).getTransactionStatus(), TransactionStatus.SUCCESS);
        Assert.assertEquals(pendingPayment.getTransactions().get(1).getTransactionStatus(), TransactionStatus.UNKNOWN);

        mockPaymentProviderPlugin.overridePaymentPluginStatus(pendingPayment.getId(), pendingPayment.getTransactions().get(1).getId(), PaymentPluginStatus.PROCESSED);

        // 2nd capture request with identical transaction external key
        pendingPayment = createPayment(TransactionType.CAPTURE, authorization.getId(), paymentExternalKey, paymentTransactionExternalKey, requestedAmount, PaymentPluginStatus.PROCESSED);
[...]

PR #643, opened 2016-11-01
fix duplicated transactions issue when there is a previously unknown …

unieagle:
…transaction using same transaction external key

Fixing Issue: https://github.com/killbill/killbill/issues/639

@sbrossie @daliwei please review, thanks!

### DIFF
--- a/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
+++ b/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
@@ -407,6 +407,12 @@ public class PaymentProcessor extends ProcessorBase {
                         throw new PaymentApiException(ErrorCode.PAYMENT_INVALID_OPERATION, paymentStateContext.getTransactionType(), transactionToComplete.getTransactionStatus());
                     }
 
+                    // If Janitor confirmed the status, just return the status to the client.
+                    if (transactionType != TransactionType.AUTHORIZE &&
+                        (transactionToComplete.getTransactionStatus() == TransactionStatus.SUCCESS || transactionToComplete.getTransactionStatus() == TransactionStatus.PAYMENT_FAILURE)) {
+                        return getPayment(paymentId, true, properties, callContext, internalCallContext);
+                    }
+
                     paymentStateContext.setPaymentTransactionModelDao(transactionToComplete);
                 }
             }
--- a/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
+++ b/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
@@ -376,6 +376,13 @@ public class PaymentProcessor extends ProcessorBase {
                 throw new PaymentApiException(ErrorCode.PAYMENT_ACTIVE_TRANSACTION_KEY_EXISTS, paymentStateContext.getPaymentTransactionExternalKey());
             }
 
+            final List<PaymentTransactionModelDao> paymentTransactionsForCurrentPayment = daoHelper.getPaymentDao().getTransactionsForPayment(paymentStateContext.getPaymentId(), paymentStateContext.getInternalCallContext());
+            // Always invoke the Janitor first to get the latest state. The state machine will then
+            // prevent disallowed transitions in case the state couldn't be fixed (or if it's already in a final state).
+            final PaymentPluginApi plugin = getPaymentProviderPlugin(paymentModelDao.getPaymentMethodId(), internalCallContext);
+            final List<PaymentTransactionInfoPlugin> pluginTransactions = getPaymentTransactionInfoPlugins(plugin, paymentModelDao, properties, callContext);
+            paymentModelDao = invokeJanitor(paymentModelDao, paymentTransactionsForCurrentPayment, pluginTransactions, internalCallContext);
+
             if (paymentStateContext.getPaymentTransactionExternalKey() != null) {
                 final List<PaymentTransactionModelDao> allPaymentTransactionsForKey = daoHelper.getPaymentDao().getPaymentTransactionsByExternalKey(paymentStateContext.getPaymentTransactionExternalKey(), internalCallContext);
                 runSanityOnTransactionExternalKey(allPaymentTransactionsForKey, paymentStateContext, internalCallContext);
@@ -383,16 +390,9 @@ public class PaymentProcessor extends ProcessorBase {
 
             if (paymentStateContext.getTransactionId() != null || paymentStateContext.getPaymentTransactionExternalKey() != null) {
                 // If a transaction id or key is passed, we are maybe completing an existing transaction (unless a new key was provided)
-                final List<PaymentTransactionModelDao> paymentTransactionsForCurrentPayment = daoHelper.getPaymentDao().getTransactionsForPayment(paymentStateContext.getPaymentId(), paymentStateContext.getInternalCallContext());
                 PaymentTransactionModelDao transactionToComplete = findTransactionToCompleteAndRunSanityChecks(paymentModelDao, paymentTransactionsForCurrentPayment, paymentStateContext, internalCallContext);
 
                 if (transactionToComplete != null) {
-                    // For completion calls, always invoke the Janitor first to get the latest state. The state machine will then
-                    // prevent disallowed transitions in case the state couldn't be fixed (or if it's already in a final state).
-                    final PaymentPluginApi plugin = getPaymentProviderPlugin(paymentModelDao.getPaymentMethodId(), internalCallContext);
-                    final List<PaymentTransactionInfoPlugin> pluginTransactions = getPaymentTransactionInfoPlugins(plugin, paymentModelDao, properties, callContext);
-                    paymentModelDao = invokeJanitor(paymentModelDao, paymentTransactionsForCurrentPayment, pluginTransactions, internalCallContext);
-
                     final UUID transactionToCompleteId = transactionToComplete.getId();
                     transactionToComplete = Iterables.<PaymentTransactionModelDao>find(paymentTransactionsForCurrentPayment,
                                                                                        new Predicate<PaymentTransactionModelDao>() {
@@ -405,8 +405,6 @@ public class PaymentProcessor extends ProcessorBase {
                     // We can't tell where we should be in the state machine - bail (cannot be enforced by the state machine unfortunately because UNKNOWN and PLUGIN_FAILURE are both treated as EXCEPTION)
                     if (transactionToComplete.getTransactionStatus() == TransactionStatus.UNKNOWN) {
                         throw new PaymentApiException(ErrorCode.PAYMENT_INVALID_OPERATION, paymentStateContext.getTransactionType(), transactionToComplete.getTransactionStatus());
-                    } else if (transactionToComplete.getTransactionStatus() == TransactionStatus.SUCCESS ){
-                        throw new PaymentApiException(ErrorCode.PAYMENT_ACTIVE_TRANSACTION_KEY_EXISTS, paymentStateContext.getPaymentTransactionExternalKey());
                     }
 
                     paymentStateContext.setPaymentTransactionModelDao(transactionToComplete);
--- a/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
+++ b/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
@@ -405,12 +405,8 @@ public class PaymentProcessor extends ProcessorBase {
                     // We can't tell where we should be in the state machine - bail (cannot be enforced by the state machine unfortunately because UNKNOWN and PLUGIN_FAILURE are both treated as EXCEPTION)
                     if (transactionToComplete.getTransactionStatus() == TransactionStatus.UNKNOWN) {
                         throw new PaymentApiException(ErrorCode.PAYMENT_INVALID_OPERATION, paymentStateContext.getTransactionType(), transactionToComplete.getTransactionStatus());
-                    }
-
-                    // If Janitor confirmed the status, just return the status to the client.
-                    if (transactionType != TransactionType.AUTHORIZE &&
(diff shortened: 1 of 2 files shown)


payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
```diff
@@ -400,6 +400,12 @@ public boolean apply(final PaymentTransactionModelDao input) {
                         throw new PaymentApiException(ErrorCode.PAYMENT_INVALID_OPERATION, paymentStateContext.getTransactionType(), transactionToComplete.getTransactionStatus());
                     }
 
+                    // If Janitor confirmed the status, just return the status to the client.
+                    if (transactionType != TransactionType.AUTHORIZE &&
```
unieagle: because, for Authorizes, there will be some logic afterwards (in the running part) to handle the duplicate situation.

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
```diff
@@ -1672,21 +1672,170 @@ public void testCompletionOfUnknownCapture() throws Exception {
...
+        Assert.assertEquals(pendingPayment.getTransactions().size(), 2);
+        Assert.assertEquals(pendingPayment.getTransactions().get(0).getTransactionStatus(), TransactionStatus.SUCCESS);
+        Assert.assertEquals(pendingPayment.getTransactions().get(1).getTransactionStatus(), TransactionStatus.UNKNOWN);
+
+        // but actually successful
+        mockPaymentProviderPlugin.overridePaymentPluginStatus(pendingPayment.getId(), pendingPayment.getTransactions().get(1).getId(), PaymentPluginStatus.PROCESSED);
+
+        final String anotherPaymentTransactionExternalKey = UUID.randomUUID().toString();
+        // 2nd capture request with identical transaction external key
+        pendingPayment = createPayment(TransactionType.CAPTURE, authorization.getId(), paymentExternalKey, anotherPaymentTransactionExternalKey, requestedAmount, PaymentPluginStatus.PROCESSED);
```
sbrossie: Comment is wrong: " 2nd capture request with identical transaction external key" -> " 2nd capture request with a DIFFERENT transaction external key"

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
```diff
@@ -1672,21 +1672,170 @@ public void testCompletionOfUnknownCapture() throws Exception {
...
+        Payment pendingPayment = createPayment(TransactionType.CAPTURE, authorization.getId(), paymentExternalKey, paymentTransactionExternalKey, requestedAmount, PaymentPluginStatus.UNDEFINED);
+        Assert.assertEquals(pendingPayment.getTransactions().size(), 2);
+        Assert.assertEquals(pendingPayment.getTransactions().get(0).getTransactionStatus(), TransactionStatus.SUCCESS);
+        Assert.assertEquals(pendingPayment.getTransactions().get(1).getTransactionStatus(), TransactionStatus.UNKNOWN);
+
+        // but actually failed
+        mockPaymentProviderPlugin.overridePaymentPluginStatus(pendingPayment.getId(), pendingPayment.getTransactions().get(1).getId(), PaymentPluginStatus.ERROR);
+
+        final String anotherPaymentTransactionExternalKey = UUID.randomUUID().toString();
+        // 2nd capture request with identical transaction external key
```
sbrossie: Same here: wrong comment

payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
sbrossie: I am also not clear about whether we should ignore `TransactionType.AUTHORIZE` here. I am tempted to say it would be better to keep the code generic (if indeed it works for all use cases), but i also understand your point about already having some logic for `TransactionType.AUTHORIZE` . I 'd like to have a conversation with @pierre to decide what is the best course of action.

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
unieagle: Good catch!

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
unieagle: Good catch!

payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
```diff
@@ -398,6 +398,8 @@ public boolean apply(final PaymentTransactionModelDao input) {
                     // We can't tell where we should be in the state machine - bail (cannot be enforced by the state machine unfortunately because UNKNOWN and PLUGIN_FAILURE are both treated as EXCEPTION)
                     if (transactionToComplete.getTransactionStatus() == TransactionStatus.UNKNOWN) {
                         throw new PaymentApiException(ErrorCode.PAYMENT_INVALID_OPERATION, paymentStateContext.getTransactionType(), transactionToComplete.getTransactionStatus());
+                    } else if (transactionToComplete.getTransactionStatus() == TransactionStatus.SUCCESS ){
+                        throw new PaymentApiException(ErrorCode.PAYMENT_ACTIVE_TRANSACTION_KEY_EXISTS, paymentStateContext.getPaymentTransactionExternalKey());
```
unieagle: this is another implementation i think it's better.
this implementation will allow only 1 successful txn per transaction external key.

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
```diff
@@ -1446,7 +1446,7 @@ public void testVerifyJanitorFromPendingDuringCompletionFlow() throws PaymentApi
             final Payment completedPayment = createPayment(TransactionType.AUTHORIZE, initialPayment.getId(), initialPayment.getExternalKey(), transactionExternalKey, authAmount, PaymentPluginStatus.PROCESSED);
             Assert.fail();
         } catch (final PaymentApiException e) {
-            Assert.assertEquals(e.getCode(), ErrorCode.PAYMENT_INVALID_OPERATION.getCode());
+            Assert.assertEquals(e.getCode(), ErrorCode.PAYMENT_ACTIVE_TRANSACTION_KEY_EXISTS.getCode());
```
unieagle: The authorize duplication will fail because of linked state machine has no entry from Auth_success to another authorize machine.
But I think it's better to indicate a "successful txn exists with this external key" error than the old check rely on the state machine.
This also answered the previous question about why the "Authorize" was picked up in the previous fixing. @daliwei @sbrossie 
But this fixing will change the behavior under this and also the use case below at line 1467, is there any concern with this modification? @sbrossie

pierre: > The authorize duplication will fail because of linked state machine has no entry from Auth_success to another authorize machine.

To double check my understanding: it used to go through for capture (and other operations) because we support partial captures (link from `CAPTURE_SUCCESS` to `CAPTURE_INIT` and the state machine doesn't enforce the unicity of keys)?

payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
pierre: Why isn't it caught by line 452 below (`// Sanity: verify we don't already have a successful transaction for that key...`)?

pierre:
It looks like your branch is old: the code has changed a bit since. Make sure to rebase on latest master (0.16.10).

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
unieagle: Yes, state machine has no idea of transaction external keys.

unieagle:
> It looks like your branch is old: the code has changed a bit since. Make sure to rebase on latest master (0.16.10).

Just double check, is that version (0.16.10) for NA?

payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
unieagle: > Suggested fix instead:
Move invokeJanitor to line 388, i.e. if there is a transaction id or key, always invoke the Janitor to fix the state

It is there currently, I think you mean put it before the original Sanity checks? But, the `findTransactionToCompleteAndRunSanityChecks` method returned the transaction to complete. Can't invoke Janitor before it.

>Move runSanityOnTransactionExternalKey to line 412, i.e. after having run the Janitor and updated the payment state context

I think this is better and doable, but need more extra and duplicated code to extract the transaction from the collection if put the sanity check here.

----
So I think the current fixing is a better solution. How do you think? @pierre

pierre: > It is there currently, I think you mean put it before the original Sanity checks?

Line numbers reflect the latest master (0.16.10) - please take a look at the latest code since it has changed a bit. You will need to update the PR before we can merge anyways.

> But, the `findTransactionToCompleteAndRunSanityChecks` method returned the transaction to complete. Can't invoke Janitor before it.

The `invokeJanitor` signature doesn't need `transactionToComplete` (it goes through all transactions).

> I think this is better and doable, but need more extra and duplicated code to extract the transaction from the collection if put the sanity check here.

I don't understand this - which code needs to be duplicated?

> So I think the current fixing is a better solution. How do you think?

Nope 😄 The right fix for this bug doesn't involve new logic (i.e. a new `if`) - we just need to invoke the Janitor earlier.

pierre: Here's what I'm suggesting:

```
diff --git a/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java b/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
index 4cd0f16..ffd1718 100644
--- a/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
+++ b/payment/src/main/java/org/killbill/billing/payment/core/PaymentProcessor.java
@@ -376,6 +376,13 @@ public class PaymentProcessor extends ProcessorBase {
                 throw new PaymentApiException(ErrorCode.PAYMENT_ACTIVE_TRANSACTION_KEY_EXISTS, paymentStateContext.getPaymentTransactionExternalKey());
             }

+            final List<PaymentTransactionModelDao> paymentTransactionsForCurrentPayment = daoHelper.getPaymentDao().getTransactionsForPayment(paymentStateContext.getPaymentId(), paymentStateContext.getInternalCallContext());
+            // Always invoke the Janitor first to get the latest state. The state machine will then
+            // prevent disallowed transitions in case the state couldn't be fixed (or if it's already in a final state).
+            final PaymentPluginApi plugin = getPaymentProviderPlugin(paymentModelDao.getPaymentMethodId(), internalCallContext);
+            final List<PaymentTransactionInfoPlugin> pluginTransactions = getPaymentTransactionInfoPlugins(plugin, paymentModelDao, properties, callContext);
+            paymentModelDao = invokeJanitor(paymentModelDao, paymentTransactionsForCurrentPayment, pluginTransactions, internalCallContext);
+
             if (paymentStateContext.getPaymentTransactionExternalKey() != null) {
                 final List<PaymentTransactionModelDao> allPaymentTransactionsForKey = daoHelper.getPaymentDao().getPaymentTransactionsByExternalKey(paymentStateContext.getPaymentTransactionExternalKey(), internalCallContext);
                 runSanityOnTransactionExternalKey(allPaymentTransactionsForKey, paymentStateContext, internalCallContext);
@@ -383,16 +390,9 @@ public class PaymentProcessor extends ProcessorBase {

             if (paymentStateContext.getTransactionId() != null || paymentStateContext.getPaymentTransactionExternalKey() != null) {
                 // If a transaction id or key is passed, we are maybe completing an existing transaction (unless a new key was provided)
-                final List<PaymentTransactionModelDao> paymentTransactionsForCurrentPayment = daoHelper.getPaymentDao().getTransactionsForPayment(paymentStateContext.getPaymentId(), paymentStateContext.getInternalCallContext());
                 PaymentTransactionModelDao transactionToComplete = findTransactionToCompleteAndRunSanityChecks(paymentModelDao, paymentTransactionsForCurrentPayment, paymentStateContext, internalCallContext);

                 if (transactionToComplete != null) {
-                    // For completion calls, always invoke the Janitor first to get the latest state. The state machine will then
-                    // prevent disallowed transitions in case the state couldn't be fixed (or if it's already in a final state).
-                    final PaymentPluginApi plugin = getPaymentProviderPlugin(paymentModelDao.getPaymentMethodId(), internalCallContext);
-                    final List<PaymentTransactionInfoPlugin> pluginTransactions = getPaymentTransactionInfoPlugins(plugin, paymentModelDao, properties, callContext);
-                    paymentModelDao = invokeJanitor(paymentModelDao, paymentTransactionsForCurrentPayment, pluginTransactions, internalCallContext);
-
                     final UUID transactionToCompleteId = transactionToComplete.getId();
                     transactionToComplete = Iterables.<PaymentTransactionModelDao>find(paymentTransactionsForCurrentPayment,
                                                                                        new Predicate<PaymentTransactionModelDao>() {
```

unieagle:
rebased to 0.16.10

unieagle:
@pierre your fixing looks reasonable, code updated, please review. thanks!

payment/src/test/java/org/killbill/billing/payment/api/TestPaymentApi.java
unieagle: @pierre this fixing will change the behavior of those 2 existing use cases (this one and the one right below). do you have an concerns about this?

pierre:
:shipit: 

cc @andrenpaes

pierre:
Merge in 0.17.x as https://github.com/killbill/killbill/commit/60f4fdd4f790a20e6389126ffb10549ad50a6098.

Issue #630, opened 2016-10-17
Improve security of killbill session id's


beccagaspard:
Current killbill session id is an incremental number which would be pretty easy to guess or brute force.  A longer and more random session id would be more secure.  Looks like OWASP [recommends](https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Session_ID_Length) session id of at least 128 bits with at least 64 bits of entropy.

pierre:
Maybe we could use a UUID instead, since it should now be pretty fast to generate them since https://github.com/killbill/killbill/issues/323?

PR #655, opened 2016-11-07
Issue #630 - use UUID for KillBill session id

beccagaspard:
Issue #630 - use UUID for KillBill session id to improve security.

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/security/shiro/dao/JDBCSessionDao.java
+++ b/util/src/main/java/org/killbill/billing/util/security/shiro/dao/JDBCSessionDao.java
@@ -20,14 +20,14 @@ package org.killbill.billing.util.security.shiro.dao;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.UUID;
 
 import javax.inject.Inject;
 
 import org.apache.shiro.session.Session;
 import org.apache.shiro.session.mgt.eis.CachingSessionDAO;
+import org.killbill.billing.util.UUIDs;
 import org.skife.jdbi.v2.IDBI;
-import org.skife.jdbi.v2.Transaction;
-import org.skife.jdbi.v2.TransactionStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -54,15 +54,10 @@ public class JDBCSessionDao extends CachingSessionDAO {
 
     @Override
     protected Serializable doCreate(final Session session) {
-        final Serializable sessionId = jdbcSessionSqlDao.inTransaction(new Transaction<Long, JDBCSessionSqlDao>() {
-            @Override
-            public Long inTransaction(final JDBCSessionSqlDao transactional, final TransactionStatus status) throws Exception {
-                transactional.create(new SessionModelDao(session));
-                return transactional.getLastInsertId();
-            }
-        });
+        final UUID sessionId = UUIDs.randomUUID();
         // See SessionModelDao#toSimpleSession for why we use toString()
         assignSessionId(session, sessionId.toString());
+        jdbcSessionSqlDao.create(new SessionModelDao(session));
         return sessionId;
     }
 
@@ -73,15 +68,9 @@ public class JDBCSessionDao extends CachingSessionDAO {
             return null;
         }
 
-        // Ignore unsupported JSESSIONID cookies
-        final Long recordId;
-        try {
-            recordId = Long.parseLong(sessionId.toString().trim());
-        } catch (final NumberFormatException e) {
-            return null;
-        }
+        final String sessionIdString = sessionId.toString();
+        final SessionModelDao sessionModelDao = jdbcSessionSqlDao.read(sessionIdString);
 
-        final SessionModelDao sessionModelDao = jdbcSessionSqlDao.read(recordId);
         if (sessionModelDao == null) {
             return null;
         }
--- a/util/src/main/java/org/killbill/billing/util/security/shiro/dao/SessionModelDao.java
+++ b/util/src/main/java/org/killbill/billing/util/security/shiro/dao/SessionModelDao.java
@@ -33,6 +33,7 @@ public class SessionModelDao {
     private final Serializer<Map> serializer = new DefaultSerializer<Map>();
 
     private Long recordId;
+    private String id;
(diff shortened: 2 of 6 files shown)


beccagaspard:
A couple questions... :)

First - I see there is also a ddl for the `sessions` table in [killbill-platform](https://github.com/killbill/killbill-platform/blob/master/server/src/main/resources/org/killbill/billing/server/ddl.sql#L119).  I can open a separate pull request in that project, but do I need to write a migration script for flyway in that project as well?

Second - was thinking about adding something like this to the migration script so that current sessions would remain valid after the migration script has run - thoughts?
`update sessions set id = record_id;`

Let me know if you have any other feedback - thanks!

util/src/main/java/org/killbill/billing/util/security/shiro/dao/JDBCSessionDao.java
```diff
@@ -54,15 +55,10 @@ protected void doDelete(final Session session) {
...
     @Override
     protected Serializable doCreate(final Session session) {
-        final Serializable sessionId = jdbcSessionSqlDao.inTransaction(new Transaction<Long, JDBCSessionSqlDao>() {
-            @Override
-            public Long inTransaction(final JDBCSessionSqlDao transactional, final TransactionStatus status) throws Exception {
-                transactional.create(new SessionModelDao(session));
-                return transactional.getLastInsertId();
-            }
-        });
+        UUID sessionId = UUID.randomUUID();
```
pierre: ⚠️ Need to use `UUIDs` instead, for performance.

Nit: can be `final`.

util/src/main/java/org/killbill/billing/util/security/shiro/dao/JDBCSessionDao.java
```diff
@@ -73,15 +69,9 @@ protected Session doReadSession(final Serializable sessionId) {
...
 
-        // Ignore unsupported JSESSIONID cookies
-        final Long recordId;
-        try {
-            recordId = Long.parseLong(sessionId.toString().trim());
-        } catch (final NumberFormatException e) {
-            return null;
-        }
-
+        final String recordId = sessionId.toString();
```
pierre: To avoid confusion, rename the variable to `sessionId`.

util/src/main/resources/org/killbill/billing/util/ddl.sql
```diff
@@ -234,14 +234,16 @@ CREATE TABLE bus_events_history (
...
 create table sessions (
-  record_id serial unique
-, start_timestamp datetime not null
-, last_access_time datetime default null
-, timeout int
-, host varchar(100) default null
-, session_data mediumblob default null
-, primary key(record_id)
+  record_id serial unique,
+  id varchar(36) NOT NULL,
```
pierre: Nit: we prefer the style with a leading comma instead. It would also make the diff smaller ;-)

util/src/main/resources/org/killbill/billing/util/migration/V20161105000000__session_id.sql
```diff
@@ -0,0 +1,2 @@
+alter table sessions add column id varchar(36) not null;
```
pierre: Let's enforce the ordering (i.e. `after record_id`), as it makes it easier to debug exported data.

util/src/main/java/org/killbill/billing/util/security/shiro/dao/SessionModelDao.java
```diff
@@ -32,7 +32,7 @@
 
     private final Serializer<Map> serializer = new DefaultSerializer<Map>();
 
-    private Long recordId;
```
pierre: We should keep reading it (you'll need to update the query too) since it's in the database.

pierre:
> First - I see there is also a ddl for the `sessions` table in [killbill-platform](https://github.com/killbill/killbill-platform/blob/master/server/src/main/resources/org/killbill/billing/server/ddl.sql#L119).  I can open a separate pull request in that project, but do I need to write a migration script for flyway in that project as well?

Just the DDL change is fine (we provide that DDL for specific deployments using the bare metal platform).

> Second - was thinking about adding something like this to the migration script so that current sessions would remain valid after the migration script has run - thoughts?
`update sessions set id = record_id;`

Having a migration path to keep sessions active would be great indeed. Could you test it manually? If it works, :shipit:!

Also, we should probably add notes in the migration guide (https://github.com/killbill/killbill/wiki/Kill-Bill-Migration-0.16.x-to-0.18.x).

beccagaspard:
Thanks for the feedback @pierre - I made the changes you suggested and opened a pull request in killbill-platform for the ddl change.  I tested the session migration and it appeared to work fine - old numeric sessions were still valid after the migration.  Wouldn't mind a second set of eyes checking it over though if you get a chance in your environment.  👀 

As far as performance...  I think everything should work basically the same as before - I added an index on the `id` column so the db query and update performance should be all right.  The sessions will be cached in ehcache the same as before, only difference is the key string will be a little longer.  Might take a bit longer when generating a new session id using `UUIDs`, but most users log in once and then continue to reuse the same session id for the rest of their API calls.   Any other issues you would anticipate with performance?

If you think everything is looking good, I can go ahead and work on updating the wiki.  😄

pierre:
> I tested the session migration and it appeared to work fine - old numeric sessions were still valid after the migration. Wouldn't mind a second set of eyes checking it over though if you get a chance in your environment.

👍 Yup, we were brainstorming about it yesterday... It's related to the performance question though.

> Any other issues you would anticipate with performance?

My main worry is the `update` query, which has been a source of issues in the past. You are right that the query is virtually the same (similar index), I'm just unclear if the `String` vs `Long` comparison will have any meaningful impact (`where` clause).

Changes look good otherwise, but it may take us a few days to convince ourselves that the upgrade as-is is safe 😸 We will definitively merge it (or modify it) before 0.18.0 though (which will happen hopefully this week-end). Any further testing you could do on your end this week is appreciated.

> If you think everything is looking good, I can go ahead and work on updating the wiki.

💯 🎉

On a related topic, it sounds like you updated your environment from 0.16.x to 0.17.x already? If so, feel free to update the migration wiki as needed if we missed anything... 😼 

Thanks!

PR #657, opened 2016-11-10
Support AuditLevel when retrieving bundles and subscriptions

beccagaspard:
Support AuditLevel when retrieving bundles and subscriptions using REST API.  I am building out a custom bundle/subscription view and needed the audit data - thought it might be useful to others as well 😀


sbrossie:
👍

Issue #652, opened 2016-11-06
DDL scripts fail on MySQL 5.7.5 or newer


beccagaspard:
DDL scripts are failing for me on MySQL 5.7.15 with error `Unknown system variable 'storage_engine'`.  Looks like it is due to `storage_engine` being [deprecated and removed](http://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_storage_engine) in MySQL 5.7.5.  It has been replaced by `default_storage_engine`.

pierre:
It looks like `default_storage_engine` works on older versions? If so, happy to merge a PR with that instead.

PR #658, opened 2016-11-10
Issue #652 - storage_engine is deprecated in mysql 5.7.5+

beccagaspard:
Issue #652 - storage_engine is deprecated in mysql 5.7.5+, replacing with default_storage_engine.  I tested this on a few different mysql versions, but would appreciate it if someone else could test as well on their versions.  Is there documentation anywhere about what specific db versions killbill supports?  I will open separate pull request for killbill-platform.  😸

### DIFF
--- a/account/src/main/resources/org/killbill/billing/account/ddl.sql
+++ b/account/src/main/resources/org/killbill/billing/account/ddl.sql
@@ -1,4 +1,4 @@
-/*! SET storage_engine=INNODB */;
+/*! SET default_storage_engine=INNODB */;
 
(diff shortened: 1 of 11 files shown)


pierre:
> I tested this on a few different mysql versions

🎉 Thanks!

> Is there documentation anywhere about what specific db versions killbill supports?

No yet, but maybe we should start it... FYI, we have first hand knowledge of production deployments on various MySQL versions, as well as on MariaDB, Percona, Aurora and PostgreSQL.

PR #705, opened 2017-02-16
Pagination optimizations

pierre:
* Skip `count(*)` query when the resultset is too large. This should be fine since the Pagination API specifies that `getMaxNbRecords()` and `getTotalNbRecords()` may not always be accurate -- this is already the case for plugin searches for instance.
* Add ability to specify an ordering in the pagination queries (negative limit means DESC ordering). This is used by Kaui to display the latest accounts, payments, etc. without having to specify an offset of `count(*) - limit`

Default (current) behavior:

image

Large dataset detected behavior:

image

/cc @andrenpaes


util/src/main/java/org/killbill/billing/util/entity/dao/DefaultPaginationSqlDaoHelper.java
```diff
@@ -55,10 +66,20 @@ public Long inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFa
...
         // Since we want to stream the results out, we don't want to auto-commit when this method returns.
         final EntitySqlDao<M, E> sqlDao = transactionalSqlDao.onDemandForStreamingResults(sqlDaoClazz);
-        final Long totalCount = context !=  null ? sqlDao.getCount(context) : null;
-        final Iterator<M> results = paginationIteratorBuilder.build((S) sqlDao, limit, context);
+        // The count to get maxNbRecords can be expensive on very large datasets. As a heuristic to check how large that number is,
+        // we retrieve 1 record at offset SIMPLE_PAGINATION_THRESHOLD (pretty fast). If we've found a record, that means the count is larger
+        // than this threshold and we don't issue the full count query
+        final Long maxNbRecords;
+        if (context == null || paginationIteratorBuilder.build((S) sqlDao, SIMPLE_PAGINATION_THRESHOLD, 1L, ordering, context).hasNext()) {
+            maxNbRecords = null;
```
andrenpaes: Loved it!

andrenpaes:
Looks good, thought I'm not familiar with the inner workings of skife, so my ability to review the Dao part is somewhat limited...

pierre:
@andrenpaes @sbrossie Regarding the review, I was more interested in agreeing on the approach (`SIMPLE_PAGINATION_THRESHOLD` heuristic). Implementation details are just mechanics, nothing interesting.

Issue #729, opened 2017-03-31
Cleanup externalBundleKeys in InvoiceJson


pierre:
`externalBundleKeys` should be renamed `bundleKeys`.

See https://groups.google.com/forum/#!topic/killbilling-users/tqLBV0xE3OY.

PR #730, opened 2017-03-31
Fix json key mapping

MaxGfeller:
As discussed on the [mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/killbilling-users/tqLBV0xE3OY/bK2GoupXAgAJ), this fixes the incorrect mapping of one JSON attribute.

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/InvoiceJson.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/InvoiceJson.java
@@ -69,7 +69,7 @@ public class InvoiceJson extends JsonBase {
                        @JsonProperty("invoiceNumber") final String invoiceNumber,
                        @JsonProperty("balance") final BigDecimal balance,
                        @JsonProperty("accountId") final String accountId,
-                       @JsonProperty("externalBundleKeys") final String bundleKeys,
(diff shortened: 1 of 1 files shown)


pierre:
Thanks @MaxGfeller! Not quite sure though how safe it is to merge it in 0.18.x (vs 0.20.x), we'll investigate first.

See https://github.com/killbill/killbill/issues/729.

sbrossie:
@pierre On the ruby client side, it looks like [we are already covered](https://github.com/killbill/killbill-client-ruby/blob/master/lib/killbill_client/models/invoice.rb#L16). 

On the client java side, the [model CTOR](https://github.com/killbill/killbill-client-java/blob/master/src/main/java/org/killbill/billing/client/model/Invoice.java#L62) expect `externalBundleKeys `, but the getters/setters are already using the [right name](https://github.com/killbill/killbill-client-java/blob/master/src/main/java/org/killbill/billing/client/model/Invoice.java#L174). Actually my suggestion for java client would be to change the api to return a Set<String> (so client does not have to split the string).

From what we know, this is only used by KAUI timeline api, so all in all, this is probably OK to change in 0.18.x (even if technically this is an api change, at least on the java side).

PR #732, opened 2017-04-04
Improve JSON pattern matching/substitution

andrenpaes:
Make sure we match `JSON` patterns with leading spaces before colons.

Also, improves obfuscation a bit for cases where value to be obfuscated appears before a closing brace (i.e. `{"cvv" : 123}`). Before, the closing brace would be replaced with a `*`.


sbrossie:
👍

Issue #716, opened 2017-03-07
Add display name in catalog for products, plans, pricelists, etc.


pierre:
The pretty name could be used in the customer front-end (e.g. pricing page), on invoices, etc.

Work to do:

* Change catalog schema
* Change invoice to use the name (if present) for the description
* Change JAX-RS
* Add support for internationalization
* Update docs

Originally reported by @beccagaspard [here](https://github.com/killbill/killbill/issues/569#issuecomment-230948188).

Internationalization might be difficult, see my comment [here](https://github.com/killbill/killbill/issues/569#issuecomment-230949710).

sbrossie:
The proposal is to enhance the [CatalogEntity](https://github.com/killbill/killbill-api/blob/killbill-api-0.50.1/src/main/java/org/killbill/billing/catalog/api/CatalogEntity.java#L21) with a new method `public String getPrettyName();`

As a result, one would have the ability to specify a `prettyName` in the XML definition (but we would make this field **non required**) for the following types:

| Catalog abstraction |  Default value | Comment |
| ------------- |-------------| -------------| 
| Product     | `Product#name` ||
| Plan     | `Plan#name`      ||
| PlanPhase | `PlanPhase#name`      | `name` is constructed from appending `Plan#name`-`phaseType`|
| Usage | `Usage#name`  ||
| PriceList | `PriceList#name`  ||
| Unit | `Unit#name`   | Not supported (only default) |

beccagaspard:
@sbrossie - tested this out and made a few additions to support prettyName when using a catalog plugin and to include the pretty name for products/plans in the catalog json when using the REST API.  Let me know if I need to make any modifications! :)

PR #739, opened 2017-04-12
Issue #716 - add support for displayName for products/plans when usin…

beccagaspard:
…g a catalog plugin, add prettyName attribute to catalog json when using REST API.


sbrossie:
Indeed, i forgot about that piece... 🎉

PR #745, opened 2017-05-03
Adding TierBlockPolicy to UsagePriceOverrideJson

sruthipendyala:
We need to display the TierBlockPolicy in our UI, so adding it to UsagePriceOverideJson :)


pierre:
Do you plan on updating the [Java client](https://github.com/killbill/killbill-client-java/blob/work-for-release-0.19.x/src/main/java/org/killbill/billing/client/model/UsagePriceOverride.java) as well?

sruthipendyala:
Just updated the Java Client :)

sbrossie:
👍

PR #755, opened 2017-05-15
Add units list to CatalogJson

beccagaspard:
Discussed w/ @sruthipendyala - we thought it is better to include one list of all the units w/ prettyName instead of including prettyName on every single tier block - this also allows the API to be backwards compatible.  Reverted the commit that removed unit and added name and prettyName to TierBlockJson, added back prettyName functionality to DefaultUnit, added units to CatalogJson.  🌴


pierre:
👍

PR #779, opened 2017-07-25
failing catalog

anatoly-kudrevatykh-hs:
Hello,

We are experiencing account parked issue with our catalog. Here is an integration test to demonstrate the steps to reproduce.

The default test catalog was replaced with a simplified version of our catalog and TestSubscription `testChangeOfPlan` test was modified:

mvn clean install -Pmysql -Dtest=TestSubscription surefire:test

This also can be reproduced with Kaui by creating a user with default external payment, subscribing that user to FREE plan, then swiching to PROFESSIONAL_PLAN and then to "TEAM3S".

Sorry for unreadable diff, couldn't figure out how to load my catalog with ` XMLLoader`


pierre:
Created https://github.com/killbill/killbill/issues/783 for tracking.

Issue #767, opened 2017-07-06
Built-in support for retry in Kill Bill plugin apis 


sbrossie:
Overview: There are situations where plugins implementing a specific [plugin api](https://github.com/killbill/killbill-plugin-api) will fail to accomplish their task  -- e.g third party system integration returns an error-- and in such a situation, we would like to extend the core Kill Bill system to retry the operation.

Except for plugins implementing the `PaymentPluginApi`, which already offers some [retry options](http://docs.killbill.io/latest/userguide_subscription.html#_payment_retries) through system properties, plugin implementors have 2 options today:
1. Pass any kind of exception back to Kill Bill core which will abort the ongoing operation
2. Catch all exceptions in the plugin -- and possibly handle the retry mechanism on its own

Option 1 is probably sufficient for most operations that are a result of an client api call: The call would fail and the client could retry later when the service -- as a whole -- becomes fully available. However for all operations that are driven through bus/notification events, this is not a good option because the system will abort the call, but this may leave the system in a bad state -- e.g invoice generation fails and from then on, customer is not invoiced anymore.

Option 2 works, but the same kind of retry mechanism possibly needs to happen on all plugins.

The goal of this task is to allow for option 2 to happen with minimum logic inside the plugins.

sbrossie:
See [bug report in avatax-plugin](https://github.com/killbill/killbill-avatax-plugin/issues/3)

sbrossie:
As a workaround until this is fleshed out -- and implemented -- one solution is to implement that logic in the plugin, possibly by leveraging the KB  notificationQ mechanism. See [wiki for direction](https://github.com/killbill/killbill/wiki/Using-NotificationQueue-From-Plugin)

pierre:
We want to retry operations invoking plugins in async threads (bus, notifications), while throwing the exception back to the user for synchronous operations (e.g. API call).

Taking a step back, it sounds like we need to instead implement a retry mechanism to bus handlers. In https://github.com/killbill/killbill/issues/348#issuecomment-225037927, we argued otherwise, because we assumed that if Kill Bill core throws an NPE, there is nothing we can do.

A better approach would actually be to always retry (e.g. 5m, 1hr then every 24hrs) so that next time Kill Bill or the plugin is deployed (or if the third party system goes back online), the state is fixed automagically.

pierre:
An initial generic implementation has been written in the branch https://github.com/killbill/killbill/tree/fix-for-348, but it is more complicated than initially thought:

[...]

PR #785, opened 2017-08-03
Built-in support for retry in Kill Bill plugin apis

pierre:
See discussion at https://github.com/killbill/killbill/issues/767.


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoicePlugin.java
```diff
@@ -0,0 +1,265 @@
...
+
+        // Invoice failed to generate
+        assertEquals(invoiceUserApi.getInvoicesByAccount(account.getId(), false, callContext).size(), 2);
+
+        // Verify notification has moved to the retry service
+        checkRetryNotifications("2012-06-01T00:05:00");
+
+        // Add 5'
+        clock.addDeltaFromReality(5 * 60 * 1000);
+        checkRetryNotifications("2012-06-01T00:15:00");
```
sbrossie: Nit: Are we sure this date is exact -- test flakiness ?

util/pom.xml
```diff
@@ -210,6 +210,10 @@
         </dependency>
         <dependency>
             <groupId>org.kill-bill.billing.plugin</groupId>
+            <artifactId>killbill-plugin-api-invoice</artifactId>
+        </dependency>
```
sbrossie: I am guessing we need this because of `InvoicePluginApiRetryException` ?

util/src/main/java/org/killbill/billing/util/listener/RetryableService.java
```diff
@@ -0,0 +1,157 @@
...
+        try {
+            final NotificationEvent retryNotificationEvent = new RetryNotificationEvent(objectMapper.writeValueAsString(originalNotificationEvent), originalNotificationEvent.getClass(), originalEffectiveDate, retryNb);
+            retryNotificationQueue.recordFutureNotification(effectiveDate, retryNotificationEvent, context.getUserToken(), context.getAccountRecordId(), context.getTenantRecordId());
+        } catch (final IOException e) {
+            log.error("Unable to schedule retry for event='{}', effectiveDate='{}'", originalNotificationEvent, effectiveDate, e);
+        }
+    }
+
+    private DateTime computeRetryDate(final InvoicePluginApiRetryException invoicePluginApiRetryException, final DateTime initialEventDateTime, final int retryNb) {
+        final List<Period> retrySchedule = invoicePluginApiRetryException.getRetrySchedule();
```
sbrossie: Yeah, i think that's nice and flexible.

util/src/main/java/org/killbill/billing/util/listener/RetryableSubscriber.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        @Override
+        public void handleReadyNotification(final NotificationEvent eventJson, final DateTime eventDateTime, final UUID userToken, final Long searchKey1, final Long searchKey2) {
+            if (!(eventJson instanceof SubscriberNotificationEvent)) {
+                log.error("SubscriberQueueHandler received an unexpected event className='{}'", eventJson.getClass());
+            } else {
+                final BusEvent busEvent = ((SubscriberNotificationEvent) eventJson).getBusEvent();
+
+                final ImmutableSet<Class<?>> eventTypes = FLATTEN_HIERARCHY_CACHE.getUnchecked(busEvent.getClass());
```
sbrossie: In our use case, we could even consider simplifying the code, because i don't think we have use cases where handlers are `@Subscribing`  to base class/super-interface events.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoicePlugin.java
pierre: Yes: the delay is added to the initial notification effective date, which is aligned with the reference time of the account in this case (created right after we set the date to `new LocalDate(2012, 4, 1)`).

For bus events, since the initial effective date is computed as `clock.getUTCNow()`, we cannot do this though.

util/pom.xml
pierre: Yes.

util/src/main/java/org/killbill/billing/util/listener/RetryableSubscriber.java
pierre: We actually cannot: `InvoiceListener#handleSubscriptionTransition` subscribes to `EffectiveSubscriptionInternalEvent` for instance and the event on the bus is a `DefaultEffectiveSubscriptionEvent`. So we would need to go one level up at least.

sbrossie: Indeed:

```
 @SuppressWarnings("RedundantTypeArguments")
    private Set<Class<?>> foo(Class<?> concreteClass) {
        return (Set<Class<?>>) TypeToken.of(concreteClass).getTypes().rawTypes();
    }

   Set<Class<?>> result = foo(DefaultEffectiveSubscriptionEvent.class);
```

Produces:
```
0 = {java.lang.Class@5159} "class org.killbill.billing.subscription.api.user.DefaultEffectiveSubscriptionEvent"
1 = {java.lang.Class@5160} "class org.killbill.billing.subscription.api.user.DefaultSubscriptionEvent"
2 = {java.lang.Class@5161} "interface org.killbill.billing.events.EffectiveSubscriptionInternalEvent"
3 = {java.lang.Class@5162} "interface org.killbill.billing.events.SubscriptionInternalEvent"
4 = {java.lang.Class@5163} "interface org.killbill.billing.events.BusInternalEvent"
5 = {java.lang.Class@5164} "class org.killbill.billing.events.BusEventBase"
6 = {java.lang.Class@5165} "interface org.killbill.bus.api.BusEvent"
7 = {java.lang.Class@5166} "interface org.killbill.queue.api.QueueEvent"
8 = {java.lang.Class@331} "class java.lang.Object"
```

invoice/src/main/java/org/killbill/billing/invoice/InvoiceListener.java
```diff
@@ -31,114 +31,187 @@
...
+                           final NotificationQueueService notificationQueueService,
+                           final Clock clock) {
+        super(notificationQueueService);
         this.dispatcher = dispatcher;
-        this.invoiceConfig = invoiceConfig;
         this.internalCallContextFactory = internalCallContextFactory;
-        this.clock = clock;
         this.invoiceApi = invoiceApi;
+
+        subscriberQueueHandler.subscribe(EffectiveSubscriptionInternalEvent.class,
```
sbrossie: Do we really need the retry layer from Kill Bill internal events ? Could we just plug that for plugins (the piece in platform `KillbillEventRetriableBusHandler`) ?

pierre: These are separate pieces:

* `KillbillEventRetriableBusHandler` only takes care of retries for `ExtBusEvent` (notification plugins)
* The subscriber here is for invoice plugins (e.g. tax plugins) retries (original piece I worked on): the handler calls `processSubscriptionForInvoiceGeneration`, which eventually calls `invoicePluginDispatcher.getAdditionalInvoiceItems` (in `processAccountWithLockAndInputTargetDate`): this is the piece that can throw an `InvoicePluginApiRetryException`

Does that answer your question?

PR #787, opened 2017-08-09
Fix for analytics issue #69

beccagaspard:
Fix for analytics issue #69 - npe was thrown when account had a system tag.  Updated TagUserApi to optionally return system tags.


sbrossie:
As discussed closing PR because we decided to fix it [this way](https://github.com/killbill/killbill/commit/7120cfd1bf4861489f2ee46f8b9fc6231bacf877).

PR #804, opened 2017-09-29
invoice: make sure to always record future notifications

pierre:
In case of exception when processing a new invoice (e.g. thrown by plugin),
make sure to record a notification regardless, to make sure we never stop
invoicing that account.

For review, the easiest is either to read the new code directly or to ignore whitespaces in the diff (`?w=1`).


sbrossie:
👍 

(I had to look at the diff using `?w=1` and pulling the code 😸

pierre:
@sbrossie Just to double check, ok to merge?

PR #805, opened 2017-10-03
Make tests less flaky

pierre:


.circleci/config.yml
```diff
@@ -0,0 +1,219 @@
...
+      - image: killbill/kbbuild:0.1.0
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-dependencies-{{ .Branch }}-{{ checksum "pom.xml" }}
+      - run:
+          name: Setup dependencies
+          command: |
+            if [ "${CIRCLE_BRANCH}" != "master" ]; then
+              for i in killbill-oss-parent killbill-api killbill-plugin-api killbill-commons killbill-platform; do
```
sbrossie: I think you might need the `killbill-client-java` as well?

.circleci/config.yml
```diff
@@ -0,0 +1,219 @@
...
+      - image: killbill/mariadb:0.18
+        environment:
+        - MYSQL_ROOT_PASSWORD=root
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-dependencies-{{ .Branch }}-{{ checksum "pom.xml" }}
+      - run:
+          name: Setup latest DDL
+          command: ./bin/db-helper -a create --driver mysql -u root -p root -t yes -h 127.0.0.1
```
sbrossie: I suppose we need that because of `work-for-release-0.19.x` branch -- in that case we don't really need image `mariadb:0.18`, we just need a default `mariadb` image?

.circleci/config.yml
```diff
@@ -0,0 +1,219 @@
...
+              if [ -n "$(git ls-remote --heads https://github.com/killbill/killbill-integration-tests.git ${CIRCLE_BRANCH})" ]; then
+                git checkout -b ${CIRCLE_BRANCH} origin/${CIRCLE_BRANCH}
+              fi
+            fi
+            bundle install
+
+            count=0
+            until $(curl --output /dev/null --silent --fail http://127.0.0.1:8080/1.0/healthcheck); do
+              if [[ "$count" == "25" ]]; then
+                exit 1
```
sbrossie: Should we log something in that case -- killbill did not start in the 25 * 5 sec we allocated?

util/src/test/java/org/killbill/billing/DBTestingHelper.java
```diff
@@ -79,14 +79,16 @@ protected synchronized void executePostStartupScripts() throws IOException {
                                "CREATE TABLE accounts (\n" +
```
sbrossie: We should brainstorm to see if we can't get rid of that: Having this hardcoded tables in our tests that end up diverging from DLL is quite painful -- and error prone -- to maintain.

util/src/test/java/org/killbill/billing/GuicyKillbillTestSuite.java
```diff
@@ -143,6 +148,24 @@ public void afterMethodAlwaysRun(final Method method, final ITestResult result)
         }
     }
 
+    // Note: assertions should not be run in before / after hooks, as the associated test result won't be correctly updated.
+    // Use this wrapper instead.
+    @Override
+    public void run(final IHookCallBack callBack, final ITestResult testResult) {
+        // Make sure we start with a clean state
+        assertListenerStatus();
```
sbrossie: Ok, i see, now that makes sense ;-)

util/src/test/java/org/killbill/billing/api/FlakyInvokedMethodListener.java
```diff
@@ -0,0 +1,47 @@
...
+        if (testResult.getStatus() != ITestResult.FAILURE) {
+            return;
+        }
+
+        final IRetryAnalyzer retryAnalyzer = testResult.getMethod().getRetryAnalyzer();
+        if (retryAnalyzer != null &&
+            retryAnalyzer instanceof FlakyRetryAnalyzer &&
+            !((FlakyRetryAnalyzer) retryAnalyzer).shouldRetry()) {
+            // Don't fail the build (flaky test), mark it as SKIPPED
+            testResult.setStatus(ITestResult.SKIP);
```
sbrossie: Ok, that's a good way to handle flaky tests

.circleci/config.yml
pierre: Ah yes, good point, I'll add it.

.circleci/config.yml
pierre: We actually also need it for the test DDL (`-t` option).

.circleci/config.yml
pierre: Ack, I'll add it.

util/src/test/java/org/killbill/billing/DBTestingHelper.java
pierre: IIRC, we need it today because `util` relies on DDL provided by "downstream" modules (like account). The dependency is backward.

To get rid of it, we would probably need to rewrite some of the tests. I could look into it.

PR #831, opened 2018-01-01
fix NullPointerException generating invoice when there is an item adjustment for a fixed item

almalki:
Happy new year.
I tried to fix the issue and came up with this fix, 
which actually fixed the issue for me and new invoices are now generated.


pierre:
> Happy new year.

Happy New Year!

> I tried to fix the issue and came up with this fix, which actually fixed the issue for me and new invoices are now generated.

Thank you for the debugging work. The patch looks good but we need a test for it. I figured it would be easier for me to write it since such low-level tests can be hard to setup. See https://github.com/killbill/killbill/pull/832.

almalki:
@pierre  Thank you

PR #832, opened 2018-01-02
fix NullPointerException generating invoice when there is an item adjustment for a fixed item

pierre:
Follow-up on https://github.com/killbill/killbill/pull/831.


sbrossie:
👍

sbrossie:
@pierre Has it been merged to `work-for-release-0.19.x` ?

pierre:
Not yet, it's only in `master` currently.

PR #834, opened 2018-01-05
Payment Method Refresh Fix

beccagaspard:
Fix for "org.killbill.billing.payment.api.PaymentApiException: Internal payment error : Payment method xxx does not exist" when a default payment method is removed and a different payment method is selected as default during a payment method refresh.  Payment method has already been deleted by the time we get to this point, so we need to include deleted payment methods when retrieving old payment method.


sbrossie:
Yep, that makes sense. 👍

PR #835, opened 2018-01-08
Payment Method Refresh / Account Payment Method fix

beccagaspard:
Fix for issue where account paymentMethodId would not be updated when a payment method refresh removed the default payment method and did not specify a new default payment method.  Removing unnecessary null check since we are already checking if the current (old) default payment method is from the same plugin before the modifying the account's payment method id (so we should never overwrite an account's payment method id if that payment method is from a different plugin.)


sbrossie:
👍

Issue #837, opened 2018-01-10
Create new api to attach/retrieve a tag or a custom field for `invoice items`


sbrossie:
For most of our resources, we have a way to manage (create/list/update/delete) tags or custom fields -- see for instance [apis for subscriptions](https://github.com/killbill/killbill/blob/killbill-0.19.0/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java#L882).

For invoice items, we don't have a resource, so such apis don't exist. The goal would be to create a new resource for invoice items and export similar apis.

PR #838, opened 2018-01-16
Create new api to attach/retrieve a tag or a custom field for `invoice items` #837

wwjbatista:


pierre:
> Looks good, might be good to add some test -- and expose the functionality in the java client.

👍

Issue #839, opened 2018-01-17
Add usage details in items


sbrossie:
The catalog allows to configure tier usages -- see example [here](https://github.com/killbill/killbill/blob/killbill-0.19.0/catalog/src/test/resources/UsageExperimental.xml#L250)-- and based on such tier definitions and reported usage, the invoicing system computes the aggregate amount for the period. This results in one `USAGE` item on the invoice.

In many cases, the detail about usage amount * tier price needs to be exposed to the customer, and the task is to allow to capture this detail.

sbrossie:
The high level proposal is the following:

1. Add a new [invoice configuration parameter](https://github.com/killbill/killbill/blob/killbill-0.19.0/util/src/main/java/org/killbill/billing/util/config/definition/InvoiceConfig.java) to control whether we want to keep current behavior (aggregate mode) or add one USAGE item for each tier level (detail mode). The current aggregate mode remains default.
2. In the aggregate mode, add a new field `item_details` in our `invoice_items` [table](https://github.com/killbill/killbill/blob/killbill-0.19.0/invoice/src/main/resources/org/killbill/billing/invoice/ddl.sql#L4), and populate this field with a json list that contains one entry per tier-- e.g `{[{'tier': 1, 'tierPrice': 10, 'quantity': 5, 'amount': 50}, {'tier': 2, 'tierPrice': 12, 'quantity': 2, 'amount': 24}]}`
3. In the detail mode, add a new field `quantity`  in our `invoice_items` table and populate it using the number of unit for each tier. Use the `rate` field to write the price for each tier. In this mode, we don't need to output any json. This mode is convenient for when we want to run queries on such data.

Based on the usage type, the computation of the usage can be found in the following methods:

- [Capacity in arrear](https://github.com/killbill/killbill/blob/killbill-0.19.0/invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java#L364)
- [Consumable in arrear, ALL TIERS mode](https://github.com/killbill/killbill/blob/killbill-0.19.0/invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java#L413)
- [Consumable in arrear, TOP TIER mode](https://github.com/killbill/killbill/blob/killbill-0.19.0/invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java#L434)

[...]

PR #849, opened 2018-01-25
 Add usage details in items #839

wwjbatista:
https://github.com/killbill/killbill-api/pull/36


invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +569,54 @@ public LocalDate getNextNotificationDate() {
...
+        for (ToBeBilledConsumableInArrearDetail toBeBilled: toBeBilledDetails){
+            result = result.add(toBeBilled.getAmount());
+        }
+        return result;
+    }
+
+    private String toJson(List<ToBeBilledConsumableInArrearDetail> toBeBilledConsumableInArrearDetails) throws CatalogApiException {
+        String result = null;
+        if (toBeBilledConsumableInArrearDetails != null && toBeBilledConsumableInArrearDetails.size() > 0){
+            ObjectMapper objectMapper = new ObjectMapper();
```
pierre: This can safely be global (`private static final`).

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +569,54 @@ public LocalDate getNextNotificationDate() {
...
+            try {
+                result = objectMapper.writeValueAsString(toBeBilledConsumableInArrearDetails);
+            } catch (JsonProcessingException e) {
+                throw new CatalogApiException(e, ErrorCode.__UNKNOWN_ERROR_CODE);
```
pierre: Maybe `IllegalStateException` instead, as it isn't really a catalog exception.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -184,21 +188,24 @@ public UsageInArrearItemsAndNextNotificationDate computeMissingItemsAndNextNotif
...
                     if (!unitTypes.contains(cur.getUnitType())) {
                         log.warn("ContiguousIntervalConsumableInArrear is skipping unitType " + cur.getUnitType());
                         continue;
                     }
 
-                    final BigDecimal toBeBilledForUnit = computeToBeBilledConsumableInArrear(cur);
-                    toBeBilledUsage = toBeBilledUsage.add(toBeBilledForUnit);
+                    toBeBilledUsageDetails.addAll(computeToBeBilledConsumableInArrear(cur, tierNum++));
                 }
+                toBeBilledUsage = toBeBilledUsage.add(toBeBilledForUnit(toBeBilledUsageDetails));
```
pierre: `toBeBilledUsage.add(...)`  isn't needed anymore here, right (`toBeBilledUsage == 0`)?

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -207,9 +214,12 @@ public UsageInArrearItemsAndNextNotificationDate computeMissingItemsAndNextNotif
             // Compare the two and add the missing piece if required.
             if (!billedItems.iterator().hasNext() || billedUsage.compareTo(toBeBilledUsage) < 0) {
                 final BigDecimal amountToBill = toBeBilledUsage.subtract(billedUsage);
+                if (billedUsage.compareTo(BigDecimal.ZERO) > 0){
+                    toBeBilledUsageDetails.add(new ToBeBilledConsumableInArrearDetail(-1, "billedUsage", billedUsage, -1));
```
pierre: FYI, for the full implementation, I'm not sure we can take such a shortcut unfortunately, as with `DETAIL_MODE`, we might be missing entries for each individual tier.

I believe instead of `computeBilledUsage`, we will need to compare existing vs _proposed_ (to be billed) items for each tier.

sbrossie: I am confused about this section: Are we just trying to add some 'details' in the json for the already billed items?

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -184,21 +188,24 @@ public UsageInArrearItemsAndNextNotificationDate computeMissingItemsAndNextNotif
         final List<RolledUpUsage> allUsage = getRolledUpUsage();
         for (final RolledUpUsage ru : allUsage) {
 
+            List<ToBeBilledConsumableInArrearDetail> toBeBilledUsageDetails = Lists.newLinkedList();
             BigDecimal toBeBilledUsage = BigDecimal.ZERO;
             if (usage.getUsageType() == UsageType.CAPACITY) {
                 toBeBilledUsage = computeToBeBilledCapacityInArrear(ru.getRolledUpUnits());
```
sbrossie: Should we also add the details for `UsageType.CAPACITY`. This is simpler since there will be only one item, but writing the detail about the chosen tier might be useful.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -426,12 +437,12 @@ BigDecimal computeToBeBilledConsumableInArrearWith_ALL_TIERS(final List<TieredBl
...
             }
-            result = result.add(tieredBlock.getPrice().getPrice(getCurrency()).multiply(new BigDecimal(nbUsedTierBlocks)));
+            toBeBilledDetails.add(new ToBeBilledConsumableInArrearDetail(tierNum, tieredBlock.getUnit().getName(), tieredBlock.getPrice().getPrice(getCurrency()), nbUsedTierBlocks));
         }
-        return result;
+        return toBeBilledDetails;
     }
 
-    BigDecimal computeToBeBilledConsumableInArrearWith_TOP_TIER(final List<TieredBlock> tieredBlocks, final Long units) throws CatalogApiException {
+    ToBeBilledConsumableInArrearDetail computeToBeBilledConsumableInArrearWith_TOP_TIER(final List<TieredBlock> tieredBlocks, final Long units, int tierNum) throws CatalogApiException {
```
sbrossie: Same remark as `UsageType.CAPACITY`: Should we also add the details for `UsageType.CONSUMABLE` and `TOP_TIER`: This is simpler since there will be only one item, but writing the detail about the chosen tier might be useful.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
wwjbatista: @sbrossie, yes it is for the already billed. Is that when I added the detail, the sum is not equal to the total amount of the item, since there could have billed items. Hence the add of a new billed detail on the JSON. But if it is not necessary I will remove it.

sbrossie: @wwjbatista Ok, that's makes sense. Keep it the way you have it.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +588,142 @@ public LocalDate getNextNotificationDate() {
...
+            this.existingUsageAmount = existingUsageAmount;
+        }
+        public void setReference(String reference) {
+            this.reference = reference;
+        }
+
+        public List<ConsumableInArrearDetail> reconcile(List<ConsumableInArrearDetail> billedUsageItemDetails){
+            List<ConsumableInArrearDetail> unreconciledUsage = Lists.newLinkedList();
+            for (ConsumableInArrearDetail billedUsageDetail : billedUsageItemDetails){
+                if( tierUnit == billedUsageDetail.tierUnit ) {
```
pierre: ⚠️ Use `equals` for `String` comparison.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +588,142 @@ public LocalDate getNextNotificationDate() {
...
+
+    public BigDecimal toBeBilledForUnit(List<ConsumableInArrearDetail> toBeBilledDetails){
+        BigDecimal result = BigDecimal.ZERO;
+        for (ConsumableInArrearDetail toBeBilled: toBeBilledDetails){
+            result = result.add(toBeBilled.getAmount());
+        }
+        return result;
+    }
+
+    private List<ConsumableInArrearDetail> reconcileExistedBilledWithToBeBilled(Iterable<InvoiceItem> billedItems, List<ConsumableInArrearDetail> toBeBilledConsumableInArrearDetails) {
```
pierre: I think it would be worth adding a unit test (`fast` test) for this method.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -376,9 +402,12 @@ BigDecimal computeToBeBilledCapacityInArrear(final List<RolledUpUnit> roUnits) t
                     complies = false;
                     break;
                 }
+                toBeBilledDetails.add(new ConsumableInArrearDetail(tierNum++, ro.getUnitType(), cur.getRecurringPrice().getPrice(getCurrency()), ro.getAmount().intValue(), BigDecimal.ZERO, null, null));
```
pierre: I am wondering if this actually should be outside of that loop, in its own separate loop: if there are three different units, the first two comply but not the third one, we would add these details nonetheless (instead of adding them only for the correct tier).

It would be good to verify this behavior in `TestContiguousIntervalCapacityInArrear`.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +588,142 @@ public LocalDate getNextNotificationDate() {
...
+                }
+            } else {
+                toBeBilledConsumableInArrearDetails.add(new ConsumableInArrearDetail(bi.getRate(), bi.getQuantity(), bi.getAmount().negate(), bi.getId().toString()));
+            }
+        }
+
+        return toBeBilledConsumableInArrearDetails;
+    }
+
+    public static final String toJson(List<ConsumableInArrearDetail> toBeBilledConsumableInArrearDetails) {
```
pierre: Could be a private method.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +588,142 @@ public LocalDate getNextNotificationDate() {
...
+        String result = null;
+        if (toBeBilledConsumableInArrearDetails != null && toBeBilledConsumableInArrearDetails.size() > 0){
+            ObjectMapper objectMapper = new ObjectMapper();
```
pierre: Should be shared (`private static final`) in the class.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +588,142 @@ public LocalDate getNextNotificationDate() {
...
+            try {
+                result = objectMapper.writeValueAsString(toBeBilledConsumableInArrearDetails);
+            } catch (JsonProcessingException e) {
+                Preconditions.checkState(false, e.getMessage());
+            }
+        }
+        return result;
+    }
+
+    public static final List<ConsumableInArrearDetail> fromJson(String itemDetails){
```
pierre: Could be a private method.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +588,142 @@ public LocalDate getNextNotificationDate() {
...
+        List<ConsumableInArrearDetail> toBeBilledConsumableInArrearDetails = null;
+        if (itemDetails != null){
+            ObjectMapper objectMapper = new ObjectMapper();
```
pierre: Use shared instance.

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
```diff
@@ -303,6 +319,15 @@ public boolean apply(final InvoiceItem input) {
         assertTrue(result.get(0).getStartDate().compareTo(startDate) == 0);
         assertTrue(result.get(0).getEndDate().compareTo(firstBCDDate) == 0);
 
+        // check item detail
+        List<MockToBeBilledConsumableInArrearDetail> itemDetails = objectMapper.reader()
+                                                                 .forType(new TypeReference<List<MockToBeBilledConsumableInArrearDetail>>() {})
```
pierre: You could simplify it:

```
objectMapper.readValue(result.get(0).getItemDetails(), new TypeReference<List<MockToBeBilledConsumableInArrearDetail>>() {})
```

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/InvoiceItemJson.java
```diff
@@ -85,7 +88,10 @@ public InvoiceItemJson(@JsonProperty("invoiceItemId") final String invoiceItemId
                            @JsonProperty("startDate") final LocalDate startDate,
                            @JsonProperty("endDate") final LocalDate endDate,
                            @JsonProperty("amount") final BigDecimal amount,
+                           @JsonProperty("rate") final  BigDecimal rate,
```
pierre: The Java client will need to be updated.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -196,21 +209,33 @@ public UsageInArrearItemsAndNextNotificationDate computeMissingItemsAndNextNotif
...
             if (!billedItems.iterator().hasNext() || billedUsage.compareTo(toBeBilledUsage) < 0) {
-                final BigDecimal amountToBill = toBeBilledUsage.subtract(billedUsage);
+                toBeBilledUsageDetails = reconcileExistedBilledWithToBeBilled(billedItems, toBeBilledUsageDetails);
+                final BigDecimal amountToBill = toBeBilledForUnit(toBeBilledUsageDetails);
+
                 if (amountToBill.compareTo(BigDecimal.ZERO) > 0) {
-                    final InvoiceItem item = new UsageInvoiceItem(invoiceId, accountId, getBundleId(), getSubscriptionId(), getPlanName(),
-                                                                  getPhaseName(), usage.getName(), ru.getStart(), ru.getEnd(), amountToBill, getCurrency());
-                    result.add(item);
+                    if (UsageDetailMode.DETAIL.compareTo(usageDetailMode) == 0){
```
pierre: You could simplify it: `UsageDetailMode.DETAIL == usageDetailMode`.

invoice/src/main/java/org/killbill/billing/invoice/config/MultiTenantInvoiceConfig.java
```diff
@@ -162,6 +163,29 @@ public boolean isInvoicingSystemEnabled(final InternalTenantContext tenantContex
         return isInvoicingSystemEnabled();
     }
 
+    @Override
+    public UsageDetailMode getItemResultBehaviorMode() {
+        final UsageDetailMode mode = staticConfig.getItemResultBehaviorMode();
+        if (mode.compareTo(UsageDetailMode.AGGREGATE) != 0 || mode.compareTo(UsageDetailMode.DETAIL) != 0) {
```
pierre: You could simplify it (use `==`).

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
wwjbatista: Using the `TestContiguousIntervalCapacityInArrear`,  the details that is been added (first result) is 

> [{"tier":1, "tierUnit":"unit1", "tierPrice":10, "quantity":100, "reference":null, "existingUsageAmount":null, "amount":0},{"tier":2, "tierUnit":"unit2", "tierPrice":10, "quantity":1000, "reference":null, "existingUsageAmount":null, "amount":10}]

I modified the test to add other unit and this is the detail that produces:
> [{"tier":1, "tierUnit":"unit1", "tierPrice":10, "quantity":100, "reference":null, "existingUsageAmount":null, "amount":0},{"tier":2, "tierUnit":"unit2", "tierPrice":10, "quantity":1000, "reference":null, "existingUsageAmount":null, "amount":0},
{"tier":3, "tierUnit":"unit3", "tierPrice":10, "quantity":50, "reference":null, "existingUsageAmount":null, "amount":10}]

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -557,4 +587,160 @@ public LocalDate getNextNotificationDate() {
...
+                toBeBilledUsageInArrearDetails = objectMapper.readValue(itemDetails, new TypeReference<List<UsageInArrearDetail>>() {});
+            } catch (IOException e) {
+                Preconditions.checkState(false, e.getMessage());
+            }
+        }
+
+        return toBeBilledUsageInArrearDetails;
+    }
+
+    // divide the price between all details to prevent a detail to be removed if the price is zero when the usage mode is detail
```
pierre: We cannot do that: even in `detail` mode, a single entry would be returned for CAPACITY. See also my comment at https://github.com/killbill/killbill-integration-tests/pull/13#discussion_r166228035.

pierre:
FYI killbill-oss-parent 0.141.27 with killbill-client 0.41.6 released. If you update your `pom.xml`, we would be able to verify CI builds.

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalCapacityInArrear.java
```diff
@@ -243,6 +246,224 @@ public boolean apply(final InvoiceItem input) {
...
+        assertEquals(result.get(1).getUsageName(), usage.getName());
+        assertTrue(result.get(1).getStartDate().compareTo(firstBCDDate) == 0);
+        assertTrue(result.get(1).getEndDate().compareTo(secondBCDDate) == 0);
+        // check item detail
+        List<UsageInArrearDetail> itemDetails2 = objectMapper.readValue(result.get(1).getItemDetails(), new TypeReference<List<UsageInArrearDetail>>() {});
+        assertEquals(itemDetails2.size(), 2);
+        assertEquals(itemDetails2.get(0).getTier(), 2);
+        assertEquals(itemDetails2.get(0).getQuantity().intValue(), 101);
+        assertEquals(itemDetails2.get(0).getTierPrice().compareTo(BigDecimal.TEN), 0);
+        assertEquals(itemDetails2.get(1).getTier(), 2);
```
sbrossie: I think this should be tier `1` -- as detailed per [here](https://github.com/killbill/killbill-integration-tests/pull/13/files#r166222783)

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalCapacityInArrear.java
```diff
@@ -243,6 +246,224 @@ public boolean apply(final InvoiceItem input) {
...
+        assertEquals(result.get(2).getSubscriptionId(), subscriptionId);
+        assertEquals(result.get(2).getPlanName(), planName);
+        assertEquals(result.get(2).getPhaseName(), phaseName);
+        assertEquals(result.get(2).getUsageName(), usage.getName());
+        assertTrue(result.get(2).getStartDate().compareTo(secondBCDDate) == 0);
+        assertTrue(result.get(2).getEndDate().compareTo(endDate) == 0);
+        // check item detail
+        List<UsageInArrearDetail> itemDetails3 = objectMapper.readValue(result.get(2).getItemDetails(), new TypeReference<List<UsageInArrearDetail>>() {});
+        assertEquals(itemDetails3.size(), 2);
+        assertEquals(itemDetails3.get(0).getTier(), 3);
```
sbrossie: Same remark here, should be tier `2`. 

Also could we add assertion on `tierUnit` (e.g `FOO` v.s `BAR`). Not only would be good for the test but to follow the test assertion -- not easy to understand if element `0` v.s `1` is `FOO` or `BAR`. And are we sure this is deterministic and all iterations of tests will produce same results?

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalCapacityInArrear.java
```diff
@@ -241,6 +248,135 @@ public boolean apply(final InvoiceItem input) {
...
+
+        itemDetails = objectMapper.readValue(result.get(0).getItemDetails(), new TypeReference<List<UsageInArrearDetail>>() {});
+        // BAR item detail
+        assertEquals(itemDetails.get(0).getTierUnit(), "BAR");
+        assertEquals(itemDetails.get(0).getTier(), 2);
+        assertEquals(itemDetails.get(0).getQuantity().intValue(), 101);
+        assertEquals(itemDetails.get(0).getTierPrice().compareTo(BigDecimal.TEN), 0);
+        // FOO item detail
+        assertEquals(itemDetails.get(1).getTierUnit(), "FOO");
+        assertEquals(itemDetails.get(1).getTier(), 2);
```
pierre: @sbrossie Same question as https://github.com/killbill/killbill-integration-tests/pull/13#discussion_r166626676

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
```diff
@@ -407,12 +436,373 @@ public void testWithRawUsageStartDateAfterEndDate() throws CatalogApiException {
...
+        assertEquals(result.get(0).getQuantity().intValue(), 101);
+        assertEquals(result.get(0).getRate().compareTo(new BigDecimal("20.0")), 0);
+        // FOO: 5 * 1 = 5
+        assertEquals(result.get(1).getAmount().compareTo(new BigDecimal("5")), 0);
+        assertEquals(result.get(1).getQuantity().intValue(), 5);
+        assertEquals(result.get(1).getRate().compareTo(BigDecimal.ONE), 0);
+
+        // Case 3
+        rawUsages = new ArrayList<RawUsage>();
+        rawUsages.add(new DefaultRawUsage(subscriptionId, new LocalDate(2014, 03, 20), "FOO", 75L));
```
pierre: Typo: should be 76.

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
```diff
@@ -407,12 +436,373 @@ public void testWithRawUsageStartDateAfterEndDate() throws CatalogApiException {
...
+        // FOO: 5 * 1 = 5
+        assertEquals(itemDetails.get(1).getTierUnit(), "FOO");
+        assertEquals(itemDetails.get(1).getTier(), 1);
+        assertEquals(itemDetails.get(1).getAmount().compareTo(new BigDecimal("5")), 0);
+        assertEquals(itemDetails.get(1).getQuantity().intValue(), 5);
+        assertEquals(itemDetails.get(1).getTierPrice().compareTo(BigDecimal.ONE), 0);
+
+        // Case 3
+        rawUsages = new ArrayList<RawUsage>();
+        rawUsages.add(new DefaultRawUsage(subscriptionId, new LocalDate(2014, 03, 20), "FOO", 75L));
```
pierre: Typo: should be 76.

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
```diff
@@ -407,12 +436,373 @@ public void testWithRawUsageStartDateAfterEndDate() throws CatalogApiException {
...
+    public void testComputeMissingItemsAggregateModeAllTier() throws CatalogApiException, IOException {
+
+        // Case 1
+        List<RawUsage> rawUsages = new ArrayList<RawUsage>();
+        rawUsages.add(new DefaultRawUsage(subscriptionId, new LocalDate(2014, 03, 20), "FOO", 5L));
+        rawUsages.add(new DefaultRawUsage(subscriptionId, new LocalDate(2014, 03, 21), "BAR", 99L));
+
+        List<InvoiceItem> result = produceInvoiceItems(rawUsages, TierBlockPolicy.ALL_TIERS, UsageDetailMode.AGGREGATE);
+        assertEquals(result.size(), 1);
+        assertEquals(result.get(0).getAmount().compareTo(new BigDecimal("203")),0);
```
pierre: @sbrossie We indeed expect one item here, right? It looks like our test plan was incorrect/confusing for this.

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalCapacityInArrear.java
sbrossie: @pierre Agreed

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
sbrossie: @pierre Yep, our test plan does not seem to match current code where we have one item across unit types. Should i update the test plan?

pierre: @sbrossie Yes, updating the test plan in the ticket would be nice, if we go back to it one day...

pierre: @sbrossie Also, it probably means we should enhance in a second pass the test plan, where we have multiple units in multiple usage sections...

PR #863, opened 2018-02-10
Rework for usage details and refactor capacity/consumable logic

sbrossie:


invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalConsumableUsageInArrear.java
```diff
@@ -0,0 +1,245 @@
...
+                                                     final LocalDate rawUsageStartDate,
+                                                     final UsageDetailMode usageDetailMode,
+                                                     final InternalTenantContext internalTenantContext) {
+        super(usage, accountId, invoiceId, rawSubscriptionUsage, targetDate, rawUsageStartDate, usageDetailMode, internalTenantContext);
+    }
+
+    @Override
+    protected void populateResults(final LocalDate startDate, final LocalDate endDate, final Iterable<InvoiceItem> billedItems, final BigDecimal billedUsage, final BigDecimal toBeBilledUsage, final UsageInArrearDetail toBeBilledUsageDetails, final boolean areAllBilledItemsWithDetails, final List<InvoiceItem> result) {
+        // Compute final amount by subtracting  amount that was already billed.
+        if (!billedItems.iterator().hasNext() || billedUsage.compareTo(toBeBilledUsage) < 0) {
```
pierre: Same remark regarding `!billedItems.iterator().hasNext()`.

Also, maybe `billedUsage.compareTo(toBeBilledUsage) < 0` should be extracted around these methods, since it's the same in both implementations:

```
if (billedUsage.compareTo(toBeBilledUsage) < 0) {
  populateResults(...);
}
```

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
```diff
@@ -0,0 +1,138 @@
...
+
+
+    @Override
+    protected void populateResults(final LocalDate startDate, final LocalDate endDate, final Iterable<InvoiceItem> billedItems, final BigDecimal billedUsage, final BigDecimal toBeBilledUsage, final UsageInArrearDetail toBeBilledUsageDetails, final boolean areAllBilledItemsWithDetails, final List<InvoiceItem> result) {
+        // Compute final amount by subtracting  amount that was already billed.
+        if (!billedItems.iterator().hasNext() || billedUsage.compareTo(toBeBilledUsage) < 0) {
+            final BigDecimal amountToBill = toBeBilledUsage.subtract(billedUsage);
+
+            if (amountToBill.compareTo(BigDecimal.ZERO) > 0) {
+                    final String itemDetails = areAllBilledItemsWithDetails ? toBeBilledUsageDetails.toJson(objectMapper) : null;
```
pierre: Why do we need to check `areAllBilledItemsWithDetails`? `toBeBilledUsageDetails` is independent of `areAllBilledItemsWithDetails` for capacity.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
```diff
@@ -0,0 +1,138 @@
...
+
+
+
+    @Override
+    protected void populateResults(final LocalDate startDate, final LocalDate endDate, final Iterable<InvoiceItem> billedItems, final BigDecimal billedUsage, final BigDecimal toBeBilledUsage, final UsageInArrearDetail toBeBilledUsageDetails, final boolean areAllBilledItemsWithDetails, final List<InvoiceItem> result) {
+        // Compute final amount by subtracting  amount that was already billed.
+        if (!billedItems.iterator().hasNext() || billedUsage.compareTo(toBeBilledUsage) < 0) {
+            final BigDecimal amountToBill = toBeBilledUsage.subtract(billedUsage);
+
+            if (amountToBill.compareTo(BigDecimal.ZERO) > 0) {
```
pierre: Do you need this check given we already verified that `billedUsage.compareTo(toBeBilledUsage) < 0`?

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalConsumableUsageInArrear.java
```diff
@@ -0,0 +1,245 @@
...
+
+            if (amountToBill.compareTo(BigDecimal.ZERO) > 0) {
+                if (UsageDetailMode.DETAIL == usageDetailMode) {
+                    for (UsageConsumableInArrearTierUnitDetail toBeBilledUsageDetail : ((UsageConsumableInArrearDetail) toBeBilledUsageDetails).getTierDetails()) {
+                        final InvoiceItem item = new UsageInvoiceItem(invoiceId, accountId, getBundleId(), getSubscriptionId(), getPlanName(),
+                                                                      getPhaseName(), usage.getName(), startDate, endDate, toBeBilledUsageDetail.getAmount(), toBeBilledUsageDetail.getTierPrice(), getCurrency(), toBeBilledUsageDetail.getQuantity(), null);
+                        result.add(item);
+                    }
+                } else {
+                    final String itemDetails = areAllBilledItemsWithDetails ? toBeBilledUsageDetails.toJson(objectMapper) : null;
```
pierre: Should we really check for `areAllBilledItemsWithDetails` here? Unlike `CAPACITY`, here `toBeBilledUsageDetails ` is dependent of `areAllBilledItemsWithDetails` but it's maybe better to always output it, rather than having random items with a `null` field. In other words, it's easier to explain why the invoice item doesn't match the details, rather than trying to figure out why we suddenly didn't populate it.

I believe it shouldn't affect further calculation: `areAllBilledItemsWithDetails` will always be false for that time period, even if we have further data points late in a subsequent invoice generation (because we have the very first item without the JSON populated): `amountToBill` calculation will then always revert to the computation of `toBeBilledUsage.subtract(billedUsage)` instead of relying on the JSON.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalConsumableUsageInArrear.java
```diff
@@ -0,0 +1,245 @@
...
+                                                     final UsageDetailMode usageDetailMode,
+                                                     final InternalTenantContext internalTenantContext) {
+        super(usage, accountId, invoiceId, rawSubscriptionUsage, targetDate, rawUsageStartDate, usageDetailMode, internalTenantContext);
+    }
+
+    @Override
+    protected void populateResults(final LocalDate startDate, final LocalDate endDate, final Iterable<InvoiceItem> billedItems, final BigDecimal billedUsage, final BigDecimal toBeBilledUsage, final UsageInArrearDetail toBeBilledUsageDetails, final boolean areAllBilledItemsWithDetails, final List<InvoiceItem> result) {
+        // Compute final amount by subtracting  amount that was already billed.
+        if (!billedItems.iterator().hasNext() || billedUsage.compareTo(toBeBilledUsage) < 0) {
+            final BigDecimal amountToBill = areAllBilledItemsWithDetails ? toBeBilledUsage : toBeBilledUsage.subtract(billedUsage);
```
pierre: Add a comment explaining that `getToBeBilledUsageDetails` took into account the `billedItems` already, since it's now the core of the reconciliation logic if I understand correctly.

invoice/src/main/java/org/killbill/billing/invoice/usage/UsageUtils.java
```diff
@@ -40,13 +38,19 @@
...
+            boolean found = false;
             for (TieredBlock tierBlock : tier.getTieredBlocks()) {
                 if (tierBlock.getUnit().getName().equals(unitType)) {
                     result.add(tierBlock);
+                    found = true;
+                    break;
                 }
             }
+            // We expect this method to return an ordered list of TieredBlock, each for each tier.
+            Preconditions.checkState(found, String.format("Catalog issue in usage section '%s': Missing tierBlock definition for unit '%s'", usage.getName(), unitType));
```
pierre: Nit: you don't need `String.format`, see following signature:

```
  /**
   * Ensures the truth of an expression involving the state of the calling instance, but not
   * involving any parameters to the calling method.
   *
   * @param expression a boolean expression
   * @param errorMessageTemplate a template for the exception message should the check fail. The
   *     message is formed by replacing each {@code %s} placeholder in the template with an
   *     argument. These are matched by position - the first {@code %s} gets {@code
   *     errorMessageArgs[0]}, etc. Unmatched arguments will be appended to the formatted message in
   *     square braces. Unmatched placeholders will be left as-is.
   * @param errorMessageArgs the arguments to be substituted into the message template. Arguments
   *     are converted to strings using {@link String#valueOf(Object)}.
   * @throws IllegalStateException if {@code expression} is false
   * @throws NullPointerException if the check fails and either {@code errorMessageTemplate} or
   *     {@code errorMessageArgs} is null (don't let this happen)
   */
  public static void checkState(
      boolean expression,
      @Nullable String errorMessageTemplate,
      @Nullable Object... errorMessageArgs) {
    if (!expression) {
      throw new IllegalStateException(format(errorMessageTemplate, errorMessageArgs));
    }
  }
```

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
```diff
@@ -65,6 +61,7 @@
 
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertTrue;
+import static org.testng.AssertJUnit.assertNull;
```
pierre: ⚠️ Wrong import

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
```diff
@@ -321,11 +318,8 @@ public boolean apply(final InvoiceItem input) {
         assertTrue(result.get(0).getStartDate().compareTo(startDate) == 0);
         assertTrue(result.get(0).getEndDate().compareTo(firstBCDDate) == 0);
 
-        // check item detail
-        List<UsageInArrearDetail> itemDetails = objectMapper.readValue(result.get(0).getItemDetails(), new TypeReference<List<UsageInArrearDetail>>() {});
-
-        assertEquals(itemDetails.size(), 1);
-        assertEquals(itemDetails.get(0).getAmount().compareTo(result.get(0).getAmount()),0);
+        // Because the existing items did not have any json (item details), the new one also don't have the details.
```
pierre: See comment above, behavior seems a bit strange to me.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
```diff
@@ -0,0 +1,138 @@
...
+                                                   final InternalTenantContext internalTenantContext) {
+        super(usage, accountId, invoiceId, rawSubscriptionUsage, targetDate, rawUsageStartDate, usageDetailMode, internalTenantContext);
+    }
+
+
+
+    @Override
+    protected void populateResults(final LocalDate startDate, final LocalDate endDate, final Iterable<InvoiceItem> billedItems, final BigDecimal billedUsage, final BigDecimal toBeBilledUsage, final UsageInArrearDetail toBeBilledUsageDetails, final boolean areAllBilledItemsWithDetails, final List<InvoiceItem> result) {
+        // Compute final amount by subtracting  amount that was already billed.
+        if (!billedItems.iterator().hasNext() || billedUsage.compareTo(toBeBilledUsage) < 0) {
```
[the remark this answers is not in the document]
sbrossie: Yep, agreed, i need to do another pass and re-read the code. I took care of the 90%.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
sbrossie: For now it is, but i think the discussion is open; we have to decide exactly which info we put in this json: If the indo we add is independent of any previously billed amount, -- which it is with my current implementation, what you are saying makes sense.  Now i went into the trouble of splitting the json between the 2 modes, with the intention to revisit this specifically. So will keep it the way it is for now, but more thoughts is needed here.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
```diff
@@ -0,0 +1,138 @@
...
+        return computeToBeBilledCapacityInArrear(rolledUpUnits);
+    }
+
+    private Limit getTierLimit(final Tier tier, final String unitType) {
+        for (final Limit cur : tier.getLimits()) {
+            if (cur.getUnit().getName().equals(unitType)) {
+                return cur;
+            }
+        }
+        Preconditions.checkState(false, "Could not find unit type " + unitType + " in usage tier ");
```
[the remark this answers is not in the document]
sbrossie: This code is not new -- just moved it -- but your point remain valid. I think we need a 'toString()' method in `DefaultTier` or remove it. Will add a TODO for me.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
```diff
@@ -0,0 +1,138 @@
...
+        return null;
+    }
+
+    @VisibleForTesting
+    UsageCapacityInArrearDetail computeToBeBilledCapacityInArrear(final List<RolledUpUnit> roUnits) throws CatalogApiException {
+        Preconditions.checkState(isBuilt.get());
```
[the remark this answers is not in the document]
sbrossie: Let's talk..

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalConsumableUsageInArrear.java
sbrossie: I think it's more tricky than it looks, moving that logic into `ContiguousIntervalUsageInArrear` would require reintroducing a bunch of `if mode = CONSUMABLE`, ...

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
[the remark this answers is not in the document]
sbrossie: See c4a2346a8dff89259f398529763f64502880edb0

However, we might need a hook to undo this behavior.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
sbrossie: Not sure i follow either: If we enter the `if` because there was no previous items to bill (`!billedItems.iterator().hasNext()`), then `amountToBill` = `toBeBilledUsage`, but we still have to check that `amountToBill` is > 0 ?

I 'd love to simplify this code path, hopefully i am missing something.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
sbrossie: So more context, current json for capacity is the following -- see new test i wrote with existing data usage for past period `TestContiguousIntervalCapacityInArrear#testMultipleItemsAndTiersWithExistingItems`:

Total usage amount is `90.00` -- taking into account `10.00` that was already billed. Now new detail json looks like:

```
{
  "tierDetails": [
    {
      "tier": 2,
      "tierUnit": "BAR",
      "tierPrice": 10,
      "quantity": 200
    },
    {
      "tier": 3,
      "tierUnit": "FOO",
      "tierPrice": 100,
      "quantity": 60
    }
  ],
  "amount": 100
}
```

It currently reflects the total number of units in this period, but i think we should add another field -- this is mostly why i broke the json to be different between capacity and consumable -- to include the part that was already paid. So maybe an extra field -- at the top level -- like `previouslyBilledAmount=10.00`

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
sbrossie: The trick is the following: Because we don't have previous details, we can't compute exact new details, we only know the total amount that was already invoiced but not the per-unit amount, so we end up with something like:

Billed amount $4
```
{
  "tierDetails": [
    {
      "tier": 1,
      "tierUnit": "unit",
      "tierPrice": 1,
      "tierBlockSize": 100,
      "quantity": 5,
      "amount": 5
    }
  ],
  "amount": 5
}
```

sbrossie: Pushed the change in 669bd343f132e40cf478347507a98da7736d54b6 , we can revisit if necessary

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalCapacityUsageInArrear.java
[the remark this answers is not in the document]
pierre: We could just check `billedUsage.compareTo(toBeBilledUsage) < 0`. If nothing was previously billed, `billedUsage` is `0`. And if `toBeBilledUsage > 0`, there's nothing to do.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalConsumableUsageInArrear.java
pierre: I don't think so, since it's the same logic for both modes.

Issue #833, opened 2018-01-05
Ability to void invoices


sbrossie:
There are situation where invoices have been generated, and where we would like to void such invoices. See [mailing list thread for instance](https://groups.google.com/forum/#!topic/killbilling-users/LFJOEPqrwvA).

In such situations, adjusting the invoice would not work, because we want this invoice to exist at some point, but it might have been generated too early.

sbrossie:
The work consists in the following:

**API changes:**

1.  Adding a new enum `VOID` field to our [InvoiceStatus enum](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/invoice/api/InvoiceStatus.java#L21)
2. Add a new `voidInvoice` api similar to what we have for [commitInvoice](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/invoice/api/InvoiceUserApi.java#L309)

**Implementation:**

1. Modify the invoice logic to exclude such invoices in the following use cases:
1.1 Invoice Balance [computation](https://github.com/killbill/killbill/blob/work-for-release-0.19.x/invoice/src/main/java/org/killbill/billing/invoice/model/DefaultInvoice.java#L260)
1.2 Account Balance [computation](https://github.com/killbill/killbill/blob/work-for-release-0.19.x/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L462)
1.3 When generating new invoices, we currently exclude `DRAFT` [here](https://github.com/killbill/killbill/blob/work-for-release-0.19.x/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L571), we should also exclude `VOID`. Maybe changing the `InvoiceDao#getAllInvoicesByAccount` implementation [here](https://github.com/killbill/killbill/blob/work-for-release-0.19.x/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L159) would make sense to filter at a lower level, or optionally adding a parameter to decide whether such invoices should be returned or not.

Probably a good thing to do -- which i did quickly-- is to verify all over the code how we use `InvoiceStatus`, knowing that now there is a new field so some equality maybe have to be rethought? Also ensure logic to make sure once invoice has been 'VOIDed' it stays in this state. `DRAFT` or `COMMITTED` invoice should be able to be voided.


2. JAXRS
[...]

PR #871, opened 2018-02-16
Ability to void invoices #833

wwjbatista:


profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestInvoiceVoid.java
```diff
@@ -146,6 +146,88 @@ public void testInvoiceVoidWithPartialPay() throws Exception {
 
     }
 
+
+    @Test(groups = "slow", description = "Void a child invoice")
+    public void testChildVoidInvoice() throws Exception {
```
sbrossie: Might be better to move this test under `TestIntegrationParentInvoice` since this is an HA test.

sbrossie:
@wwjbatista Will merge PR, but possible we will have a follow up PR with the following:
* Add a test case to verify what happens when an invoice on which credit was applied is then voided -> would assume the credit is lost
* Revisit hierarchical account (HA) . Your tests show the current behavior, but we may have to revisit what this behavior is.

Issue #850, opened 2018-01-26
DefaultInternalBillingApi optimizations


pierre:
The goal of this task is to optimize `DefaultInternalBillingApi` and limit the number of database queries made.

Work to be done:
* [ ] One query to bulk fetch all tags instead of doing one query per account (see `getBillingEventsForAccountAndUpdateAccountBCD`) and per bundle (see `addBillingEventsForBundles`) (`TagInternalApi` might need to be enhanced to expose `getTagsForAccount`)
* [ ] One query to bulk fetch all subscriptions at once instead of one query per bundle (see `addBillingEventsForBundles`)

You can validate that no database call (DAO call) is issued twice by triggering an invoice generation (see https://github.com/killbill/killbill-docs/blob/HEAD/userguide/subscription/includes/invoicing-dry-run.adoc) and enabling profiling (see https://github.com/killbill/killbill/wiki/Kill-Bill-Profiling):

```
curl  -v \
    -X POST \
    -u admin:password \
    -H "X-Killbill-ApiKey: bob" \
    -H "X-Killbill-ApiSecret: lazar" \
    -H "X-Killbill-CreatedBy: demo" \
    -H "Content-Type: application/json" \
    -H 'X-Killbill-Profiling-Req: DAO_DETAILS'  \
    --data-binary '{"dryRunType":"UPCOMING_INVOICE"}' \
    'http://127.0.0.1:8080/1.0/kb/invoices/dryRun?accountId=<ACCOUNT_ID>'
       | grep 'X-Killbill-Profiling-Resp' \
       | awk '{$1=$2=""; print $0}' \
       | python -mjson.tool
```

PR #873, opened 2018-02-20
 DefaultInternalBillingApi optimizations #850

wwjbatista:


api/src/main/java/org/killbill/billing/tag/TagInternalApi.java
```diff
@@ -43,6 +43,8 @@
 
     public List<Tag> getTagsForAccountType(ObjectType objectType, boolean includedDeleted, InternalTenantContext internalTenantContext);
 
+    public List<Tag> getTagsForAccountId(boolean includedDeleted, InternalTenantContext context);
```
pierre: Rename to `getTagsForAccount` since the `accountId` isn't passed explicitly.

junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java
```diff
@@ -93,7 +95,8 @@ public BillingEventSet getBillingEventsForAccountAndUpdateAccountBCD(final UUID
         final Catalog currentCatalog =  catalogInternalApi.getFullCatalog(true, true, context);
 
         // Check to see if billing is off for the account
-        final List<Tag> accountTags = tagApi.getTags(accountId, ObjectType.ACCOUNT, context);
+        final List<Tag> tags = tagApi.getTagsForAccountId(false, context);
```
pierre: We should rename the variable `tags` here and below (in `addBillingEventsForBundles`) to avoid any confusion. Maybe `tagsForAccount`?

junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java
```diff
@@ -151,15 +154,22 @@ private void addBillingEventsForBundles(final List<SubscriptionBaseBundle> bundl
 
         }
 
+        final List<Tag> bundleTags = getTagsForObjectType(ObjectType.BUNDLE, tags);
```
pierre: ⚠️ You need to filter by `bundleId` too (see original call below), because each bundle could be independently tagged with `AUTO_INVOICING_OFF`.

junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java
```diff
@@ -151,15 +154,22 @@ private void addBillingEventsForBundles(final List<SubscriptionBaseBundle> bundl
...
             final DryRunArguments dryRunArgumentsForBundle = (dryRunArguments != null &&
                                                               dryRunArguments.getBundleId() != null &&
                                                               dryRunArguments.getBundleId().equals(bundle.getId())) ?
                                                              dryRunArguments : null;
-            final List<SubscriptionBase> subscriptions = subscriptionApi.getSubscriptionsForBundle(bundle.getId(), dryRunArgumentsForBundle, context);
+            final List<SubscriptionBase> subscriptions;
+            if (dryRunArgumentsForBundle == null || dryRunArgumentsForBundle.getAction() == null) {
+                subscriptions = subscriptionsForAccount.get(bundle.getId());
+            } else {
+                subscriptions = subscriptionApi.getSubscriptionsForBundle(bundle.getId(), dryRunArgumentsForBundle, context);
```
pierre: Yup, I'm happy not optimizing that path, since it's not the common one. But let's add a comment to explain that we deliberately ignored the problem here, so we remember.

pierre:
Will merge as-is, as I wasn't able to reproduce the CI failure:

```
Error: test_change_multiple_times_over_initial_trial_period_2(KillBillIntegrationTests::TestEntitlementChangeSkipPhase)
  KillBillClient::API::InternalServerError: {"className":"org.skife.jdbi.v2.exceptions.CallbackFailedException","code":null,"message":"java.lang.NullPointerException","causeClassName":"java.lang.NullPointerException","causeMessage":null,"stackTrace":[]}
```

Issue #807, opened 2017-10-11
Ability to update certain invoice items for a DRAFT invoice


sbrossie:
In a scenario where an a given `Invoice` is in `DRAFT` mode, it might make sense to allow invoice plugins to modify some previously returned items instead or adding new correction items. The list of items would be limited to the one [the plugins currently can return](https://github.com/killbill/killbill/blob/killbill-0.18.13/invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java#L46).

I suppose the main idea here is to make the invoice more readable, but should there be instead a presentation layer displaying invoices to customer -- for instance by compacting items.

sbrossie:
@beccagaspard @sruthipendyala I am trying to understand your use case. I suppose you are relying on #767 and #789 in such a way that invoicing code will invoke your custom invoice plugin with an **existing** invoice in **Draft** mode. In that case, it seems like what is needed is an extension of the `InvoicePluginApi` to allow modifying items in current invoice.

The main issue with this,  is that this becomes really custom to using the system with both #767 and #789.

PR #890, opened 2018-03-08
invoice: verify we can set linked_item_id on external charges

pierre:
This fixes https://github.com/killbill/killbill/issues/887.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
@@ -48,7 +48,6 @@ import org.killbill.billing.catalog.api.BillingActionPolicy;
 import org.killbill.billing.catalog.api.CatalogApiException;
 import org.killbill.billing.catalog.api.PlanPhasePriceOverride;
 import org.killbill.billing.catalog.api.PlanPhaseSpecifier;
-import org.killbill.billing.entitlement.api.Entitlement.EntitlementState;
 import org.killbill.billing.entitlement.api.SubscriptionEventType;
 import org.killbill.billing.events.BusInternalEvent;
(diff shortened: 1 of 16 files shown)


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -571,7 +571,30 @@ public boolean apply(final InvoiceItem input) {
                     if (exitingItem != null) {
                         invoice.removeInvoiceItem(exitingItem);
                     }
-                    invoice.addInvoiceItem(cur);
+
+                    final InvoiceItem sanitizedInvoiceItemFromPlugin = new InvoiceItemCatalogBase(cur.getId(),
+                                                                                                  cur.getCreatedDate(),
+                                                                                                  MoreObjects.firstNonNull(cur.getInvoiceId(), invoice.getId()),
```
sbrossie: Interesting, so does this mean that we allow to add additional items outside or the current invoice being generated?

pierre: Yes, but this is not new (I believe we allowed this when we implemented the `UPDATE` feature, see https://github.com/killbill/killbill/issues/807).

pierre:
> Looks good: The main concern is that we allow more and more (sensitive) things to be set be set by the plugin, without running much validation on our end

I had the same thought while writing the patch: it's easier now for newcomers to write dangerous invoice plugins.

However, I then thought about the other end of the spectrum, i.e. power users who know exactly what they are doing. A lot of the changes we've made recently was specifically to remove extra sanity checks and constraints we put in place to unblock these power users.

It's a bit similar to third party libraries developers who insist on making all internal classes `private final` - while I understand why they would want to do this, how many times did I complain to you about it?...

PR #892, opened 2018-03-08
expose log management to JMX

laurentbh:


profiles/killbill/src/main/java/org/killbill/billing/server/listeners/CleanupListener.java
```diff
@@ -76,5 +77,9 @@ public void contextDestroyed(final ServletContextEvent servletContextEvent) {
                 logger.warn("Unable to de-register driver", e);
             }
         }
+
+//         avoid memory leaks: https://logback.qos.ch/manual/jmxConfig.html
+        LoggerContext lc = (LoggerContext) LoggerFactory.getILoggerFactory();
+        lc.stop();
```
pierre: @laurentbh Did you verify that, if `jmxConfigurator` isn't set in the `logback.xml`, the `stop` here is a no-op?

laurentbh: yes, no exception raised when jmx is not enable in`logback.xml`

pierre:
> Probably you will need to also update the [template in ansible](https://github.com/killbill/killbill-cloud/blob/work-for-release-0.19.x/ansible/templates/killbill/logback.xml.j2)

Done in https://github.com/killbill/killbill-cloud/commit/088777ec734821843e73a87631ed2351798af6f8.

Issue #897, opened 2018-03-13
NPE in DefaultInternalBillingApi (future cancellation with pending PHASE event)


pierre:
Scenario:

* Base plan + add-on
* Future cancel the base plan when the add-on has a `PENDING` phase event
* Trigger a dry-run invoice and notice the following stacktrace:

```
java.lang.NullPointerException
at org.killbill.billing.junction.plumbing.billing.DefaultInternalBillingApi.getPlanPhaseSpecifierFromTransition(DefaultInternalBillingApi.java:243)
at org.killbill.billing.junction.plumbing.billing.DefaultInternalBillingApi.calculateBcdForTransition(DefaultInternalBillingApi.java:229)
at org.killbill.billing.junction.plumbing.billing.DefaultInternalBillingApi.addBillingEventsForSubscription(DefaultInternalBillingApi.java:213)
at org.killbill.billing.junction.plumbing.billing.DefaultInternalBillingApi.addBillingEventsForBundles(DefaultInternalBillingApi.java:168)
at org.killbill.billing.junction.plumbing.billing.DefaultInternalBillingApi.getBillingEventsForAccountAndUpdateAccountBCD(DefaultInternalBillingApi.java:109)
at org.killbill.billing.invoice.InvoiceDispatcher.processAccountWithLock(InvoiceDispatcher.java:279)
at org.killbill.billing.invoice.InvoiceDispatcher.processAccount(InvoiceDispatcher.java:250)
at org.killbill.billing.invoice.api.user.DefaultInvoiceUserApi.triggerInvoiceGeneration(DefaultInvoiceUserApi.java:215)
```

PR #900, opened 2018-03-14
subscription: code review for #897

pierre:
Found a bug in the process in `DefaultSubscriptionDao.java`.

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
@@ -671,7 +671,28 @@ public class DefaultSubscriptionBase extends EntityBase implements SubscriptionB
 
         this.events = inputEvents;
 
-        filterOutDuplicateCancelEvents(events);
+        Collections.sort(inputEvents, new Comparator<SubscriptionBaseEvent>() {
+            @Override
+            public int compare(final SubscriptionBaseEvent o1, final SubscriptionBaseEvent o2) {
+                final int res = o1.getEffectiveDate().compareTo(o2.getEffectiveDate());
+                if (res != 0) {
+                    return res;
+                }
+
+                // In-memory events have a total order of 0, make sure they are after on disk event
+                if (o1.getTotalOrdering() == 0 && o2.getTotalOrdering() > 0) {
+                    return 1;
+                } else if (o1.getTotalOrdering() > 0 && o2.getTotalOrdering() == 0) {
+                    return -1;
+                } else if (o1.getTotalOrdering() == o2.getTotalOrdering()) {
+                    return 0;
+                } else {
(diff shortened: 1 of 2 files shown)


subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -811,73 +813,41 @@ public boolean apply(final SubscriptionBaseEvent input) {
...
+                // Keep the cancellation event
+            } else if (input.getType() == EventType.API_USER && (((ApiEvent) input).getApiEventType() == ApiEventType.TRANSFER || ((ApiEvent) input).getApiEventType() == ApiEventType.CREATE)) {
+                // Keep the initial event (SOT use-case)
+            } else if (input.getType() == EventType.API_USER && (((ApiEvent) input).getApiEventType() == ApiEventType.CANCEL) && !((ApiEvent) input).isFromDisk()) {
+                // Also we remove !onDisk cancel events when there is an onDisk cancel event (can happen during the path where we process the base plan cancel notification, and are
+                // in the process of adding the new cancel events for the AO)
+                it.remove();
+            } else if (input.getEffectiveDate().compareTo(cancellationEvent.getEffectiveDate()) >= 0) {
+                // Event to ignore past cancellation date
+                it.remove();
```
sbrossie: Trying to step back from what we had and thinking about what i think we should have:

Assumptions: events have been ordered using the `Comparator` above -- you fixed.

1. First  event should be `CREATE ` or `TRANSFER` => Delete every strictly before  this.
2. One we hit the first `CANCEL`, delete everything after or equals to this date. Note that because of the ordering, this should take into account use case where we have a `!isFromDisk` `CANCEL` event because it would come after in the stream and therefore would be deleted with this logic.

Also as we remove events from the stream --maybe it would be worth logging a `WARN` as this represent data issues i would think.

pierre: > One we hit the first `CANCEL`, delete everything after or equals to this date. Note that because of the ordering, this should take into account use case where we have a `!isFromDisk` `CANCEL` event because it would come after in the stream and therefore would be deleted with this logic.

Right, I think this branch could be deleted now that the Dao bug has been fixed. I'll check.

> Also as we remove events from the stream --maybe it would be worth logging a `WARN` as this represent data issues i would think.

I don't think so, unless we change other parts of the code. For add-ons, for instance, we almost always have two `CANCEL` events, one from the disk and one added in memory.

pierre: > Right, I think this branch could be deleted now that the Dao bug has been fixed. I'll check.

Confirmed in 221405dfcd02dbfc1e3858bc50fbcfe6f48d0a03.

PR #901, opened 2018-03-14
Extend ObfuscatorConverter to accept additional obfusacator

daliwei:
The `Obfuscator.java` is just renamed to `BaseObfuscator.java` and added an interface in `Obfuscator.java`. 

The way to add the additional Obfusactor into logback configuration is like:

`%maskedMsg{"<full class name of the new obfuscator 1>", "<full class name of the new obfuscator 2>", ...}`

cc @pierre @sbrossie for review.

### DIFF
--- a/profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/Obfuscator.java
+++ b/profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/Obfuscator.java
@@ -1,6 +1,6 @@
 /*
- * Copyright 2014-2015 Groupon, Inc
- * Copyright 2014-2015 The Billing Project, LLC
+ * Copyright 2014-2018 Groupon, Inc
+ * Copyright 2014-2018 The Billing Project, LLC
  *
  * The Billing Project licenses this file to you under the Apache License, version 2.0
  * (the "License"); you may not use this file except in compliance with the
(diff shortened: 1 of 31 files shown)


profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/Obfuscator.java
```diff
@@ -17,90 +17,10 @@
...
-            }
-        }
-        return false;
-    }
+public interface Obfuscator {
 
-    private String obfuscateConfidentialData(final CharSequence confidentialSequence) {
-        return obfuscateConfidentialData(confidentialSequence, null);
-    }
+    String obfuscate(final String originalString, final ILoggingEvent event);
```
pierre: If the goal is to share this interface (e.g. with plugins), I think it would make sense to remove the dependency on `Logback`. We could have our own, lightweight, `LoggingEvent` class?

profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/ObfuscatorConverter.java
```diff
@@ -61,4 +70,25 @@ public String convert(final ILoggingEvent event) {
...
+        try {
+            Obfuscator newObfuscator = createDynamicObfuscator(obfuscatorClassName);
+            obfuscators.add(newObfuscator);
+        } catch (Exception e) {
+            // Ignore? Not sure the impact of importing a logger here
+        }
+    }
+
+    private Obfuscator createDynamicObfuscator(final String className) throws Exception {
+        return (Obfuscator) Class.forName(className).getConstructor().newInstance();
```
pierre: Is it actually working? I wouldn't be surprised if you run into OSGI import/export problems (`ClassCastException`).

One work-around would be to share the `Obfuscator` interface (maybe in https://github.com/killbill/killbill-platform/tree/master/osgi-bundles/libs/killbill/src/main/java/org/killbill/billing/osgi/libs/killbill, package that is already imported both by plugins and by killbill server).

daliwei:
This solution seems to be quite complicated. I am closing this one since #905 can also achieve what I wanted. Thanks @pierre and @sbrossie for the review.

PR #904, opened 2018-03-15
Make pattern obsfucator accept parameters from logback config

daliwei:


profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/ObfuscatorConverter.java
```diff
@@ -44,10 +45,16 @@
...
-                                                                                    new LoggingFilterObfuscator(),
-                                                                                    new PatternObfuscator(),
-                                                                                    new LuhnMaskingObfuscator());
+    private final Collection<Obfuscator> obfuscators = new ArrayList<Obfuscator>();
+
+    @Override
+    public void start() {
+        obfuscators.addAll(Arrays.asList(new ConfigMagicObfuscator(),
+                                         new LoggingFilterObfuscator(),
+                                         new PatternObfuscator(getOptionList()),
```
pierre: Since we're at it, should we pass these options to all obfuscators? They all have some form of hardcoded keys that we might have to dynamically configure in the future (e.g. custom internal header in `LoggingFilterObfuscator`)?

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestObfuscatorConverter.java
```diff
@@ -60,11 +63,52 @@ public void testLogSensitiveData() throws Exception {
...
+
+    class ExtendedObfuscatorConverter extends ObfuscatorConverter {
+        @Override
+        public void start() {
+            super.start();
+        }
+
+        @Override
+        public List<String> getOptionList() {
+            return ImmutableList.of("address1");
```
pierre: Could you document in the Javadocs of `ObfuscatorConverter` how you would configure this in the `logback.xml`?

sbrossie:
👍

profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/ObfuscatorConverter.java
daliwei: I guess the exception is the LuhnMaskingObfuscator. Since they are very similar to the PatternObfuscator, we can come back to add it easily when we need that extra flexibility.

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestObfuscatorConverter.java
daliwei: Sure.

daliwei:
Close this one in favor of #905

PR #905, opened 2018-03-15
Extend pattern obfuscator

daliwei:
Same PR as #904 but against master. cc @sbrossie @pierre


daliwei:
@pierre I adjusted the pattern for xml in 13fcdaa as I mentioned offline

pierre:
Merged in 9532b5bc2f99fb5bfb98f6cf6c3238a65e78b17e.

PR #910, opened 2018-03-19
Add new endpoint to retrieve a list of catalog versions; Add query parameter `requestDate` to retrieve only one catalog xml

wwjbatista:
Ref killbill/killbill-admin-ui#141

Add new endpoint to retrieve a list of catalog versions
Add query parameter `requestDate` to retrieve only one catalog xml

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
@@ -126,15 +126,16 @@ public class CatalogResource extends JaxRsResourceBase {
                                             null;
 
         final VersionedCatalog catalog = (VersionedCatalog) catalogUserApi.getCatalog(catalogName, tenantContext);
-
-        String result = null;
+        final String result;
         if (catalogDateVersion != null) {
+            final VersionedCatalog oneVersionCatalog = new VersionedCatalog();
             for (final StandaloneCatalog v : catalog.getVersions()) {
-                if (v.getEffectiveDate().compareTo(catalogDateVersion.toDate()) == 0) {
-                    result = XMLWriter.writeXML(v, StandaloneCatalog.class);
+                if (v.getEffectiveDate().compareTo(catalogDateVersion.toDate()) >= 0) {
+                    oneVersionCatalog.add(v);
                     break;
                 }
             }
+            result = XMLWriter.writeXML(oneVersionCatalog, VersionedCatalog.class);
         } else {
             result = XMLWriter.writeXML(catalog, VersionedCatalog.class);
         }
@@ -189,7 +190,7 @@ public class CatalogResource extends JaxRsResourceBase {
     @GET
     @Path("/versions")
     @Produces(APPLICATION_JSON)
-    @ApiOperation(value = "Retrieve a list of catalog versions", response = DateTime.class)
+    @ApiOperation(value = "Retrieve a list of catalog versions", response = DateTime.class, responseContainer = "List")
     @ApiResponses(value = {})
     public Response getCatalogVersionJson(@javax.ws.rs.core.Context final HttpServletRequest request) throws Exception {
 
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
@@ -166,6 +166,25 @@ public class CatalogResource extends JaxRsResourceBase {
         return Response.status(Status.OK).entity(result).build();
     }
 
+    @TimedResource
+    @GET
+    @Path("/versions")
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Retrieve a list of catalog versions", response = DateTime.class)
+    @ApiResponses(value = {})
+    public Response getCatalogVersionJson(@javax.ws.rs.core.Context final HttpServletRequest request) throws Exception {
+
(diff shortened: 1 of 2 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
```diff
@@ -166,6 +185,25 @@ public Response getCatalogJson(@QueryParam(QUERY_REQUESTED_DT) final String requ
         return Response.status(Status.OK).entity(result).build();
     }
 
+    @TimedResource
+    @GET
+    @Path("/versions")
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Retrieve a list of catalog versions", response = DateTime.class)
```
pierre: You also need `, responseContainer = "List"` because we are returning a list.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
```diff
@@ -118,9 +118,28 @@ public CatalogResource(final JaxrsUriBuilder uriBuilder,
...
+        final DateTime catalogDateVersion = requestedDate != null ?
+                                            DATE_TIME_FORMATTER.parseDateTime(requestedDate).toDateTime(DateTimeZone.UTC) :
+                                            null;
+
+        final VersionedCatalog catalog = (VersionedCatalog) catalogUserApi.getCatalog(catalogName, tenantContext);
+
+        String result = null;
+        if (catalogDateVersion != null) {
+            for (final StandaloneCatalog v : catalog.getVersions()) {
+                if (v.getEffectiveDate().compareTo(catalogDateVersion.toDate()) == 0) {
```
sbrossie: I am wondering if we should not use `v.getEffectiveDate().compareTo(catalogDateVersion.toDate()) >= 0`, otherwise this forces the user to know exactly the timestamp of each catalog. Let 's assume two versions: `2018-01-01T00:00:00+00:00` and `2018-03-01T00:00:00+00:00`:

* Providing any date >=  `2018-01-01T00:00:00+00:00` and < `2018-03-01T00:00:00+00:00` would return first version
* Providing any date >= `2018-03-01T00:00:00+00:00` would return second version.

That way the parameter request date represents the date within the range of date for which this catalog version is effective.

@pierre Any thoughts?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
```diff
@@ -118,9 +118,28 @@ public CatalogResource(final JaxrsUriBuilder uriBuilder,
...
+                    result = XMLWriter.writeXML(v, StandaloneCatalog.class);
+                    break;
```
sbrossie: The other issue is that this end point now returns different type of response based on the query param. I am wondering if it would not be cleaner to either have 2 separate endpoints or at least return the same type of response. A `VersionedCatalog` is pretty much a list of `StandaloneCatalog` so we could return one one element when query param is provided. 

@pierre Any thoughts?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
```diff
@@ -166,6 +185,25 @@ public Response getCatalogJson(@QueryParam(QUERY_REQUESTED_DT) final String requ
...
+        final TenantContext tenantContext = context.createTenantContextNoAccountId(request);
+        final VersionedCatalog catalog = (VersionedCatalog) catalogUserApi.getCatalog(catalogName, tenantContext);
+
+        final List<DateTime> result = new ArrayList<DateTime>();
+        for (final StandaloneCatalog v : catalog.getVersions()) {
+            result.add(new DateTime(v.getEffectiveDate()));
+        }
+
+        return Response.status(Status.OK).entity(result).build();
+    }
```
sbrossie: Does not this endpoint [already exist](https://github.com/killbill/killbill/blob/work-for-release-0.19.x/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java#L147) ?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
pierre: @sbrossie I had indeed thought about it. However, in this case, versions are used as a number or `id`: we are looking for a specific catalog version here (it's different than retrieving the catalog at a given point in time).

That being said, I'm happy either way - your call.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
pierre: 👍 for returning a `VersionedCatalog` with a single catalog in that case.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
pierre: This endpoint only returns the versions, not the full json/xml.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
sbrossie: Agreed that for the KAUI use cases, where exact versions -- as specified with their `effectiveDate` -- this works 'as is', but thinking of api, might be more powerful to all for any date within the range ?

Issue #893, opened 2018-03-12
Invoice Plugin Api Enhancements


sbrossie:
The proposal is to enhance our current [invoice plugin api](https://github.com/killbill/killbill-plugin-api/blob/killbill-plugin-api-0.24.4/invoice/src/main/java/org/killbill/billing/invoice/plugin/api/InvoicePluginApi.java) in a similar fashion with what we have with other plugin apis -- e.g [PaymentControlPluginApi](https://github.com/killbill/killbill-plugin-api/blob/killbill-plugin-api-0.24.4/control/src/main/java/org/killbill/billing/control/plugin/api/PaymentControlPluginApi.java):

* Ability to provide a callback,`priorCall`, before invoice generation: This would allow plugin implementors to abort invoice generation, delay invoice generation by returning a new `effectiveDate`, ...
* Ability to provide a callback, `onSuccessCall` -- or `onFailureCall ` -- after invoice generation. We would return the full invoice generated when successful to avoid having to fetch the invoice again.

PR #923, opened 2018-03-22
invoice: Invoice Plugin Api Enhancements

pierre:
Initial implementation for https://github.com/killbill/killbill/issues/893.

Once we agree on the semantics, I'll add tests and update dependencies (e.g. killbill-plugin-framework-java library).

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
@@ -390,6 +390,10 @@ public class InvoiceDispatcher {
             log.warn("Failed to retrieve BillingEvents for accountId='{}', dryRunArguments='{}'", accountId, dryRunArguments, e);
             return null;
         } catch (final InvoiceApiException e) {
+            if (e.getCode() == ErrorCode.INVOICE_PLUGIN_API_ABORTED.getCode()) {
+                return null;
+            }
+
             if (e.getCode() == ErrorCode.UNEXPECTED_ERROR.getCode() && !isDryRun) {
                 log.warn("Illegal invoicing state detected for accountId='{}', dryRunArguments='{}', parking account", accountId, dryRunArguments, e);
                 parkAccount(accountId, context);
--- a/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
@@ -509,11 +509,17 @@ public class InvoiceDispatcher {
                                                              final List<Invoice> existingInvoices,
                                                              final boolean isDryRun,
                                                              final InternalCallContext internalCallContext) throws InvoiceApiException {
+        final boolean isRescheduled = false; // TODO
+
+        final CallContext callContext = buildCallContext(internalCallContext);
+        invoicePluginDispatcher.priorCall(targetDate, existingInvoices, isDryRun, isRescheduled, callContext, internalCallContext);
+
         final ImmutableAccountData account;
         try {
             account = accountApi.getImmutableAccountDataById(accountId, internalCallContext);
         } catch (final AccountApiException e) {
             log.error("Unable to generate invoice for accountId='{}', a future notification has NOT been recorded", accountId, e);
+            invoicePluginDispatcher.onFailureCall(targetDate, null, existingInvoices, isDryRun, isRescheduled, callContext, internalCallContext);
             return null;
         }
 
@@ -525,6 +531,8 @@ public class InvoiceDispatcher {
 
         // If invoice comes back null, there is nothing new to generate, we can bail early
         if (invoice == null) {
+            invoicePluginDispatcher.onSuccessCall(targetDate, null, existingInvoices, isDryRun, isRescheduled, callContext, internalCallContext);
+
             if (isDryRun) {
                 log.info("Generated null dryRun invoice for accountId='{}', targetDate='{}'", accountId, targetDate);
             } else {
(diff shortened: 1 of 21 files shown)


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -508,12 +510,27 @@ private Invoice processAccountWithLockAndInputTargetDate(final UUID accountId,
...
         } catch (final AccountApiException e) {
             log.error("Unable to generate invoice for accountId='{}', a future notification has NOT been recorded", accountId, e);
+            invoicePluginDispatcher.onFailureCall(targetDate, null, existingInvoices, isDryRun, isRescheduled, callContext, internalCallContext);
+            return null;
+        }
+
+        final DateTime rescheduleDate = invoicePluginDispatcher.priorCall(targetDate, existingInvoices, isDryRun, isRescheduled, callContext, internalCallContext);
+        if (rescheduleDate != null) {
+            if (isDryRun) {
+                log.warn("Ignoring rescheduleDate='{}', delayed scheduling is unsupported in dry-run", rescheduleDate);
```
sbrossie: We could even potentially throw -- plugin implementors should fix their code to not return `rescheduleDate` on dryRun.

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
```diff
@@ -48,30 +55,116 @@
...
+            if (priorInvoiceResult == null) {
+                // Naughty plugin...
+                continue;
+            }
+
+            if (priorInvoiceResult.getRescheduleDate() != null &&
+                (earliestRescheduleDate == null || earliestRescheduleDate.compareTo(priorInvoiceResult.getRescheduleDate()) > 0)) {
+                earliestRescheduleDate = priorInvoiceResult.getRescheduleDate();
+                log.info("Invoice plugin {} rescheduled invoice generation to {} for targetDate {}", invoicePluginName, earliestRescheduleDate, targetDate);
+            }
```
sbrossie: Yeah, technically, there is some semantics discussions to have. In practice, if different plugins return different dates, finding the right semantics may be tricky...

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
```diff
@@ -48,30 +55,116 @@
...
+
+        // We clone the original invoice so plugins don't remove/add items
+        final Invoice clonedInvoice = (Invoice) originalInvoice.clone();
+        final InvoiceContext invoiceContext = new DefaultInvoiceContext(targetDate, clonedInvoice, existingInvoices, isDryRun, isRescheduled, callContext);
+
+        for (final InvoicePluginApi invoicePlugin : invoicePlugins) {
+            if (isSuccess) {
+                invoicePlugin.onSuccessCall(invoiceContext, ImmutableList.<PluginProperty>of());
+            } else {
+                invoicePlugin.onFailureCall(invoiceContext, ImmutableList.<PluginProperty>of());
```
sbrossie: If i understand correctly the `onFailureCall` scenario maps to drastic error on our end -- cannot retrieve `Account`from id, fail to write to DB, ... I suppose this will inform plugin that shit happened but we don't allow plugin to take any action. At this point, `Account` should move into `__PARKED__` mode ?

invoice/src/main/java/org/killbill/billing/invoice/notification/DefaultNextBillingDatePoster.java
```diff
@@ -111,7 +112,7 @@ private void insertNextBillingFromTransactionInternal(final EntitySqlDaoWrapperF
             if (existingNotificationForEffectiveDate == null) {
                 log.info("Queuing next billing date notification at {} for subscriptionId {}", futureNotificationTime.toString(), JOINER.join(subscriptionIds));
 
-                final NextBillingDateNotificationKey newNotificationEvent = new NextBillingDateNotificationKey(null, subscriptionIds, targetDate, isDryRunForInvoiceNotification);
+                final NextBillingDateNotificationKey newNotificationEvent = new NextBillingDateNotificationKey(null, subscriptionIds, targetDate, isDryRunForInvoiceNotification, isRescheduled);
```
sbrossie: ⚠️  The issue here is that if we end up inserting a new notification, all is good, but if not, how do we handle the `isRescheduled` flag?

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
pierre: Yup, I went back and forth on this but I was worried it would break things in a subtle way (e.g. AJAX call in Kaui to display the next invoice day).

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
pierre: > At this point, `Account` should move into `__PARKED__` mode ? 

Actually, back then, we had decided to park accounts only for billing errors (when `ErrorCode.UNEXPECTED_ERROR` is thrown) or when the billing system is disabled. Any other exception (`CatalogApiException`, `AccountApiException`, etc.) is ignored.

See logic in `processAccountWithLock`.

invoice/src/main/java/org/killbill/billing/invoice/notification/DefaultNextBillingDatePoster.java
pierre: See https://github.com/killbill/killbill/pull/923/files#diff-9f199d591e4f5b2d3ba23e4b87e150d4R532, where I add a notification at `rescheduleDate` for each subscription id.

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
sbrossie: Ah, interesting, so maybe a useful thing would be to allow plugin to decide whether account should be parked or not?

Issue #825, opened 2017-12-05
Duplicate key error when creating a tenant with wrong credentials


pierre:
When providing wrong credentials (e.g. `bob`/`foo`) and trying to create a new tenant already existing, we get a duplicate key error from the database.

It looks like `DefaultTenantUserApi#createTenant` is missing something like:

```
// Not transactional, but there is a db constraint on that column
if (data.getExternalKey() != null && getTenantByApiKey(data.getExternalKey(), context) != null) {
  throw new TenantApiException(ErrorCode.TENANT_ALREADY_EXISTS, data.getExternalKey());
}
```

PR #925, opened 2018-03-23
 Duplicate key error when creating a tenant with wrong credentials #825

wwjbatista:


tenant/src/main/java/org/killbill/billing/tenant/api/user/DefaultTenantUserApi.java
```diff
@@ -91,6 +91,11 @@ public Tenant createTenant(final TenantData data, final CallContext context) thr
             throw new TenantApiException(ErrorCode.EXTERNAL_KEY_LIMIT_EXCEEDED);
         }
 
+        // Not transactional, but there is a db constraint on that column
+        if (data.getApiKey() != null && getTenantByApiKey(data.getApiKey()) != null) {
+            throw new TenantApiException(ErrorCode.TENANT_ALREADY_EXISTS, data.getExternalKey());
```
sbrossie: It seems like we have [such constraint](https://github.com/killbill/killbill/blob/killbill-0.19.6/tenant/src/main/resources/org/killbill/billing/tenant/ddl.sql#L18) though.

pierre:
> I think this should be fixed at the `DefaultTenantDao` level

I had actually suggested to fix it here in the ticket, as it's how we do it for `AccountUserApi` for instance.

> code [seems in place](https://github.com/killbill/killbill/blob/killbill-0.19.6/tenant/src/main/java/org/killbill/billing/tenant/dao/DefaultTenantDao.java#L78), but need to investigate why this does not work.

`generateAlreadyExistsException` is thrown only when `checkEntityAlreadyExists` is true, which is when the `record_ids` match (except for tag and custom fields).

@wwjbatista Let me sync up with @sbrossie and I'll update the PR with changes required (if any).

tenant/src/main/java/org/killbill/billing/tenant/api/user/DefaultTenantUserApi.java
sbrossie: I just sync'ed up with @pierre and this is is fine -- seems like this is already the model we have in place for other resources.

pierre:
@wwjbatista 👍 for the implementation but could you take a look at the test failure?

```
testTenant - org.killbill.billing.tenant.api.user.TestDefaultTenantUserApi
```

PR #928, opened 2018-03-27
Test coverage for 893

pierre:


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoicePlugin.java
```diff
@@ -246,6 +266,209 @@ public boolean apply(final InvoiceItem input) {
...
+                                    new ExpectedInvoiceItemCheck(new LocalDate(2012, 4, 1), null, InvoiceItemType.FIXED, new BigDecimal("0")));
+        subscriptionChecker.checkSubscriptionCreated(bpSubscription.getId(), internalCallContext);
+
+        Assert.assertEquals(testInvoicePluginApi.invocationCount, 1);
+
+        // Reschedule invoice generation at the time of the PHASE event
+        testInvoicePluginApi.rescheduleDate = new DateTime(clock.getUTCNow()).plusDays(30);
+
+        try {
+            invoiceUserApi.triggerInvoiceGeneration(account.getId(), clock.getUTCToday(), null, callContext);
```
sbrossie: Interesting... I am wondering if the rescheduling should not be only for recurring invoices; in this case user explicitly asks for an invoice. I also don't know if we care too much.

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -651,8 +651,8 @@ public boolean apply(final InvoiceItem input) {
                 commitInvoiceAndSetFutureNotifications(account, null, futureAccountNotifications, internalCallContext);
             }
 
-            if (success) {
-                final DefaultInvoice refreshedInvoice = new DefaultInvoice(invoiceDao.getById(invoice.getId(), internalCallContext));
+            if (isDryRun || success) {
```
sbrossie: Seems weird to call `onSuccessCall` if `isDryRun` its `true` but `success` is `false` ?

pierre: Indeed, but it's because how the code is written: if `isDryRun`, `success` is always false (`success` means that we could write to disk the invoice and notifications if not dry run).

Issue #637, opened 2016-10-31
Dry run invoice applies credit to balance without taking into account invoice plugin items


pierre:
We always compute CBA items first, then call the invoice plugins.

In the non-dry-run scenario, [committing](https://github.com/killbill/killbill/blob/6823aa9ecbbb42a153312c0845aee0f8f3837b54/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L378) the invoice and items will result in a re-computation of the CBA (taking into account the items added by the plugins).

In the dry-run scenario, because we don't commit, the CBA is stale. It looks like we should re-run the CBA computation [after calling the plugins](https://github.com/killbill/killbill/blob/6823aa9ecbbb42a153312c0845aee0f8f3837b54/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L358-L362).

sbrossie:
I would suggest something slightly different.

Let's review the steps:
1. Code [calls invoiceGenerator](https://github.com/killbill/killbill/blob/killbill-0.18.2/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L389) to compute `FIXED`, `RECURRING`, `USAGE` items -> At this point balance is **necessarily** >= 0.
2. Today, we then compute `computeCBAOnExistingInvoice` prior calling the plugin 
3. We call the plugin to potentially add items.

I am not convinced about step 2, as the only thing that could happen is that we consume existing credit (but should have no impact in things like `TAX` item generation and such.

So we could now have a step 4. that recomputes again after the plugin the CBA situation (but then we would have to potentially merge multiple `CBA` items if we added one in step 2.

Instead i would suggest to move step 2 into step 4 (one computation of `CBA` after calling plugin). In essence plugin is the continuation of invoiceGenerator and `CBA` items are only added at the very end.

pierre:
> 1. Code [calls invoiceGenerator](https://github.com/killbill/killbill/blob/killbill-0.18.2/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L389) to compute `FIXED`, `RECURRING`, `USAGE` items -> At this point balance is **necessarily** >= 0.

No necessarily in case of repairs? In case of a change plan with pro-ration for instance, the new invoice will have a `RECURRING` and a `REPAIR_ADJ` at this point. Without the CBA item, the plugin would receive an invoice with a negative balance.

[...]

PR #934, opened 2018-04-03
Fix for 911

pierre:


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -854,9 +829,23 @@ private CallContext buildCallContext(final InternalCallContext context) {
         return internalCallContextFactory.createCallContext(context);
     }
 
-    private void setChargedThroughDates(final Collection<InvoiceItem> fixedPriceItems,
-                                        final Collection<InvoiceItem> recurringItems,
-                                        final InternalCallContext context) throws SubscriptionBaseApiException {
+    private void setChargedThroughDates(final Invoice invoice, final InternalCallContext context) throws SubscriptionBaseApiException {
```
sbrossie: Independently of your change, i am wondering if we should not also handle the `USAGE` types : Don't we want to end up with a non null CTD if we only bill for usage ? Might be worth creating a bug to track this-- understand what we do now, and whether this needs to be fixed.

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
```diff
@@ -167,22 +172,77 @@ private void onCompletionCall(final boolean isSuccess,
...
-            final List<InvoiceItem> items = invoicePlugin.getAdditionalInvoiceItems(clonedInvoice, isDryRun, ImmutableList.<PluginProperty>of(), callContext);
-            if (items != null) {
-                for (final InvoiceItem item : items) {
-                    validateInvoiceItemFromPlugin(item, invoicePlugin);
-                    additionalInvoiceItems.add(item);
+            final List<InvoiceItem> additionalInvoiceItemsForPlugin = invoicePlugin.getAdditionalInvoiceItems(clonedInvoice, isDryRun, ImmutableList.<PluginProperty>of(), callContext);
+            if (additionalInvoiceItemsForPlugin != null) {
+                for (final InvoiceItem additionalInvoiceItem : additionalInvoiceItemsForPlugin) {
+                    final InvoiceItem sanitizedInvoiceItem = validateAndSanitizeInvoiceItemFromPlugin(originalInvoice, additionalInvoiceItem, invoicePlugin);
+                    additionalInvoiceItems.add(sanitizedInvoiceItem);
```
sbrossie: Since you are improving this method -- validation piece -- on suggestion would be to change the following as well:

Today we pass `originalInvoice` and return a list of `additionalInvoiceItems` -- those are not really additional items. they are a mixed bags of modified existing and additional, therefore requiring logic on the caller side to sort that out. Maybe we could encapsulate that logic inside that method instead?

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
```diff
@@ -167,22 +172,77 @@ private void onCompletionCall(final boolean isSuccess,
...
+                                          mutableField("prettyPlanName", existingItem != null ? existingItem.getPrettyPlanName() : null, additionalInvoiceItem.getPrettyPlanName(), invoicePlugin),
+                                          mutableField("prettyPhaseName", existingItem != null ? existingItem.getPrettyPhaseName() : null, additionalInvoiceItem.getPrettyPhaseName(), invoicePlugin),
+                                          mutableField("prettyUsageName", existingItem != null ? existingItem.getPrettyUsageName() : null, additionalInvoiceItem.getPrettyUsageName(), invoicePlugin),
+                                          immutableField("startDate", existingItem, existingItem != null ? existingItem.getStartDate() : null, additionalInvoiceItem.getStartDate(), invoicePlugin),
+                                          immutableField("endDate", existingItem, existingItem != null ? existingItem.getEndDate() : null, additionalInvoiceItem.getEndDate(), invoicePlugin),
+                                          mutableField("amount", existingItem != null ? existingItem.getAmount() : null, additionalInvoiceItem.getAmount(), invoicePlugin),
```
sbrossie: Do we have use case to change the `amount` today ? Could this not lead to some repair/credit scenario?

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
```diff
@@ -167,22 +172,77 @@ private void onCompletionCall(final boolean isSuccess,
...
+
+        if (!ALLOWED_INVOICE_ITEM_TYPES.contains(additionalInvoiceItem.getInvoiceItemType()) && existingItem == null) {
+            log.warn("Ignoring invoice item of type {} from InvoicePlugin {}: {}", additionalInvoiceItem.getInvoiceItemType(), invoicePlugin, additionalInvoiceItem);
+            throw new InvoiceApiException(ErrorCode.INVOICE_ITEM_TYPE_INVALID, additionalInvoiceItem.getInvoiceItemType());
+        }
+
+        final UUID invoiceId = MoreObjects.firstNonNull(mutableField("invoiceId", existingItem != null ? existingItem.getInvoiceId() : null, additionalInvoiceItem.getInvoiceId(), invoicePlugin),
+                                                        originalInvoice.getId());
+        return new InvoiceItemCatalogBase(additionalInvoiceItem.getId(),
+                                          mutableField("createdDate", existingItem != null ? existingItem.getCreatedDate() : null, additionalInvoiceItem.getCreatedDate(), invoicePlugin),
```
sbrossie: Do we really want to allow modifying system generated `createdDate` ?

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
```diff
@@ -167,22 +172,77 @@ private void onCompletionCall(final boolean isSuccess,
...
+                      invoicePlugin, fieldName, existingValue, updatedValue);
+            return updatedValue;
+        } else {
+            return existingValue;
+        }
+    }
+
+    private <T> T immutableField(final String fieldName, @Nullable final InvoiceItem existingItem, @Nullable final T existingValue, @Nullable final T updatedValue, final InvoicePluginApi invoicePlugin) {
+        if (existingItem == null) {
+            return updatedValue;
```
sbrossie: In that case, it means that this is a new item so we end up accepting everything from plugin, but still did not validate that any of the core fundamental fields like `accountId`, `bundleId`, `subscriptionId`, .. are correct. Maybe this is good enough.

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
pierre: Fixed in 6d5af5ec7c07aaf8bd6e01a3f548132afc396a66.

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
pierre: > Do we have use case to change the `amount` today ?

We actually do, to update `TAX` (`DRAFT` mode scenario where the invoice is being updated) and potentially `USAGE` items (old incorrect amounts).

See:

* https://github.com/killbill/killbill/issues/807
* https://github.com/killbill/killbill/blob/fa757385fced8155f9d96053c2f2a846f7334167/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithTaxItems.java#L269-L270

> Could this not lead to some repair/credit scenario?

Definitively an advanced use-case... I could add logic to limit updates of specific fields to specific items, but I'm thinking that if a plugin developer is updating such fields, she's an advanced user and knows what she is doing.

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
pierre: `originalInvoice` in the non-API use-case is actually a temporary cloned invoice, with a temporary CBA item computed pre- invoice plugins (https://github.com/killbill/killbill/issues/637). Unfortunately the shuffling of that logic doesn't _fully_ help simplify the `processAccountWithLockAndInputTargetDate` implementation. See  15964b0c682fdfdc0bd130e94ac6113a58b3d016.

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
pierre: No strong feeling but I figured it could be useful if a plugin would like to update the `createdDate` of a `DRAFT` invoice (updated over several days).

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
pierre: Actually, by that time, if `existingItem == null`, we have verified that this new item is one of the `ALLOWED_INVOICE_ITEM_TYPES`, so we don't really care about most of these fields.

The main one to check would be `accountId`, although maybe one day we should deprecate it (since we already have the `account_record_id`).

PR #939, opened 2018-04-06
Proposed implementation to address issue killbill/killbill-admin-ui#137

wwjbatista:
Ref. killbill/killbill-api#41
Ref. killbill/killbill-admin-ui#137

### DIFF
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
@@ -351,6 +351,62 @@ public class DefaultEntitlementApi extends DefaultEntitlementApiBase implements 
     }
 
     @Override
+    public Entitlement addEntitlement(final UUID bundleId, final UUID accountId, final PlanPhaseSpecifier planPhaseSpecifier, final List<PlanPhasePriceOverride> overrides, @Nullable final LocalDate entitlementEffectiveDate, @Nullable final LocalDate billingEffectiveDate,
+                                      final boolean isMigrated, final Iterable<PluginProperty> properties, final CallContext callContext) throws EntitlementApiException {
+        logCreateEntitlement(log, bundleId, planPhaseSpecifier, overrides, entitlementEffectiveDate, billingEffectiveDate);
+
+        final EntitlementSpecifier entitlementSpecifier = new DefaultEntitlementSpecifier(planPhaseSpecifier, overrides);
+        final List<EntitlementSpecifier> entitlementSpecifierList = new ArrayList<EntitlementSpecifier>();
+        entitlementSpecifierList.add(entitlementSpecifier);
+        final BaseEntitlementWithAddOnsSpecifier baseEntitlementWithAddOnsSpecifier = new DefaultBaseEntitlementWithAddOnsSpecifier(
+                bundleId,
+                null,
+                entitlementSpecifierList,
+                entitlementEffectiveDate,
+                billingEffectiveDate,
+                isMigrated);
+        final List<BaseEntitlementWithAddOnsSpecifier> baseEntitlementWithAddOnsSpecifierList = new ArrayList<BaseEntitlementWithAddOnsSpecifier>();
+        baseEntitlementWithAddOnsSpecifierList.add(baseEntitlementWithAddOnsSpecifier);
+
+        final EntitlementContext pluginContext = new DefaultEntitlementContext(OperationType.CREATE_SUBSCRIPTION,
+                                                                               null,
+                                                                               null,
+                                                                               baseEntitlementWithAddOnsSpecifierList,
+                                                                               null,
+                                                                               properties,
+                                                                               callContext);
+
+        final WithEntitlementPlugin<Entitlement> addEntitlementWithPlugin = new WithEntitlementPlugin<Entitlement>() {
+            @Override
+            public Entitlement doCall(final EntitlementApi entitlementApi, final EntitlementContext updatedPluginContext) throws EntitlementApiException {
+
+                final DateTime now = clock.getUTCNow();
+                final InternalCallContext context = internalCallContextFactory.createInternalCallContext(bundleId, ObjectType.BUNDLE, callContext);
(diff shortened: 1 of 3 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -216,9 +216,10 @@ public Entitlement doOperation(final CallContext ctx) throws InterruptedExceptio
                 final LocalDate resolvedEntitlementDate = requestedDate != null ? toLocalDate(requestedDate) : toLocalDate(entitlementDate);
                 final LocalDate resolvedBillingDate = requestedDate != null ? toLocalDate(requestedDate) : toLocalDate(billingDate);
                 final List<PlanPhasePriceOverride> overrides = PhasePriceOverrideJson.toPlanPhasePriceOverrides(entitlement.getPriceOverrides(), spec, account.getCurrency());
-                final Entitlement result = createAddOnEntitlement ?
-                                           entitlementApi.addEntitlement(getBundleIdForAddOnCreation(entitlement), spec, overrides, resolvedEntitlementDate, resolvedBillingDate, isMigrated, pluginProperties, callContext) :
-                                           entitlementApi.createBaseEntitlement(account.getId(), spec, entitlement.getExternalKey(), overrides, resolvedEntitlementDate, resolvedBillingDate, isMigrated, renameKeyIfExistsAndUnused, pluginProperties, callContext);
+                final Entitlement result = createAddOnEntitlement || (ProductCategory.STANDALONE.toString().equals(entitlement.getProductCategory()) && entitlement.getBundleId() != null)
```
pierre: Nit: add parens around `(ProductCategory.STANDALONE.toString().equals(entitlement.getProductCategory()) && entitlement.getBundleId() != null)` for clarity (or maybe even extract this as `final boolean createStandaloneEntitlement`).

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -350,6 +340,37 @@ public Entitlement doCall(final EntitlementApi entitlementApi, final Entitlement
...
+            subscriptionsByBundle = subscriptionBaseInternalApi.getSubscriptionsForBundle(bundleId, null, context);
+
+            if (subscriptionsByBundle == null || subscriptionsByBundle.size() == 0) {
+                throw new EntitlementApiException(ErrorCode.SUB_NO_ACTIVE_SUBSCRIPTIONS, bundleId);
+            }
+        } catch (SubscriptionBaseApiException e) {
+            throw new EntitlementApiException(e);
+        }
+
+        if (!subscriptionsByBundle.get(0).getCategory().equals(ProductCategory.STANDALONE)) {
```
pierre: Nit: flip `equals()`.

sbrossie: I would suggest to harden this check:
1. If the subscription we are trying to add is a `STANDALONE`, existing bundle should only contain `STANDALONE` subscriptions -- maybe this is enough to check the first one
2.  If the subscription we are trying to add is a `ADD_ON`, existing bundle should one `BASE` -- and potentially other `ADD_ON`, we could skip this validation part. In this case, the existing logic around cancellation should remain.

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApi.java
```diff
@@ -845,4 +845,45 @@ public void testCreateBaseSubscriptionsWithAddOnsBadOrdering() throws AccountApi
...
+        final Entitlement anotherStandaloneEntitlement = entitlementApi.addEntitlement(baseEntitlement.getBundleId(), spec1, null, initialDate, initialDate, false, ImmutableList.<PluginProperty>of(), callContext);
+        assertListenerStatus();
+
+        assertEquals(anotherStandaloneEntitlement.getAccountId(), account.getId());
+        assertEquals(anotherStandaloneEntitlement.getExternalKey(), account.getExternalKey());
+
+        assertEquals(anotherStandaloneEntitlement.getLastActivePriceList().getName(), "notrial");
+        assertEquals(anotherStandaloneEntitlement.getLastActiveProduct().getName(), "Knife");
+        assertEquals(anotherStandaloneEntitlement.getLastActivePlan().getName(), "knife-monthly-notrial");
+        assertEquals(anotherStandaloneEntitlement.getLastActiveProductCategory(), ProductCategory.STANDALONE);
```
sbrossie: Can we add assertion the `anotherStandaloneEntitlement` has the same `bundleId` as the `baseEntitlement#bundleId` ?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
sbrossie: I would suggest to keep the same thing with do for `ADD_ON`, where we don't mix setting boolean value `createStandaloneEntitlement ` with validation of arguments:
```

final boolean createStandaloneEntitlement = ProductCategory.STANDALONE.toString().equals(entitlement.getProductCategory());
if (createStandaloneEntitlement) {
            Preconditions.checkArgument(entitlement.getExternalKey() != null || entitlement.getBundleId() != null, "SubscriptionJson bundleId or externalKey should be specified for STANDALONE");
        }
```

In the end, it seems like `Precondition` could be shared between `STANDALONE`  and `ADD_ON` ?

sbrossie: I am even wondering if one variable `isStandaloneOrAddOnEntitlement` would not be enough since they seem to share validation and end up calling same `EntitlementApi.addEntitlement` api?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
sbrossie: @wwjbatista I am not really following, i think this is fine. I'll take a look once the code is merged, and potentially tweak things if needed.

PR #946, opened 2018-04-13
 perf: change specific internal APIs to pass Catalog object

pierre:
This avoids fetching over and over the Catalog in some codepaths (especially costly when using a non-cached Catalog Plugin).


pierre:
> 1. Should it make sense to exploit more these internal apis to allow passing objects from top to bottom-- avoiding re-fetching state when going from one module to the next?

Yes, I will open a subsequent PR.

> * How do we decide which one get to be changed, and how to we ensure this stays relevant as code changes?

I would simply update them as needed, i.e. only focus on entitlement at the moment. The changes are quite consequent already...

> * Should we create a model where such parameter are optional, maybe using `Optional<Foo>` and callee decides what to do then.

I had actually tried something along these lines yesterday for the `PlanAligner`, which requires a bunch of info (e.g. bundle start date), but only a subset of that info is used depending on your alignment. While it would help optimize specific codepaths, the code started to become hard to follow / unmaintainable. If we want to go down that road, a much bigger, deeper, refactoring is needed.

Issue #937, opened 2018-04-06
New API to return all audit logs for an account


pierre:
The implementation of `AccountAuditLogs` (`DefaultAccountAuditLogs`) already pulls all audit logs for an account, but doesn't expose them.

1. Add a new method to the `AccountAuditLogs` interface:

```
public List<AuditLog> getAuditLogs()
```

2. Expose a new endpoint `/1.0/kb/accounts/<ID>/auditLogs` to return a list of `AuditLogJson`

3. Update `AuditLogJson` so it also contains the `objectId` and the `objectType`

Note that we might have to consider eventually a streaming API for very large deployments. See https://github.com/killbill/killbill/blob/13f925a667f475c34f9b4edb7dfe06ba6412e569/util/src/main/java/org/killbill/billing/util/audit/DefaultAccountAuditLogs.java#L49

I don't suggest we do it just yet though.

PR #948, opened 2018-04-16
 New API to return all audit logs for an account #937

wwjbatista:


jaxrs/src/main/java/org/killbill/billing/jaxrs/json/AuditLogJson.java
```diff
@@ -36,24 +39,31 @@
...
     @JsonCreator
     public AuditLogJson(@JsonProperty("changeType") final String changeType,
                         @JsonProperty("changeDate") final DateTime changeDate,
                         @JsonProperty("changedBy") final String changedBy,
                         @JsonProperty("reasonCode") final String reasonCode,
                         @JsonProperty("comments") final String comments,
-                        @JsonProperty("userToken") final String userToken) {
+                        @JsonProperty("userToken") final String userToken,
+                        @JsonProperty("objectType") final ObjectType objectType,
+                        @JsonProperty("objectId") final UUID objectId) {
```
sbrossie: Because we start generating api code based on swagger def, ordering of argument becomes more important to ensure that various api are constructed the same way. I started to [create a page to define some -- arbitrary-- rules](https://github.com/killbill/killbill-swagger-coden/blob/master/SCHEMA_RULES.md#kill-bill-rules):

I would suggest to move args to have:
* ...
* `changeDate`
* **`objectType`**
* **`objectId `**
* `changedBy`
* ...

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -1484,4 +1487,38 @@ public Response transferChildCreditToParent(@PathParam("childAccountId") final U
...
+     * *************************     AUDIT LOGS     *****************************
+     */
+
+    @TimedResource
+    @GET
+    @Path("/{accountId:" + UUID_PATTERN + "}/" + AUDIT_LOG)
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Retrieve audit logs by account id", response = AuditLogJson.class, responseContainer = "List")
+    @ApiResponses(value = {@ApiResponse(code = 400, message = "Invalid account id supplied"),
+                           @ApiResponse(code = 404, message = "Account not found")})
```
sbrossie: 🎉 Great to have the swagger annotation, incl response type, ...

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -1484,4 +1487,38 @@ public Response transferChildCreditToParent(@PathParam("childAccountId") final U
...
+    public Response getAccountAuditLogs(@PathParam("accountId") final UUID accountId,
+                               @javax.ws.rs.core.Context final HttpServletRequest request) throws AccountApiException {
+        final TenantContext tenantContext = context.createTenantContextWithAccountId(accountId, request);
+        final Account account = accountUserApi.getAccountById(accountId, tenantContext);
```
sbrossie: Do we actually need to fetch the `account` ?

util/src/main/java/org/killbill/billing/util/audit/DefaultAccountAuditLogs.java
```diff
@@ -135,6 +135,11 @@ public AccountAuditLogsForObjectType getAuditLogs(final ObjectType objectType) {
         return auditLogsCache.get(objectType);
     }
 
+    @Override
+    public List<AuditLog> getAuditLogs() {
+        return ImmutableList.<AuditLog>copyOf(accountAuditLogs);
```
sbrossie: It seems like we could change for free the definition of `accountAuditLogs` to be  `private final List<AuditLog> accountAuditLogs;` and avoid that copy.

PR #961, opened 2018-04-26
Entitlement fixes

pierre:
See https://github.com/killbill/killbill-integration-tests/pull/27.

### DIFF
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
@@ -38,6 +38,9 @@ import org.killbill.billing.account.api.AccountInternalApi;
 import org.killbill.billing.callcontext.InternalCallContext;
 import org.killbill.billing.callcontext.InternalTenantContext;
 import org.killbill.billing.catalog.api.BillingActionPolicy;
+import org.killbill.billing.catalog.api.Catalog;
+import org.killbill.billing.catalog.api.CatalogApiException;
+import org.killbill.billing.catalog.api.CatalogInternalApi;
 import org.killbill.billing.catalog.api.PlanPhasePriceOverride;
 import org.killbill.billing.catalog.api.PlanPhaseSpecifier;
 import org.killbill.billing.catalog.api.ProductCategory;
@@ -102,6 +105,7 @@ public class DefaultEntitlementApi extends DefaultEntitlementApiBase implements 
     private final NotificationQueueService notificationQueueService;
     private final EntitlementPluginExecution pluginExecution;
     private final SecurityApi securityApi;
+    private final CatalogInternalApi catalogInternalApi;
 
     @Inject
     public DefaultEntitlementApi(final PersistentBus eventBus, final InternalCallContextFactory internalCallContextFactory,
@@ -110,6 +114,7 @@ public class DefaultEntitlementApi extends DefaultEntitlementApiBase implements 
                                  final BlockingChecker checker, final NotificationQueueService notificationQueueService,
                                  final EventsStreamBuilder eventsStreamBuilder, final EntitlementUtils entitlementUtils,
                                  final EntitlementPluginExecution pluginExecution,
+                                 final CatalogInternalApi catalogInternalApi,
                                  final SecurityApi securityApi) {
         super(eventBus, null, pluginExecution, internalCallContextFactory, subscriptionInternalApi, accountApi, blockingStateDao, clock, checker, notificationQueueService, eventsStreamBuilder, entitlementUtils, securityApi);
         this.internalCallContextFactory = internalCallContextFactory;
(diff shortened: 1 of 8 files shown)


entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -506,8 +372,184 @@ public UUID doCall(final EntitlementApi entitlementApi, final EntitlementContext
...
+
+        final Iterator<BaseEntitlementWithAddOnsSpecifier> iterator = baseEntitlementWithAddOnsSpecifiers.iterator();
+        if (!iterator.hasNext()) {
+            throw new SubscriptionBaseApiException(ErrorCode.SUB_CREATE_INVALID_ENTITLEMENT_SPECIFIER);
+        }
+
+        return iterator.next();
+    }
+
+    private void preCheckAddEntitlement(final BaseEntitlementWithAddOnsSpecifier baseEntitlementWithAddOnsSpecifier,
```
sbrossie: Nit: This is slightly more than a preCheck as it also populate`populateEventsStreamForBaseSubscriptionPerBundleCache ` -- and  cache.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -506,8 +372,184 @@ public UUID doCall(final EntitlementApi entitlementApi, final EntitlementContext
...
+                final EventsStream eventsStreamForBaseSubscription = eventsStreamBuilder.buildForBaseSubscription(bundleId, callContext);
+                eventsStreamForBaseSubscriptionPerBundle.put(bundleId, Optional.<EventsStream>of(eventsStreamForBaseSubscription));
+            } else {
+                eventsStreamForBaseSubscriptionPerBundle.put(bundleId, Optional.<EventsStream>absent());
+            }
+        }
+    }
+
+    private void preCheckAddEntitlement(final UUID bundleId, final DateTime entitlementRequestedDate, final BaseEntitlementWithAddOnsSpecifier baseEntitlementWithAddOnsSpecifier, final EventsStream eventsStreamForBaseSubscription) throws EntitlementApiException {
+        if (eventsStreamForBaseSubscription.isEntitlementCancelled() ||
```
sbrossie: In theory we should check the cancellation date with respect to the `entitlementRequestedDate` and allow if such cancellation date is more in the future.

I think this goes into the discussion we had the other day with my fix for usage record. Ideally we would have a method which given an effective date a a subscription checks for validity. At a high level something like:
* If `requestedDate` < `startDate` -> Denied
* If `requestedDate` >= `cancelDate` -> Denied
* If `requestedDate` is during time period where subscription is not Active -> denied

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -506,8 +372,184 @@ public UUID doCall(final EntitlementApi entitlementApi, final EntitlementContext
...
+                                        final Catalog catalog,
+                                        final TenantContext callContext,
+                                        final InternalCallContext contextWithValidAccountRecordId) throws EntitlementApiException {
+        // In the addEntitlement codepath, bundleId is always set. But, technically, an existing bundle could be specified by externalKey in
+        // the createBaseEntitlementsWithAddOns codepath. In that case, we should also check if that bundle is blocked.
+        UUID bundleId = baseEntitlementWithAddOnsSpecifier.getBundleId();
+        if (bundleId == null && baseEntitlementWithAddOnsSpecifier.getExternalKey() != null) {
+            populateBundleKeyToIdMappingCache(baseEntitlementWithAddOnsSpecifier, bundleKeyToIdMapping, catalog, contextWithValidAccountRecordId);
+
+            final Optional<UUID> bundleIdForKey = bundleKeyToIdMapping.get(baseEntitlementWithAddOnsSpecifier.getExternalKey());
```
sbrossie: Nit: Are we embracing `Optional` inside KB core?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
pierre: Fixed in 052a6a9dcdd525870f04b37fa6755a3154c73641.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
pierre: I believe this is already in place?

Here's the full condition:

```
if (eventsStreamForBaseSubscription.isEntitlementCancelled() ||
     eventsStreamForBaseSubscription.isEntitlementPending() &&
     (baseEntitlementWithAddOnsSpecifier.getEntitlementEffectiveDate() == null ||
              baseEntitlementWithAddOnsSpecifier.getEntitlementEffectiveDate().compareTo(eventsStreamForBaseSubscription.getEntitlementEffectiveStartDate()) < 0))) {
            throw new EntitlementApiException(ErrorCode.SUB_GET_NO_SUCH_BASE_SUBSCRIPTION, bundleId);
}
```

Note that `isEntitlementCancelled` means we are past the effective cancellation date.

PR #962, opened 2018-04-26
util: always redirect bus events and notifications to RW DB instance

pierre:
RO DBI is now invoked only for JAX-RS `GET` endpoints and when plugins specifically ask for it.

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/db/TestDBRouterAPI.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/db/TestDBRouterAPI.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2014-2018 Groupon, Inc
+ * Copyright 2014-2018 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.killbill.billing.beatrix.integration.db;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import javax.inject.Inject;
+
+import org.killbill.billing.KillbillApi;
+import org.killbill.billing.util.callcontext.CallContext;
+import org.killbill.billing.util.callcontext.TenantContext;
+import org.killbill.billing.util.entity.dao.DBRouterUntyped;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+import org.skife.jdbi.v2.Handle;
+import org.skife.jdbi.v2.IDBI;
+
+public class TestDBRouterAPI implements KillbillApi {
+
+    private final AtomicInteger rwCalls = new AtomicInteger(0);
+    private final AtomicInteger roCalls = new AtomicInteger(0);
+
+    private final DBRouterUntyped dbRouter;
+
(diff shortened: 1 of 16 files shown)


profiles/killbill/src/main/java/org/killbill/billing/server/modules/JaxRSAopModule.java
```diff
@@ -0,0 +1,98 @@
...
+                                                            logger.debug("Entering JAX-RS call {}, arguments: {}", invocation.getMethod(), invocation.getArguments());
+                                                            final Object proceed = invocation.proceed();
+                                                            logger.debug("Exiting  JXA-RS call {}, returning: {}", invocation.getMethod(), proceed);
+                                                            return proceed;
+                                                        }
+                                                    });
+        }
+
+        private boolean isRODBIAllowed(final MethodInvocation invocation) {
+            return invocation.getMethod().getAnnotation(GET.class) != null;
```
sbrossie: Questions/issues ?

* Are we really guaranteed that `GET` operation never end up writing state ? For instance, in a path of retrieving payments, could not we end up calling Janitor logic, then completing payment ?
* What would happen if a plugin gets invoked during get -- e.g PaymentApiType.NOTIFICATION. What would then happen to this thread ?

sbrossie: Also, conversely, less important, but calls like dryRun invoice, currently implemented using `POST` would not benefit from this logic.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/db/TestDBRouterAPI.java
```diff
@@ -0,0 +1,93 @@
...
+    public void doRWCall(final CallContext callContext) {
+        dbRouter.getHandle(false);
+    }
+
+    public void doROCall(final TenantContext tenantContext) {
+        dbRouter.getHandle(true);
+    }
+
+    // Nesting dolls
+    public void doChainedROCall(final TenantContext tenantContext) {
```
sbrossie: Confused about this one. I would expect a RO call calling another one (series of) RO calls?

pierre: Each API call is wrapped by the `ProfilingMethodInterceptor`, even when the API is calling itself.

In other words, the stack is `handleExternalEvents` -> `ProfilingMethodInterceptor` -> `testDBRouterAPI.doChainedROCall` -> `ProfilingMethodInterceptor` -> `testDBRouterAPI.doROCall`.

profiles/killbill/src/main/java/org/killbill/billing/server/modules/JaxRSAopModule.java
pierre: > * Are we really guaranteed that `GET` operation never end up writing state ? For instance, in a path of retrieving payments, could not we end up calling Janitor logic, then completing payment ?

Nope, we aren't. That's why I have this check in the lower levels: https://github.com/killbill/killbill/pull/962/files#diff-cc1298714ab6fed9365a7fab751f800cR72.

See test: https://github.com/killbill/killbill/pull/962/files#diff-3aa8c9662fc08477a4d50d574aba0408R83.

> * What would happen if a plugin gets invoked during get -- e.g PaymentApiType.NOTIFICATION. What would then happen to this thread ?

When is `PaymentApiType.NOTIFICATION` used?

Regarding plugins called synchronously, we switch to a different thread pool (`pay-plugin-grp`) and the state isn't carried over (so it would be `RW_ONLY` by default, unless a `ROTenantContext` is used).

> Also, conversely, less important, but calls like dryRun invoice, currently implemented using `POST` would not benefit from this logic.

Indeed, that's one drawback of the new implementation (simpler, but less optimized in some edge cases).

pierre: > Regarding plugins called synchronously, we switch to a different thread pool (`pay-plugin-grp`) and the state isn't carried over (so it would be `RW_ONLY` by default, unless a `ROTenantContext` is used).

We actually don't switch for `GET`, only for RW calls. Behavior tested with 7aa63b274a1e458cca5991c618e214cb8c742668.

PR #970, opened 2018-05-02
Test stability iteration

pierre:


catalog/src/test/java/org/killbill/billing/catalog/dao/TestCatalogOverrideUsageTierSqlDao.java
```diff
@@ -35,6 +35,10 @@
 
     @BeforeClass(groups = "slow")
     public void beforeClass() throws Exception {
+        if (hasFailed()) {
+            return;
+        }
+
         super.beforeClass();
```
sbrossie: I see a lot of those changed, but since most (all ?) of these methods call `super.beforeClass();` could we not put this logic in the base class?

pierre: The base method already bails early - but since most tests sprinkle some extra initialization stuff, we need the overridden method to bail early too (and that logic has to be in the overridden method).

Surprised that there isn't an easier way to do this with TestNG (or maybe there is, couldn't figure it out though).

sbrossie: I see.. that sucks indeed. All these repetitions, just annoyed that as we write more tests we also have to remember doing that, but agree, don't see a good way to do it.

Issue #335, opened 2015-06-14
New API to retrieve history info from any objects 


sbrossie:
_See bottom comment for the design proposal_

Note: This task really belong under API, but will be more visible here

In the same way we have an [AuditLog API](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/util/api/AuditUserApi.java), we probably need something for history.

Also, we should thing whether audit and history should be returned at the same time, or are there 2 APIs.  Probably something to look at as we redesign our UI.

pierre:
See also https://github.com/killbill/killbill/issues/247.

sbrossie:
/cc @pierre A few topics for discussion:

1. Do we want streaming - like api -- [see existing audit log implementation](https://github.com/killbill/killbill/blob/master/util/src/main/java/org/killbill/billing/util/audit/dao/DefaultAuditDao.java#L73) ?
2. Do we want to also add some caching similarly to what is being done for [audit logs](https://github.com/killbill/killbill/blob/master/util/src/main/resources/ehcache.xml#L34) ?
3.  Do we want end point for history, or combine history on a per call basis, the way this is done for audit logs -- probably we should revisit the use case ?

pierre:
> 1. Do we want streaming - like api -- [see existing audit log implementation](https://github.com/killbill/killbill/blob/master/util/src/main/java/org/killbill/billing/util/audit/dao/DefaultAuditDao.java#L73) ?

I don't think we stream audit logs today: https://github.com/killbill/killbill/blob/master/util/src/main/java/org/killbill/billing/util/audit/DefaultAccountAuditLogs.java#L49

In any case, I would offer similar APIs between history and audit logs, for simplicity.

> 2. Do we want to also add some caching similarly to what is being done for [audit logs](https://github.com/killbill/killbill/blob/master/util/src/main/resources/ehcache.xml#L34) ?

I wouldn't bother: the caching for audit logs is mainly to speed up Analytics and Kaui, but I don't expect the history APIs to be as heavily used (mainly for adhoc debugging / investigations and tests).

> 3.  Do we want end point for history, or combine history on a per call basis, the way this is done for audit logs -- probably we should revisit the use case ?

I would just add a new specific endpoint, for simplicity. My thinking is that it will be used as an explicit AJAX call on some Kaui screens (➕ sign -- or similar -- to display history, and I don't even think we would want to extend the billing timeline, as it's already verbose), or as adhoc cURLs.

PR #972, opened 2018-05-03
New API to retrieve history info from any objects #335

wwjbatista:
ref. killbill/killbill-api#45
ref. killbill/killbill-commons#48


account/src/main/java/org/killbill/billing/account/dao/DefaultAccountDao.java
```diff
@@ -334,4 +334,15 @@ public Integer inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrappe
             }
         });
     }
+
+    @Override
+    public List<AuditLogWithHistory> getEmailAuditLogsWithHistoryForId(final UUID accountEmailId, final AuditLevel auditLevel, final InternalTenantContext context) throws AccountApiException {
+        return transactionalSqlDao.execute(true, AccountApiException.class, new EntitySqlDaoTransactionWrapper<List<AuditLogWithHistory>>() {
```
pierre: Can you use the other signature (without the exception)? It doesn't look like `AccountApiException` is thrown here (same remark below).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -1434,6 +1434,20 @@ public Response removeEmail(@PathParam(ID_PARAM_NAME) final UUID accountId,
         return Response.status(Status.OK).build();
     }
 
+    @TimedResource
+    @GET
+    @Path("/{accountId:" + UUID_PATTERN + "}/" + EMAILS + "/{accountEmailId:" + UUID_PATTERN + "}/" + AUDIT_LOG_WITH_HISTORY)
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Retrieve account email audit logs with history by id", response = AuditLogJson.class, responseContainer = "List")
+    @ApiResponses(value = {@ApiResponse(code = 404, message = "Account not found")})
+    public Response getAccountEmailAuditLogsWithHistory(@PathParam("accountId") final UUID accountId,
```
pierre: I don't think you need the `accountId` (same remark below).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxrsResource.java
```diff
@@ -211,6 +212,9 @@
     public static final String PAYMENT_METHODS_PATH = PREFIX + "/" + PAYMENT_METHODS;
     public static final String PAYMENT_METHODS_DEFAULT_PATH_POSTFIX = "setDefault";
 
+    public static final String PAYMENT_ATTEMPTS = "paymentAttempts";
```
pierre: Not used.

payment/src/test/java/org/killbill/billing/payment/TestJanitor.java
```diff
@@ -521,35 +526,6 @@ public Object execute() throws Throwable {
         return result;
     }
 
-    // I wish we had a simplest way to query our history rows..
```
pierre: We now have! 😺 🎉

payment/src/test/java/org/killbill/billing/payment/dao/TestPaymentDao.java
```diff
@@ -26,6 +26,7 @@
 
 import org.joda.time.DateTime;
 import org.killbill.billing.account.api.Account;
+import org.killbill.billing.account.dao.AccountModelDao;
```
pierre: Unused.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestAccount.java
```diff
@@ -28,6 +29,7 @@
 
 import org.killbill.billing.ErrorCode;
 import org.killbill.billing.ObjectType;
+import org.killbill.billing.account.dao.AccountModelDao;
```
pierre: Unused.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestAccount.java
```diff
@@ -551,4 +555,37 @@ public void testGetAccountAuditLogs() throws Exception {
...
+        final Account updatedAccount = killBillClient.updateAccount(newInput, requestOptions);
+
+
+        final List<AuditLog> auditLogWithHistories = killBillClient.getAccountAuditLogsWithHistory(accountJson.getAccountId());
+        assertEquals(auditLogWithHistories.size(), 2);
+        assertEquals(auditLogWithHistories.get(0).getChangeType(), ChangeType.INSERT.toString());
+        assertEquals(auditLogWithHistories.get(0).getObjectType(), ObjectType.ACCOUNT);
+        assertEquals(auditLogWithHistories.get(0).getObjectId(), accountJson.getAccountId());
+
+        final LinkedHashMap history1 = (LinkedHashMap) auditLogWithHistories.get(0).getHistory();
```
pierre: @sbrossie Curious how it will work with the new code (would be nice if `AuditLog<E>` can be generated).

util/src/main/java/org/killbill/billing/util/dao/HistorySqlDao.java
```diff
@@ -29,11 +29,15 @@
...
 import org.skife.jdbi.v2.sqlobject.SqlUpdate;
+import org.skife.jdbi.v2.sqlobject.customizers.Define;
 
 public interface HistorySqlDao<M extends EntityModelDao<E>, E extends Entity> {
 
+    //  bypassMappingRegistryCache is to be used in the mapping registry to bypass cache. This is useful for generic classes since it will prevent to return previously cached class, that could result in wrong maps.
+    //  https://github.com/killbill/killbill-commons/blob/work-for-release-0.19.x/jdbi/src/main/java/org/skife/jdbi/v2/MappingRegistry.java#L67
     @SqlQuery
-    public List<EntityHistoryModelDao<M, E>> getHistoryForTargetRecordId(@Bind("targetRecordId") final long targetRecordId,
+    public List<EntityHistoryModelDao<M, E>> getHistoryForTargetRecordId(@Define("bypassMappingRegistryCache") final boolean bypassMappingRegistryCache,
```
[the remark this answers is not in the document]
wwjbatista: @sbrossie, After I completed the implementation to get the audit logs with history for the other objects and started testing, I noticed that the first call was good but the others calls where bad, since bring the last history entity but with new data (only matched fields). Digging out, I notice that all the mappers are cached via its class; in the case of the history the class is `EntityHistoryModelDao`, the difference is on the parameter passed into the class, but the map is stored and retrieved from the cache via the class, thus always fetching from cache the first one that was stored. That is why I introduced a flag to decide if store the mapper on the cache or not, and I am using the `@Define("bypassMappingRegistryCache")` to pass that flag to the register.

PR #973, opened 2018-05-04
invoice: make sure we never insert notifications in the past

pierre:
Trivial fix but took me forever to write the test 😅

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationBase.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationBase.java
@@ -664,6 +664,26 @@ public class TestIntegrationBase extends BeatrixTestSuiteWithEmbeddedDB implemen
                                                                                              final BillingPeriod billingPeriod,
                                                                                              final List<PlanPhasePriceOverride> overrides,
                                                                                              final NextEvent... events) {
+        return createBaseEntitlementWithPriceOverrideAndCheckForCompletion(accountId,
+                                                                           bundleExternalKey,
+                                                                           productName,
+                                                                           productCategory,
+                                                                           billingPeriod,
+                                                                           overrides,
+                                                                           null,
+                                                                           PriceListSet.DEFAULT_PRICELIST_NAME,
+                                                                           events);
+    }
+
+    protected DefaultEntitlement createBaseEntitlementWithPriceOverrideAndCheckForCompletion(final UUID accountId,
+                                                                                             final String bundleExternalKey,
+                                                                                             final String productName,
+                                                                                             final ProductCategory productCategory,
+                                                                                             final BillingPeriod billingPeriod,
+                                                                                             final List<PlanPhasePriceOverride> overrides,
+                                                                                             final LocalDate billingEffectiveDate,
+                                                                                             final String priceList,
+                                                                                             final NextEvent... events) {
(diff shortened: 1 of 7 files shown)


pierre:
> The main risk is that we also plugged that fix for non dry run scenario, so could we hit cases where notification in the past is actually a useful thing ?

My understanding is that we schedule this notification in the past while processing an invoice generation. At the end of that invoice run, we should be up-to-date (for a given target date), so re-scheduling a notification immediately should be a no-op.

Now, in case the target date in the notification is different, it's another story... I'll think about it to see if/how it's possible.

pierre:
> The main risk is that we also plugged that fix for non dry run scenario, so could we hit cases where notification in the past is actually a useful thing ?

tl;dr Indeed, I think there are cases where this could happen. For safety, I'll revert the changes for the non-dry-run scenario (see c34d9eba1d15adcded61526c6d720f10f94144ae).


My thinking is the following: if an invoice is generated at T2, and we try to insert a notification for T1 where T1 < T2, for the bug to happen (missing valid invoice generation), something has to happen between T1 and T2 (or the account would be up to date).
This can't be entitlement/subscription related (or we would have an effective bus event, forcing an invoice generation anyways). So this can only happen for recurring notifications (no external trigger).

To write a test to trigger this, I've tried tricking the system with DST gaps, but the system seems smart enough nowadays to figure it out... However, it looks like it can be triggered if the bus (or notification queue) is late.

This test can trigger it if the VM is paused after creating the base entitlement (to introduce a delay):

```
@Test(groups = "slow")
public void testIntegrationWithNotificationInThePastSameDay() throws Exception {
    // Set reference time to midnight
    final DateTime initialCreationDate = new DateTime(2018, 3, 24, 0, 0, 0, 0, testTimeZone);
    clock.setTime(initialCreationDate);

    final Account account = createAccountWithNonOsgiPaymentMethod(getAccountData(24));

    // Entitlement creation at 11:59:59pm (invoice generation will have a 2018-03-24 targetDate)
    clock.setTime(new DateTime(2018, 3, 24, 23, 59, 59, 0, testTimeZone));
    assertListenerStatus();

    busHandler.pushExpectedEvents(NextEvent.CREATE, NextEvent.BLOCK, NextEvent.INVOICE, NextEvent.INVOICE_PAYMENT, NextEvent.PAYMENT);
    final PlanPhaseSpecifier spec = new PlanPhaseSpecifier("Blowdart", BillingPeriod.DAILY, "notrial", null);
    final UUID entitlementId = entitlementApi.createBaseEntitlement(account.getId(),
                                                                    spec,
                                                                    UUID.randomUUID().toString(),
                                                                    null,
                                                                    null,
                                                                    null,
                                                                    false,
                                                                    false,
                                                                    ImmutableList.<PluginProperty>of(),
                                                                    callContext);
    // If the bus is slow, next notification date could be in the past (2018-03-25T00:00:00.000Z)
    assertListenerStatus();

    // Verify next (daily) invoice is generated at the reference time
    busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.INVOICE_PAYMENT, NextEvent.PAYMENT);
    clock.setTime(new DateTime(2018, 3, 25, 0, 0, 0, 0, testTimeZone));
    assertListenerStatus();
}
```

Unfortunately, I haven't been able to make the test fail consistently (I think the only way to write it would be to change our bus code to introduce hooks to force a slowdown), so I won't check it in.

Issue #978, opened 2018-05-16
Unable to change to a plan defined in a subsequent catalog version


pierre:
Scenario triggering the issue:

1. Catalog effective at T0
2. Subscription created at T1
3. Update catalog at T1, adding a new plan
4. Change subscription to new plan

Step 4. fails as the catalog cannot find the new plan.

Issue #784, opened 2017-08-02
Catalog APIs review


pierre:
* Calling these APIs is difficult and error prone (https://github.com/killbill/killbill/issues/56#issuecomment-135749604, https://github.com/killbill/killbill-analytics-plugin/blob/6c55487b401135a2ec06ca16774f4a3ceac0a1bb/src/main/java/org/killbill/billing/plugin/analytics/dao/factory/BusinessFactoryBase.java#L381)
* Some APIs are missing dates for grand-fathering (https://github.com/killbill/killbill/blob/4320ce9106e78f9ce533c4d21d214bf77cc070c3/subscription/src/main/java/org/killbill/billing/subscription/alignment/PlanAligner.java#L204)

sbrossie:
The issue has resurfaced recently on the [mailing list](https://groups.google.com/forum/#!searchin/killbilling-users/Changing$20Billing$20Alignment$20From$20Account$20to$20Bundle$2FSubscription%7Csort:date/killbilling-users/bWwQ_049pNw/VuEeLDoCAQAJ).

There are 2 use cases for when fetching catalog information:
1. Given a date, return a given catalog entity (e.g `Plan`). The date is used to find the active version of this catalog. Often the date is NOW, and we return latest version.
2. Given an existing subscription that was started in the past and a date, return the *effective* catalog entity (e.g `Plan` associated to this `Subscription` on that date).  This is based on a complicated logic between subscription startDate and potential `effectiveDateForExistingSubscriptions` catalog attribute

We have all the apis to return catalog entities based on date, use case `1.`, but we only have 4 apis for use case `2.`:
* Retrieve a `Plan` associated to a given subscription: [findPlan](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/catalog/api/Catalog.java#L94) and [createOrFindPlan](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/catalog/api/Catalog.java#L103)
* Retrieve a `PriceList` associated to a given subscription: [findPriceListForPlan](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/catalog/api/Catalog.java#L129)
* Retrieve a `PlanPhase` associated to a given subscription: [findPhase](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/catalog/api/Catalog.java#L140)

[...]

PR #979, opened 2018-05-16
catalog: ability to change to a plan defined in a subsequent catalog version

pierre:
This fixes https://github.com/killbill/killbill/issues/978.

See also https://github.com/killbill/killbill-integration-tests/pull/28.

### DIFF
--- a/catalog/src/main/java/org/killbill/billing/catalog/VersionedCatalog.java
+++ b/catalog/src/main/java/org/killbill/billing/catalog/VersionedCatalog.java
@@ -160,20 +160,22 @@ public class VersionedCatalog extends ValidatingConfig<VersionedCatalog> impleme
 
     private CatalogPlanEntry findCatalogPlanEntry(final PlanRequestWrapper wrapper,
                                                   final DateTime requestedDate,
-                                                  final DateTime subscriptionStartDate)
-            throws CatalogApiException {
+                                                  final DateTime subscriptionStartDate) throws CatalogApiException {
         final List<StandaloneCatalog> catalogs = versionsBeforeDate(requestedDate.toDate());
         if (catalogs.isEmpty()) {
             throw new CatalogApiException(ErrorCode.CAT_NO_CATALOG_FOR_GIVEN_DATE, requestedDate.toDate().toString());
         }
 
+        CatalogPlanEntry candidateInSubsequentCatalog = null;
         for (int i = catalogs.size() - 1; i >= 0; i--) { // Working backwards to find the latest applicable plan
             final StandaloneCatalog c = catalogs.get(i);
+
             final Plan plan;
             try {
                 plan = wrapper.findPlan(c);
             } catch (final CatalogApiException e) {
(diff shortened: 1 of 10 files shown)


catalog/src/main/java/org/killbill/billing/catalog/VersionedCatalog.java
```diff
@@ -182,21 +184,29 @@ private CatalogPlanEntry findCatalogPlanEntry(final PlanRequestWrapper wrapper,
...
+                    // Keep the most recent one
+                    candidateInSubsequentCatalog = new CatalogPlanEntry(c, plan);
                 }
             }
         }
 
+        if (candidateInSubsequentCatalog != null) {
+            return candidateInSubsequentCatalog;
+        }
+
```
sbrossie: What you have implemented should work as a fix but fundamentally this api is overloaded with respect to various use cases. At least those two categories:

- Billing operation where we need to find the right catalog entry (and therefore the right price): In this scenario, going through versions from most recent to oldest and taking into account `effectiveDateForExistingSubscriptions ` make sense
- For other subscription scenario -- let's say change Plan-- the logic is much simpler, we should just return the entry from the catalog version matching the requested date, i.e `catalogs.get(0). findPlan()` 

I am now wondering if the fix is not to call `public Plan createOrFindPlan(final PlanSpecifier spec,
                                 final PlanPhasePriceOverridesWithCallContext overrides,
                                 final DateTime requestedDate)` [here](https://github.com/killbill/killbill/blob/master/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java#L411) instead of  `public Plan createOrFindPlan(final PlanSpecifier spec,
                                 final PlanPhasePriceOverridesWithCallContext overrides,
                                 final DateTime requestedDate,
                                 final DateTime subscriptionStartDate)` ?

pierre: Agreed - I've made the change and verified that the APIs without the `subscriptionStartDate` are always called for subscription operations, except for internal billing operations (see also https://github.com/killbill/killbill/issues/784#issuecomment-390244107).

This is mainly true today, except for `rebuildTransitions`: https://github.com/killbill/killbill/blob/5d63b9b5e88f2c603046b9f60139a93caee7fbbd/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java#L767-L769 which is called in a lot of code paths (including `doChangePlan`). Because of this, I also need to keep my original fix, or the bug is still present: if `getEffectiveDateForExistingSubscriptions` returns `null` (not specified for a new introduced plan for instance), `findCatalogPlanEntry` never finds anything otherwise.

Intuitively anyways, regardless of whether the right API is used or not, it seems that the implementation of this method should not assume that `effectiveDateForExistingSubscriptions` is set. It would be interesting to write a test for instance with V0 (price P), V1 (price P1) and V2 (price P2) where `effectiveDateForExistingSubscriptions` is set only in V1, and see which plan is returned when computing billing events at different times.

PR #982, opened 2018-05-21
Remove validation for charge amount less than or equal to zero while …

sruthipendyala:
creating external charge item.. 
We would like to show users $0 external charge line items.
Removed charge amount <0 validation as we would like to have the ability to add credit line item to an invoice without adjusting a line item. Please let me know if that works or if I need to do anything else!


sruthipendyala:
That is perfect...CREDIT_ADJ line item is better for sure.. I totally missed the API call.
I will update the pull request :)
Thank you!

sruthipendyala:
hey @pierre Also noticed that linkedinvoiceitemid was not being set when creating an external charge line item using the API. Updated pull request with the same..

PR #984, opened 2018-05-23
Add ability to display rate and quantity for invoice line items when …

sruthipendyala:
…using Retrieve an invoice by id API call

We set rate and quantity while creating Adjustment and External Charge Line Items. We would like to have the ability to get that information as well. Please let me know if my changes work or if I need to do anything else :)


sruthipendyala:
Updated pull request with the ordering of arguments.
I tested for the scenario where we set ITEM_ADJ using the plugin and it works fine.
However I realized that quantity and rate are not set while creating an EXTERNAL_CHARGE line item :(

sruthipendyala:
Updated pull request to set quantity while creating an EXTERNAL_CHARGE line item as well.
Also tested scenario where rate, quantity are returned while trying to GET external charge line items..

PR #986, opened 2018-05-24
Integrate EurekaModule for service discovery

sethdaugherty:
This contribution allows KB to register with a Eureka server (https://github.com/Netflix/eureka/) for automated service discovery. The new Guice EurekaModule isn't included in the default `killbill` profile, so this is disabled by default.

I'm unclear if I need to also submit a PR to update the documentation. Should I?

For context, much of this code is based on the Guice module Netflix provides for Eureka: 
https://github.com/Netflix/eureka/blob/master/eureka-client/src/main/java/com/netflix/discovery/guice/EurekaModule.java

### DIFF
--- a/util/pom.xml
+++ b/util/pom.xml
@@ -74,6 +74,11 @@
             <artifactId>guice-multibindings</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.netflix.eureka</groupId>
+            <artifactId>eureka-client</artifactId>
+            <version>1.8.7</version>
+        </dependency>
+        <dependency>
             <groupId>com.ning</groupId>
             <artifactId>async-http-client</artifactId>
         </dependency>
@@ -85,6 +90,11 @@
         <dependency>
             <groupId>com.samskivert</groupId>
             <artifactId>jmustache</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.sun.jersey</groupId>
+            <artifactId>jersey-client</artifactId>
+            <version>1.19.4</version>
         </dependency>
         <dependency>
             <groupId>com.zaxxer</groupId>
--- a/util/src/main/java/org/killbill/billing/util/discovery/DefaultEurekaKillbillService.java
+++ b/util/src/main/java/org/killbill/billing/util/discovery/DefaultEurekaKillbillService.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2014-2018 Groupon, Inc
+ * Copyright 2014-2018 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.killbill.billing.util.discovery;
+
+
+import com.google.inject.Inject;
+import com.netflix.appinfo.ApplicationInfoManager;
+import com.netflix.appinfo.InstanceInfo;
+
+import org.killbill.billing.platform.api.KillbillService;
+import org.killbill.billing.platform.api.LifecycleHandlerType;
+import org.killbill.billing.platform.api.LifecycleHandlerType.LifecycleLevel;
+import org.killbill.billing.platform.config.DefaultKillbillConfigSource;
+import org.killbill.billing.tenant.api.TenantKV.TenantKey;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultEurekaKillbillService implements EurekaKillbillService {
+
+    private static final Logger logger = LoggerFactory.getLogger(DefaultEurekaKillbillService.class);
+
+    private ApplicationInfoManager applicationInfoManager;
+
+    public static final String EUREKA_SERVICE_NAME = "eureka-service";
+
+    @Inject
+    public DefaultEurekaKillbillService(ApplicationInfoManager applicationInfoManager) {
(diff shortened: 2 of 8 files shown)


util/pom.xml
```diff
@@ -86,6 +91,11 @@
             <groupId>com.samskivert</groupId>
             <artifactId>jmustache</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.sun.jersey</groupId>
```
pierre: Similar remark: Kill Bill itself is a library at its core and doesn't depend on `com.sun.jersey`. Very large companies, which have their own in-house Java container, don't use the Kill Bill profiles, but instead wire up the Kill Bill library jars.

util/src/main/java/org/killbill/billing/util/discovery/provider/KillbillEurekaInstanceInfoProvider.java
```diff
@@ -0,0 +1,128 @@
...
+
+            if (vipAddressResolver == null) {
+                vipAddressResolver = new Archaius1VipAddressResolver();
+            }
+
+            // Builder the instance information to be registered with eureka server
+            InstanceInfo.Builder builder = InstanceInfo.Builder.newBuilder(vipAddressResolver);
+
+            // set the appropriate id for the InstanceInfo, falling back to datacenter Id if applicable, else hostname
+            String instanceId = config.getInstanceId();
```
pierre: I think we should use the [CreatorName](https://github.com/killbill/killbill-commons/blob/a2f5ea6effc2fa256fe8ace9fb968252c2caf00e/queue/src/main/java/org/killbill/CreatorName.java#L35) instead. This is how instances identifiers are defined today when deploying in multiple datacenters or using AWS Auto Scaling.

FWIW, we also have a `region` property (https://github.com/killbill/killbill-platform/blob/7039977db0b9618a09454673b983c699bbc16740/server/src/main/java/org/killbill/billing/server/config/KillbillServerConfig.java#L44-L47).

util/src/main/java/org/killbill/billing/util/discovery/DefaultEurekaKillbillService.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public String getName() {
+        return EUREKA_SERVICE_NAME;
+    }
+
+    @LifecycleHandlerType(LifecycleLevel.INIT_SERVICE)
+    public synchronized void initialize() {
+        logger.info("--------------> Setting Eureka Status to UP <----------------");
+        applicationInfoManager.setInstanceStatus(InstanceInfo.InstanceStatus.UP);
```
pierre: `INIT_SERVICE` is very early in the initialization sequence (https://github.com/killbill/killbill-platform/blob/7039977db0b9618a09454673b983c699bbc16740/platform-api/src/main/java/org/killbill/billing/platform/api/LifecycleHandlerType.java#L59), Kill Bill wouldn't be ready to accept requests yet.

Also, I don't think a new Kill Bill service is the way to go for this. Instead, this should happen at the end of the Guice initialization sequence, when we enable the healthcheck: https://github.com/killbill/killbill-platform/blob/7039977db0b9618a09454673b983c699bbc16740/server/src/main/java/org/killbill/billing/server/filters/KillbillGuiceFilter.java#L50-L52

util/src/main/java/org/killbill/billing/util/discovery/provider/KillbillEurekaInstanceInfoProvider.java
```diff
@@ -0,0 +1,128 @@
...
+                                       config.getHealthCheckUrl(), config.getSecureHealthCheckUrl());
+
+
+            // Start off with the STARTING state to avoid traffic
+            if (!config.isInstanceEnabledOnit()) {
+                InstanceStatus initialStatus = InstanceStatus.STARTING;
+                LOG.info("Setting initial instance status as: " + initialStatus);
+                builder.setStatus(initialStatus);
+            } else {
+                LOG.info("Setting initial instance status as: {}. This may be too early for the instance to advertise "
```
pierre: I am indeed wondering if this shouldn't be more tightly integrated with our healthcheck. For instance, when calling `KillbillHealthcheck.putOutOfRotation`, you would probably send `OUT_OF_SERVICE` to Eureka?

One way to do this could be to have an optionally injected field in `KillbillHealthcheck`: `List<ServiceDiscovery>` where `ServiceDiscovery` is a generic interface which exposes a few methods for broadcasting (we can get inspired by `ApplicationInfoManager`). When injected, the healthcheck would also call the service discoveries methods (one implementation would be the Eureka one) and you wouldn't even need to worry about initialization/shutdown sequence (already done for you, and even controllable by JMX).

If we go down that road, we could also move the Eureka-specific code to a new Maven module in `killbill-platform`, which would solve some dependencies issues mentioned above (Kill Bill wouldn't have any dependency on Eureka, unless you add that jar in your webapp). That being said, in your environment, the dependencies conflict do exist and you have no idea at runtime which jar is being used 😬Best would probably be to have that new module produce a shaded uber jar.

(/cc @sbrossie)

util/src/main/java/org/killbill/billing/util/discovery/provider/KillbillEurekaInstanceInfoProvider.java
```diff
@@ -0,0 +1,128 @@
...
+import com.netflix.appinfo.InstanceInfo.InstanceStatus;
+import com.netflix.appinfo.InstanceInfo.PortType;
+import com.netflix.appinfo.LeaseInfo;
+import com.netflix.appinfo.UniqueIdentifier;
+import com.netflix.appinfo.providers.Archaius1VipAddressResolver;
+import com.netflix.appinfo.providers.EurekaConfigBasedInstanceInfoProvider;
+import com.netflix.appinfo.providers.VipAddressResolver;
+
+@Singleton
+public class KillbillEurekaInstanceInfoProvider extends EurekaConfigBasedInstanceInfoProvider {
```
pierre: Looking at the base class, I'm not even sure you need that `extends`?

sethdaugherty:
Thanks @pierre these are all really good suggestions. I moved the eureka code into a new module inside killbill-platform. I'm going to close this PR in lieu of https://github.com/killbill/killbill-platform/pull/33 

(Sorry it took so long. Kept getting distracted by other things.)

Issue #980, opened 2018-05-16
AccountResource#payAllInvoices improvements


pierre:
We should consider:

* Allowing the user to specify a `paymentMethodId`
* Allowing the user to override the `targetDate` when calling `getUnpaidInvoicesByAccountId`

Clients to regenerate and Slate docs to update.

PR #988, opened 2018-05-29
AccountResource#payAllInvoices improvements

wwjfrodriguez:
[Reff](https://github.com/killbill/killbill/issues/980)

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
@@ -807,7 +807,9 @@ public class AccountResource extends JaxRsResourceBase {
             final BigDecimal amountToPay = (remainingRequestPayment.compareTo(invoice.getBalance()) >= 0) ?
                                            invoice.getBalance() : remainingRequestPayment;
             if (amountToPay.compareTo(BigDecimal.ZERO) > 0) {
-                final UUID paymentMethodId = externalPayment ? account.getPaymentMethodId() : inputPaymentMethodId;
+                final UUID paymentMethodId = externalPayment ?
+                                             null :
+                                             (inputPaymentMethodId != null ? inputPaymentMethodId : account.getPaymentMethodId());
                 createPurchaseForInvoice(account, invoice.getId(), amountToPay, paymentMethodId, externalPayment, null, null, pluginProperties, callContext);
             }
             remainingRequestPayment = remainingRequestPayment.subtract(amountToPay);
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
@@ -791,13 +791,8 @@ public class AccountResource extends JaxRsResourceBase {
 
         final Account account = accountUserApi.getAccountById(accountId, callContext);
 
-        LocalDate inputDate;
-        if(targetDate == null){
-            inputDate = clock.getUTCToday();
-        }
-        else{
-            inputDate = toLocalDate(targetDate);
-        }
+        final LocalDate inputDate = targetDate == null ? clock.getUTCToday() : toLocalDate(targetDate);
+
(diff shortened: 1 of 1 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -787,7 +790,15 @@ public Response payAllInvoices(@PathParam("accountId") final UUID accountId,
...
-        final Collection<Invoice> unpaidInvoices = invoiceApi.getUnpaidInvoicesByAccountId(account.getId(), clock.getUTCToday(), callContext);
+
+        LocalDate inputDate;
+        if(targetDate == null){
+            inputDate = clock.getUTCToday();
+        }
+        else{
+            inputDate = toLocalDate(targetDate);
+        }
+        final Collection<Invoice> unpaidInvoices = invoiceApi.getUnpaidInvoicesByAccountId(account.getId(), inputDate, callContext);
```
sbrossie: Nit: formatting:
* Add space between words, so for example  `if` and `(` becomes `if (`, or `else` `{` becomes `else {`
* `else` is on the same line as `{`

Also, you could simplify this expression with following:
`final LocalDate inputDate = targetDate == null ? clock.getUTCToday() : toLocalDate(targetDate);`

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -801,7 +812,14 @@ public Response payAllInvoices(@PathParam("accountId") final UUID accountId,
...
                                            invoice.getBalance() : remainingRequestPayment;
             if (amountToPay.compareTo(BigDecimal.ZERO) > 0) {
-                final UUID paymentMethodId = externalPayment ? null : account.getPaymentMethodId();
+                UUID paymentMethodId;
+                if(inputPaymentMethodId == null){
+                    paymentMethodId = externalPayment ? null : account.getPaymentMethodId();
+                }
+                else{
+                    paymentMethodId = inputPaymentMethodId;
+                }
```
sbrossie: Nit: Same formatting remarks as above.

sbrossie: Also, by default if `externalPayment` is not set, the code will assume a default of false; therefore, it would make sense to assume that if this is true, this was intentional, and you could rewrite it:
```
final UUID paymentMethodId = externalPayment ? account.getPaymentMethodId() : inputPaymentMethodId;
```

wwjfrodriguez: Ok, Thanks for the tips.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -801,7 +807,7 @@ public Response payAllInvoices(@PathParam("accountId") final UUID accountId,
             final BigDecimal amountToPay = (remainingRequestPayment.compareTo(invoice.getBalance()) >= 0) ?
                                            invoice.getBalance() : remainingRequestPayment;
             if (amountToPay.compareTo(BigDecimal.ZERO) > 0) {
-                final UUID paymentMethodId = externalPayment ? null : account.getPaymentMethodId();
+                final UUID paymentMethodId = externalPayment ? account.getPaymentMethodId() : inputPaymentMethodId;
```
sbrossie: Actually, this part is wrong -- my bad!! This became obvious as i reviewed the current diff: When `externalPayment` is null, we pass a null `paymentMethodId` and the flag `externalPayment` to the api. So you should have the following instead:

```
UUID paymentMethodId = null;
if (!externalPayment) {
    paymentMethodId = inputPaymentMethodId != null ? inputPaymentMethodId : account.getPaymentMethodId();
}
```

OR if using ternary forms all the way:

```
final UUID paymentMethodId = externalPayment ?
                             null :
                             (inputPaymentMethodId != null ? inputPaymentMethodId : account.getPaymentMethodId());
```

PR #989, opened 2018-05-30
Usage with 0 price items

sbrossie:


sbrossie:
@pierre I think so: 
https://github.com/killbill/killbill/blob/usage_with_0_price_items/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java#L113

PR #995, opened 2018-06-08
Fix for 827

sbrossie:
TODO: Regenerate clients, add a profile test and update slack

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
@@ -442,6 +442,7 @@ public class InvoiceResource extends JaxRsResourceBase {
     public Response adjustInvoiceItem(@PathParam("invoiceId") final UUID invoiceId,
                                       final InvoiceItemJson json,
                                       @QueryParam(QUERY_REQUESTED_DT) final String requestedDateTimeString,
+                                      @QueryParam(QUERY_PLUGIN_PROPERTY) final List<String> pluginPropertiesString,
                                       @HeaderParam(HDR_CREATED_BY) final String createdBy,
                                       @HeaderParam(HDR_REASON) final String reason,
                                       @HeaderParam(HDR_COMMENT) final String comment,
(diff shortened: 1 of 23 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
```diff
@@ -525,6 +524,50 @@ public InvoiceItemJson apply(final InvoiceItem input) {
...
+                                   @QueryParam(QUERY_PLUGIN_PROPERTY) final List<String> pluginPropertiesString,
+                                   @HeaderParam(HDR_CREATED_BY) final String createdBy,
+                                   @HeaderParam(HDR_REASON) final String reason,
+                                   @HeaderParam(HDR_COMMENT) final String comment,
+                                   @javax.ws.rs.core.Context final HttpServletRequest request,
+                                   @javax.ws.rs.core.Context final UriInfo uriInfo) throws AccountApiException, InvoiceApiException {
+        verifyNonNullOrEmpty(taxItemJson, "Body should be specified");
+        verifyNonNullOrEmpty(accountId, "AccountId needs to be set");
+
+        // TODO do we actually support passing that to invoice for invoice plugins  ?
```
pierre: Would be good to verify at least that we are all set from an API perspective.

sbrossie: Exactly, and we are not. For instance [existing api](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/invoice/api/InvoiceUserApi.java#L190) for external charge, does not allow to pass these properties. So that's the question: Should we extend now, or decide we don't care controlling such plugins using properties ?

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestInvoice.java
```diff
@@ -596,6 +596,36 @@ public void testExternalChargeForBundleOnNewInvoice() throws Exception {
...
+        taxItem.setCurrency(accountJson.getCurrency());
+        taxItem.setBundleId(bundleId);
+        final InvoiceItems input = new InvoiceItems();
+        input.add(taxItem);
+        final List<InvoiceItem> createdTaxItems = invoiceApi.createTaxItems(accountJson.getAccountId(), input, true, clock.getUTCToday(), NULL_PLUGIN_PROPERTIES, requestOptions);
+        assertEquals(createdTaxItems.size(), 1);
+        final Invoice invoiceWithItems = invoiceApi.getInvoice(createdTaxItems.get(0).getInvoiceId(), true, null, AuditLevel.NONE, requestOptions);
+        assertEquals(invoiceWithItems.getBalance().compareTo(taxAmount), 0);
+        assertEquals(invoiceWithItems.getItems().size(), 1);
+        assertEquals(invoiceWithItems.getItems().get(0).getBundleId(), bundleId);
```
pierre: Nit: check the invoice item type.

sbrossie: Ack, will do.

Issue #1015, opened 2018-06-27
Possible double payments with invoice payments in UNKNOWN state


pierre:
Scenario:

1. Invoice payment in `UNKNOWN` state
2. Trigger a new successful payment by calling `invoicePaymentApi.createPurchaseForInvoicePayment`
3. Fix the original payment (`UNKNOWN -> SUCCESS`)

Result: the invoice ends up double payed and the invoice balance is negative (absolute value is the amount of the payment).

PR #1016, opened 2018-06-28
payment: abort invoice payment if existing payment is UNKNOWN

pierre:
This fixes https://github.com/killbill/killbill/issues/1015.


payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
```diff
@@ -359,6 +359,16 @@ private PriorPaymentControlResult getPluginPurchaseResult(final PaymentControlCo
                 return new DefaultPriorPaymentControlResult(true);
             }
 
+            final List<InvoicePayment> existingInvoicePayments = invoiceApi.getInvoicePaymentsByInvoice(invoiceId, internalContext);
+            for (final InvoicePayment existingInvoicePayment : existingInvoicePayments) {
```
sbrossie: Would it make sense to optimize and skip any `isSuccess` `existingInvoicePayment ` (to avoid fetching all associated payment transactions in that case?

pierre: I don't think it's worth the optimization: 99.99% of the time, `existingInvoicePayments` is empty.

Issue #996, opened 2018-06-12
Swagger UI upgrade


sbrossie:
We have an older version of the swagger ui embedded into killbill core. The work is to upgrade the version to a more recent one -- and verify this still works as expected.

PR #1019, opened 2018-06-29
Swagger UI upgrade #996

wwjbatista:
image


pierre:
Verified! 🎉 

Unfortunately, it looks like the changes break the build: the `apache-rat-plugin` never completes its checks. Maybe new files need to be excluded in the base `pom.xml`? You can reproduce it via:

```
cd profiles/killbill
mvn install -Ptravis -DskipTests=true
```

Issue #1032, opened 2018-08-01
Non deterministic behavior in overdue


sbrossie:
The overdue system listens to a few events to decide wether state should be recomputed. In particular it listens to invoice events, and because we have no ordering across event handlers, we have no guarantee on whether the payment subsystem -- also listening to invoice events --  or the overdue subsystem will process the event first.

This can create some non deterministic behavior as the one reported on the [mailing list](https://groups.google.com/forum/#!searchin/killbilling-users/Using$20timeSinceEarliestUnpaidInvoiceEqualsOrExceeds%7Csort:date/killbilling-users/sXgfKLgbUTw/U_h_AuXoBgAJ)

Mostly the issue could arise when the overdue configuration is set with a `timeSinceEarliestUnpaidInvoiceEqualsOrExceeds=0`

sbrossie:
See also #782

PR #1034, opened 2018-08-03
Introduce new ordering among Killbill Services to ensure initializati…

sbrossie:
…on of services is deterministic and follow bus event propagation. See #1032

In particular we are hoping to fix well known issues where various killbill services listen to same events and based on which bus event was registered
lead to different results -- e.g Payment service should process invoice bus events prior Overdue service.


util/src/test/java/org/killbill/billing/api/TestApiListener.java
```diff
@@ -63,7 +63,7 @@
 
     private static final Joiner SPACE_JOINER = Joiner.on(" ");
 
-    private static final long DELAY = 60000;
+    private static final long DELAY = 6000000;
```
pierre: Expected change?

sbrossie: oupsy... wil fix 😁

Issue #1042, opened 2018-09-21
Invoice dry run (TARGET_DATE mode) computes wring invoice in far future


sbrossie:
The issue was reported on the [mailing list](https://groups.google.com/forum/#!topic/killbilling-users/Byp-Kw7rb-8).

In the `TARGET_DATE` mode we expect the system to generate invoices as they would be generated, where we let the system generate it on the given date. In particular, we don't expect such dry-run invoice to merge past invoices.

When running the operation in the far future, the system does not correctly remove items from previous (non existent) invoices.

PR #1043, opened 2018-09-21
Invoice dry run (TARGET_DATE mode) computes wring invoice in far future

pierre:
See https://github.com/killbill/killbill/issues/1042.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
@@ -28,6 +28,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.PriorityQueue;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.UUID;
@@ -120,6 +121,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Ordering;
+import com.google.common.collect.Sets;
 import com.google.inject.Inject;
 
 public class InvoiceDispatcher {
@@ -340,7 +342,8 @@ public class InvoiceDispatcher {
                                                                                                         }));
             final Invoice invoice;
             if (!isDryRun) {
-                invoice = processAccountWithLockAndInputTargetDate(accountId, inputTargetDate, billingEvents, existingInvoices, false, isRescheduled, context);
+                final InvoiceWithFutureNotifications invoiceWithFutureNotifications = processAccountWithLockAndInputTargetDate(accountId, inputTargetDate, billingEvents, existingInvoices, false, isRescheduled, context);
+                invoice = invoiceWithFutureNotifications != null ? invoiceWithFutureNotifications.getInvoice() : null;
                 if (parkedAccount) {
                     try {
                         log.info("Illegal invoicing state fixed for accountId='{}', unparking account", accountId);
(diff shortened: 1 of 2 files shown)


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -448,28 +452,36 @@ private Invoice processDryRun_UPCOMING_INVOICE_FILTERING_Invoice(final UUID acco
...
-    private Invoice processDryRun_TARGET_DATE_Invoice(final UUID accountId, final LocalDate targetDate, final List<LocalDate> allCandidateTargetDates, final BillingEventSet billingEvents, final List<Invoice> existingInvoices, final InternalCallContext context) throws InvoiceApiException {
+    private Invoice processDryRun_TARGET_DATE_Invoice(final UUID accountId, final LocalDate targetDate, final Set<LocalDate> allCandidateTargetDates, final BillingEventSet billingEvents, final List<Invoice> existingInvoices, final InternalCallContext context) throws InvoiceApiException {
 
-        LocalDate prevLocalDate = null;
-        for (final LocalDate cur : allCandidateTargetDates) {
-            if (cur.compareTo(targetDate) < 0) {
-                prevLocalDate = cur;
-            } else {
+
+        // Since allCandidateTargetDates is a SortedSet, the same comparator will be reused for the priorityQueue
```
pierre: Is there a specific reason to use a `PriorityQueue` over a `TreeSet` directly (the implementation of `getUpcomingInvoiceCandidateDates` already returns a `TreeSet`, we could change the returned type)?

I'm afraid that if we refactor the `getUpcomingInvoiceCandidateDates` code in the future, `allCandidateTargetDates` could become a generic `Collection` and this assumption would be gone (also, the magic is technically jdk specific).

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -448,28 +452,36 @@ private Invoice processDryRun_UPCOMING_INVOICE_FILTERING_Invoice(final UUID acco
...
+        // The list is an ordered list of items merged from existing notifications and upcoming notifications, each of these the result of a previous invoice being generated.
+        final List<Invoice> augmentedExistingInvoices = new ArrayList<Invoice>(existingInvoices);
+        Invoice additionalInvoice = null;
+        LocalDate cur;
+        while ((cur = pq.poll()) != null) {
+            if (cur.compareTo(targetDate) >= 0) {
                 break;
             }
+            // Loop through each boundary date prior to our given targetDate
+            final InvoiceWithFutureNotifications result = processAccountWithLockAndInputTargetDate(accountId, cur, billingEvents, augmentedExistingInvoices, true, false, context);
```
pierre: Just thinking out loud:

* From a performance perspective, it looks like it's actually not that bad since we don't go back to the DB anymore at this stage (we have all billing events and existing invoices). 👌 
* It doesn't look like this would trigger safety bounds, since stats are kept on a per `generateInvoice` basis, not across calls. 👌

pierre:
@sbrossie :shipit: Just one implementation change suggestion.

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
sbrossie: After re-reading code, simply removed the comment. There is no need for a special Comparator as the `PriorityQueue` (and the `TreeSet`) will both use the *natural order of the element* : `LocalDate` implements `Comparable` already.

PR #1050, opened 2018-10-16
tests: refactoring to improve stability

pierre:
* Revisit clock wiring
* Revisit properties configuration
* Revisit misconfigured Guice modules

Note: let's wait for green CI before merging.


pierre:
@sbrossie :white_check_mark: 😅

I've added a few more commits since your review, but I don't think there is anything controversial, so will merge as-is.

PR #1056, opened 2018-10-22
cache: Redis integration

pierre:
All tests using H2 and Redis (for both the clock and caching) pass.

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/glue/RedisShiroManager.java
+++ b/util/src/main/java/org/killbill/billing/util/glue/RedisShiroManager.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2014-2018 Groupon, Inc
+ * Copyright 2014-2018 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.killbill.billing.util.glue;
+
+import java.util.Collection;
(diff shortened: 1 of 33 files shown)


util/src/main/java/org/killbill/billing/util/glue/RedisShiroManager.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.Set;
+
+import javax.cache.Cache.Entry;
+
+import org.apache.shiro.cache.Cache;
+import org.apache.shiro.cache.CacheException;
+import org.apache.shiro.cache.CacheManager;
```
sbrossie: I am not sure i follow exactly what is happening. I would have expected to import `javax.cache.CacheManager` not `org.apache.shiro.cache.CacheManager`, but i am guessing `RedisShiroManager` manages the Shiro cache using redis as an implementation?

util/src/main/java/org/killbill/billing/util/glue/RedisShiroManager.java
```diff
@@ -0,0 +1,145 @@
...
+        @Override
+        public V get(final K key) throws CacheException {
+            return cache.get(key);
+        }
+
+        @Override
+        public V put(final K key, final V value) throws CacheException {
+            V previousValue;
+            while (true) {
+                previousValue = cache.get(key);
```
sbrossie: Confused too

util/src/main/java/org/killbill/billing/util/glue/RedisShiroManager.java
pierre: The `RedisShiroManager` indeed implements a `Shiro` `org.apache.shiro.cache.CacheManager`. `Shiro` doesn't support `javax.cache.CacheManager` and has its own thing instead, so this does the bridge between the two cache managers.

The `javax.cache.CacheManager` implementation for Redis is done via `Redis107CacheManagerProvider` (it's a `org.redisson.jcache.JCacheManager`).

util/src/main/java/org/killbill/billing/util/glue/RedisShiroManager.java
pierre: Which part specifically?

pierre:
Merging to make subsequent PR easier to read, but happy to continue the conversation above.

PR #1057, opened 2018-10-23
security: Shiro changes

pierre:
1. Only create sessions via the /1.0/kb/security/permissions path
2. Make sure old sessions are purged from the database
3. Authenticate plugin requests if credentials are provided 

Regarding 1.: any other API call with an Authorization header won't create sessions anymore: this will prevent useless state to be created with our official API clients, which don't use the `JSESSIONID` mechanism, and leave the Kaui integration unchanged.

Custom implementations may need to be updated.

```
# Generic API call with Authorization header
~ > curl -v \
         -uadmin:password \
         -H'X-Killbill-ApiKey: bob' \
         -H'X-Killbill-ApiSecret: lazar' \
         http://127.0.0.1:8080/1.0/kb/accounts/8313b0fd-99e2-4d35-bde3-035550f83293

...
< Set-Cookie: rememberMe=deleteMe; Path=/; Max-Age=0; Expires=Mon, 22-Oct-2018 08:24:48 GMT

# Get permissions API call with Authorization header
~ > curl -v \
         -uadmin:password \
         -H'X-Killbill-ApiKey: bob' \
         -H'X-Killbill-ApiSecret: lazar' \
         http://127.0.0.1:8080/1.0/kb/security/permissions

...
< Set-Cookie: JSESSIONID=406e6b1c-b1f9-4dd9-a4bd-5b4cae1184b0; Path=/; HttpOnly
< Set-Cookie: rememberMe=deleteMe; Path=/; Max-Age=0; Expires=Mon, 22-Oct-2018 08:25:13 GMT

# Generic API call with JSESSIONID cookie
curl -v \
         -H'Cookie: JSESSIONID=406e6b1c-b1f9-4dd9-a4bd-5b4cae1184b0' \
         -H'X-Killbill-ApiKey: bob' \
         -H'X-Killbill-ApiSecret: lazar' \
         http://127.0.0.1:8080/1.0/kb/accounts/8313b0fd-99e2-4d35-bde3-035550f83293

// No Set-Cookie header returned
```

Other behaviors regarding the `JSESSIONID` cookie haven't changed. Specifically:

* the session `last_access_time` is updated each time the session is used (i.e. each API call)
* if a timed out session is re-used (`last_access_time` older than 60 minutes by default), a `HTTP/1.1 401 Unauthorized` is returned with `Set-Cookie: JSESSIONID=deleteMe`, and the session is deleted from the database (and cache) on the server side
* if an invalid session is used (i.e. not found), a `HTTP/1.1 401 Unauthorized` is returned with no `Set-Cookie` header

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/glue/RedisShiroManagerProvider.java
+++ b/util/src/main/java/org/killbill/billing/util/glue/RedisShiroManagerProvider.java
@@ -23,6 +23,7 @@ import javax.inject.Provider;
 
 import org.apache.shiro.mgt.DefaultSecurityManager;
 import org.apache.shiro.mgt.SecurityManager;
+import org.apache.shiro.mgt.SubjectDAO;
 import org.redisson.api.RedissonClient;
 
 import com.codahale.metrics.MetricRegistry;
@@ -32,17 +33,20 @@ import static org.killbill.billing.util.glue.CacheModule.REDIS_CACHE_CLIENT;
 
 public class RedisShiroManagerProvider implements Provider<RedisShiroManager> {
 
-    private final CacheManager eh107CacheManager;
     private final SecurityManager securityManager;
+    private final SubjectDAO subjectDAO;
+    private final CacheManager eh107CacheManager;
     private final MetricRegistry metricRegistry;
     private final RedissonClient redissonClient;
 
     @Inject
     public RedisShiroManagerProvider(final SecurityManager securityManager,
+                                     final SubjectDAO subjectDAO,
                                      final CacheManager eh107CacheManager,
                                      final MetricRegistry metricRegistry,
                                      @Named(REDIS_CACHE_CLIENT) final RedissonClient redissonClient) {
         this.securityManager = securityManager;
+        this.subjectDAO = subjectDAO;
         this.eh107CacheManager = eh107CacheManager;
         this.metricRegistry = metricRegistry;
         this.redissonClient = redissonClient;
@@ -57,7 +61,7 @@ public class RedisShiroManagerProvider implements Provider<RedisShiroManager> {
             // For RBAC only (see also KillbillJdbcTenantRealmProvider)
             final DefaultSecurityManager securityManager = (DefaultSecurityManager) this.securityManager;
             securityManager.setCacheManager(shiroRedisManager);
-            securityManager.setSubjectDAO(new KillBillSubjectDAO());
+            securityManager.setSubjectDAO(subjectDAO);
         }
 
         return shiroRedisManager;
--- a/util/src/main/resources/org/killbill/billing/util/security/shiro/dao/JDBCSessionSqlDao.sql.stg
+++ b/util/src/main/resources/org/killbill/billing/util/security/shiro/dao/JDBCSessionSqlDao.sql.stg
@@ -46,3 +46,18 @@ delete from sessions
 where id = :id
 ;
 >>
+
+findOldActiveSessions() ::= <<
+select
+  record_id
+, id
+, start_timestamp
+, last_access_time
+, timeout
+, host
+, session_data
+from sessions
+order by record_id asc
+limit 100
+;
+>>
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>killbill-oss-parent</artifactId>
         <groupId>org.kill-bill.billing</groupId>
-        <version>0.142.6</version>
+        <version>0.142.7</version>
     </parent>
     <artifactId>killbill</artifactId>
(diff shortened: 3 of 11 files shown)


util/src/main/resources/org/killbill/billing/util/security/shiro/dao/JDBCSessionSqlDao.sql.stg
```diff
@@ -45,4 +45,19 @@ delete() ::= <<
...
+findOldActiveSessions() ::= <<
+select
+  record_id
+, id
+, start_timestamp
+, last_access_time
+, timeout
+, host
+, session_data
+from sessions
```
sbrossie: I was expecting to see a where close that filters the *old* sessions ?

util/src/main/resources/org/killbill/billing/util/security/shiro/dao/JDBCSessionSqlDao.sql.stg
```diff
@@ -45,4 +45,19 @@ delete() ::= <<
...
+order by record_id asc
+limit 100
```
sbrossie: What if there is more than 100?

util/src/main/java/org/killbill/billing/util/glue/RedisShiroManagerProvider.java
```diff
@@ -32,17 +33,20 @@
...
-    private final CacheManager eh107CacheManager;
     private final SecurityManager securityManager;
+    private final SubjectDAO subjectDAO;
+    private final CacheManager eh107CacheManager;
     private final MetricRegistry metricRegistry;
     private final RedissonClient redissonClient;
 
     @Inject
     public RedisShiroManagerProvider(final SecurityManager securityManager,
+                                     final SubjectDAO subjectDAO,
```
sbrossie: Do we support storing our sessions into Redis -- instead of database?

sbrossie:
@Pierre : question on following comment:

> if a timed out session is re-used (last_access_time older than 60 minutes by default), a HTTP/1.1 401 Unauthorized is returned with Set-Cookie: JSESSIONID=deleteMe, and the session is deleted from the database (and cache) on the server side

So in that case, the client needs to re-issue the `/1.0/kb/security/permissions` call to get a new sessionID - is this how KAUI already works?

util/src/main/resources/org/killbill/billing/util/security/shiro/dao/JDBCSessionSqlDao.sql.stg
pierre: The check for `old` (orphaned) is done in the code. The `sessions` table should almost always be empty, except for currently active sessions (Kaui users).

sbrossie: Nit: In that case renaming that query `findOldActiveSessions` -> `findActiveSessions` would make sense

util/src/main/resources/org/killbill/billing/util/security/shiro/dao/JDBCSessionSqlDao.sql.stg
pierre: It's just an arbitrary limit for when a very Glarge deployment with billions of rows decide to upgrade to this version (without it -- OOM) 😬The check is run every hour so the idea is that the table would be cleaned up 100 rows at a time.

There is one potential issue which is if 100 users created a session before more recent users **and** these 100 users are still active (while the recent users aren't), then the recent sessions wouldn't be cleaned-up. This is quite a corner case though. The real fix would be to `order by last_access_time asc` but I want to avoid that for the scenario described above (it would trigger a full table scan -- this could be mitigated if the team truncates the table before upgrading but if they don't read carefully the release notes...).

util/src/main/java/org/killbill/billing/util/glue/RedisShiroManagerProvider.java
pierre: Yup 😎

pierre:
> So in that case, the client needs to re-issue the `/1.0/kb/security/permissions` call to get a new sessionID - is this how KAUI already works?

Yup, this is what happens when your Kaui session times out and you are redirect to the login screen.

util/src/main/resources/org/killbill/billing/util/security/shiro/dao/JDBCSessionSqlDao.sql.stg
pierre: Not quite, it's really `grabASubsetOfTheSessionsOrderedByCreatedDateAndCheckIfTheyAreOrphaned`... Naming is hard 😖

PR #1059, opened 2018-10-26
Fixes issue with the endDate mapping for tax items

mrmstn:
The `endDate` is a possible propertie for a tax-item, since the
following methode from killbill-base-plugin makes use of it:
`org.killbill.billing.plugin.api.invoice.PluginInvoiceItem#createTaxItem`
https://github.com/killbill/killbill-plugin-framework-java/blob/master/src/main/java/org/killbill/billing/plugin/api/invoice/PluginInvoiceItem.java#L58

This PR adds the mapping entry needed to map the endDate from the InvoiceItemDao back to the TaxItem item

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithTaxItems.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithTaxItems.java
@@ -251,6 +251,102 @@ public class TestWithTaxItems extends TestIntegrationBase {
     }
 
     @Test(groups = "slow")
+    public void testTaxRateStartAndEndDate() throws Exception {
+        // Set clock to the initial start date - we implicitly assume here that the account timezone is UTC
+        clock.setDay(new LocalDate(2017, 11, 15));
+
+        final AccountData accountData = getAccountData(null);
+        final Account account = createAccountWithNonOsgiPaymentMethod(accountData);
+        accountChecker.checkAccount(account.getId(), accountData, callContext);
+
+        // Create original subscription (Trial PHASE) -> $0 invoice.
+        final DefaultEntitlement bpSubscription = createBaseEntitlementAndCheckForCompletion(
+                account.getId(),
+                "bundleKey",
+                "Pistol",
+                ProductCategory.BASE,
+                BillingPeriod.MONTHLY,
+                NextEvent.CREATE,
+                NextEvent.BLOCK,
+                NextEvent.INVOICE
+                                                                                            );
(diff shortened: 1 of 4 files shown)


pierre:
How do you plan on using `startDate` and `endDate` for tax items?

It looks like you are developing your own tax plugin? It would be easier for us if you would open a PR with all the changes you need once your plugin is working, so we have an overall better understanding of your integration.

mrmstn:
In Europe, it's possible that the VAT-rates change over the year (and also taxes in generall). For example, Switzerland had a change from 8% to 7,7% in 2017 to 2018. 

Let's say, Alice has a subscription for a car, billed every month. In december, Alice will get the invoice for the periode 2017-12-15 until 2018-01-15.
The Billingsystem should now (imho) add 2 tax items, 1 from 2017-12-15 until 2017-12-31 with a rate of 8% and another one from 2018-01-01 till the end of the periode on a rate of 7,7%.

I fully understand your point in waiting for all my changes,  but I really think this is a bug in the current system, since the methode `createTaxItem` in the `PluginInvoiceItem` class from the killbill-plugin-framework allows you to add the endDate to the database - the only issue for now is, you won't get the endDate back again from the database.  

https://github.com/killbill/killbill-plugin-framework-java/blob/master/src/main/java/org/killbill/billing/plugin/api/invoice/PluginInvoiceItem.java#L58

pierre:
> In Europe, it's possible that the VAT-rates change over the year (and also taxes in generall). For example, Switzerland had a change from 8% to 7,7% in 2017 to 2018.

Yup. I think it's expected globally, including both at the federal and state level in the US.

> Let's say, Alice has a subscription for a car, billed every month. In december, Alice will get the invoice for the periode 2017-12-15 until 2018-01-15.
> The Billingsystem should now (imho) add 2 tax items, 1 from 2017-12-15 until 2017-12-31 with a rate of 8% and another one from 2018-01-01 till the end of the periode on a rate of 7,7%.

This would actually depend on your use-case. In some scenarii, you would only use 8%, in others only 7.7% or in some others even pro-rate as you suggested (depends if it's goods or services, when the supplies were shipped, when the services were rendered, etc.). This is why tax calculation is delegated to plugins so you can implement your own logic.

> I fully understand your point in waiting for all my changes, but I really think this is a bug in the current system

Sure, it could well be a bug. I'd still like to see all of the changes you need for your plugin in one PR though, including at least some basic unit tests, so we can catch this in the future.

mrmstn:
You are right, I should also add some tests to prove my point. I'll look into that next week

mrmstn:
I added some tests, feel free to merge

PR #1066, opened 2018-11-29
Fixes for 0.20.5

pierre:
* Relax `DefaultPlan` validation
* Fix usage invoicing after BCD change

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithBCDUpdate.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithBCDUpdate.java
@@ -823,9 +823,9 @@ public class TestWithBCDUpdate extends TestIntegrationBase {
                                     new ExpectedInvoiceItemCheck(new LocalDate(2012, 5, 1), new LocalDate(2013, 5, 1), InvoiceItemType.RECURRING, new BigDecimal("2399.95")),
                                     new ExpectedInvoiceItemCheck(new LocalDate(2012, 4, 1), new LocalDate(2012, 5, 1), InvoiceItemType.USAGE, new BigDecimal("5.90")));
 
-        final DateTime bpExpectedCTD = account.getReferenceTime().withYear(2013).withMonthOfYear(5).withDayOfMonth(1);
+        final DateTime bpExpectedCTD = new DateTime("2013-05-01T00:00:00.000Z");
         assertEquals(subscriptionBaseInternalApiApi.getSubscriptionFromId(bpSubscription.getId(), internalCallContext).getChargedThroughDate().compareTo(bpExpectedCTD), 0);
-        DateTime aoExpectedCTD = account.getReferenceTime().withMonthOfYear(5).withDayOfMonth(1);
+        DateTime aoExpectedCTD = new DateTime("2012-05-01T00:00:00.000Z");
         assertEquals(subscriptionBaseInternalApiApi.getSubscriptionFromId(aoSubscription.getId(), internalCallContext).getChargedThroughDate().compareTo(aoExpectedCTD), 0);
 
         // 2012-05-05
@@ -852,7 +852,7 @@ public class TestWithBCDUpdate extends TestIntegrationBase {
         invoiceChecker.checkInvoice(account.getId(), 4, callContext,
                                     new ExpectedInvoiceItemCheck(new LocalDate(2012, 5, 5), new LocalDate(2012, 6, 5), InvoiceItemType.USAGE, new BigDecimal("5.90")));
 
-        aoExpectedCTD = account.getReferenceTime().withMonthOfYear(6).withDayOfMonth(5);
+        aoExpectedCTD = new DateTime("2012-06-05T00:00:00.000Z");
         assertEquals(subscriptionBaseInternalApiApi.getSubscriptionFromId(aoSubscription.getId(), internalCallContext).getChargedThroughDate().compareTo(aoExpectedCTD), 0);
     }
 }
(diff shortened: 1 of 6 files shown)


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithBCDUpdate.java
```diff
@@ -793,4 +787,72 @@ public void testWithBCDOnOperations() throws Exception {
...
+        clock.addMonths(1);
+        assertListenerStatus();
+
+        invoiceChecker.checkInvoice(account.getId(), 2, callContext,
+                                    new ExpectedInvoiceItemCheck(new LocalDate(2012, 5, 1), new LocalDate(2013, 5, 1), InvoiceItemType.RECURRING, new BigDecimal("2399.95")),
+                                    new ExpectedInvoiceItemCheck(new LocalDate(2012, 4, 1), new LocalDate(2012, 5, 1), InvoiceItemType.USAGE, new BigDecimal("5.90")));
+
+        final DateTime bpExpectedCTD = account.getReferenceTime().withYear(2013).withMonthOfYear(5).withDayOfMonth(1);
+        assertEquals(subscriptionBaseInternalApiApi.getSubscriptionFromId(bpSubscription.getId(), internalCallContext).getChargedThroughDate().compareTo(bpExpectedCTD), 0);
+        DateTime aoExpectedCTD = account.getReferenceTime().withMonthOfYear(5).withDayOfMonth(1);
```
sbrossie: Don't you need the `withYear ()` ?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithBCDUpdate.java
```diff
@@ -793,4 +787,72 @@ public void testWithBCDOnOperations() throws Exception {
...
+        // 2012-05-01
+        busHandler.pushExpectedEvents(NextEvent.PHASE, NextEvent.NULL_INVOICE, NextEvent.INVOICE, NextEvent.PAYMENT, NextEvent.INVOICE_PAYMENT);
+        clock.addMonths(1);
+        assertListenerStatus();
+
+        invoiceChecker.checkInvoice(account.getId(), 2, callContext,
+                                    new ExpectedInvoiceItemCheck(new LocalDate(2012, 5, 1), new LocalDate(2013, 5, 1), InvoiceItemType.RECURRING, new BigDecimal("2399.95")),
+                                    new ExpectedInvoiceItemCheck(new LocalDate(2012, 4, 1), new LocalDate(2012, 5, 1), InvoiceItemType.USAGE, new BigDecimal("5.90")));
+
+        final DateTime bpExpectedCTD = account.getReferenceTime().withYear(2013).withMonthOfYear(5).withDayOfMonth(1);
```
sbrossie: What is the goal behind extracting date from `account.getReferenceTime()` and overriding values - wouldn't it be clearer to just hardcode what we thing the expected date should be?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithBCDUpdate.java
```diff
@@ -793,4 +787,72 @@ public void testWithBCDOnOperations() throws Exception {
 
 
     }
+
+    @Test(groups = "slow")
+    public void testBCDChangeForConsumableInArrearPlan() throws Exception {
+        // We take april as it has 30 days (easier to play with BCD)
+        // Set clock to the initial start date - we implicitly assume here that the account timezone is UTC
+        clock.setDay(new LocalDate(2012, 4, 1));
```
sbrossie: Nit: Potentially we could update the start date - e.g `2016` should behave like `2012`

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -413,10 +428,6 @@ public Usage getUsage() {
         return usage;
     }
 
-    public int getBCD() {
-        return billingEvents.get(0).getBillCycleDayLocal();
```
sbrossie: 😨

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -129,23 +129,21 @@ public ContiguousIntervalUsageInArrear build(final boolean closedInterval) {
...
-                }
+
+        for (int i = 0; i < billingEvents.size(); i++) {
+            final BillingEvent billingEvent = billingEvents.get(i);
+            if (i == billingEvents.size() - 1) {
+                addTransitionTimesForBillingEvent(startDate, endDate, billingEvent.getBillCycleDayLocal());
+            } else {
+                final BillingEvent nextBillingEvent = billingEvents.get(i + 1);
+                final LocalDate nextEndDate = internalTenantContext.toLocalDate(nextBillingEvent.getEffectiveDate());
+                addTransitionTimesForBillingEvent(startDate, nextEndDate, billingEvent.getBillCycleDayLocal());
```
sbrossie: I am confused why we reuse the same `startDate` on each iteration though: Shouldn't the `addTransitionTimesForBillingEvent` return the new `startDate` for the next iteration? Are we not gonna end up with duplicate transitions -- or even worst incorrect transitions as the BCD could now change?

pierre: I actually went back and forth on that one...

The way we use `BillingIntervalDetail` here is a bit odd as we only want to access `getFutureBillingDateFor(numberOfPeriod)`. `startDate` in `BillingIntervalDetail` is only used to compute the `firstBillingCycleDate`, but because we will always call `alignProposedBillCycleDate`, I don't think the value matters. The code could probably benefit from some refactoring but I wanted to minimize the changes.

That being said, I've changed it in cde63173c83f82dd10dd323536241857c31a5dc2 and added a unit test with two BCD changes.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithBCDUpdate.java
pierre: Fixed in c483cbc44d3a19616ac5a6de840a35bd507f40cd.

Issue #865, opened 2018-02-13
Wrong BillingEvent BCD calculation in some codepaths


pierre:
Let's assume the following scenario:

* Create account with no BCD set
* Bulk create a bundle with two subscriptions:
  * `BASE`: Monthly, with 30 days trial
  * `ADD_ON`: Triannual (i.e. every 4 months), with no trial

_The add-on could represent a non refundable setup fee for instance_

Both plans are `ACCOUNT` aligned. Subscription creation happens on `2013-08-07`, the `PHASE` event for the `BASE` subscription is on `2013-09-06`.

The expectation is that the account BCD is set to the 7 (date of first recurring non-$0 charge).

What we see in practice:

* BCD is set to the 6, i.e. the date of the first recurring non-$0 charge for the `BASE` plan
* Before `BCD` is computed and set:
  * `Entitlement` BCD for the `BASE`: 6
  * `Entitlement` BCD for the `ADD_ON`: 7
* Before `BCD` is computed and set:
  * `BillingEvents` (`CREATE` and `PHASE`) BCD for the `BASE`: 6
  * `BillingEvent` (`CREATE`) BCD for the `ADD_ON`: 7
* After `BCD` is computed and set:
  * `Entitlement` BCD for the `BASE`: 6
  * `Entitlement` BCD for the `ADD_ON`: 6
* After `BCD` is computed and set:
  * `BillingEvents` (`CREATE` and `PHASE`) BCD for the `BASE`: 6
  * `BillingEvent` (`CREATE`) BCD for the `ADD_ON`: 6

One issue is that `currentAccountBCD` isn't updated [here](https://github.com/killbill/killbill/blob/8006340ebd8fe9bff1ffdd340a4cf886839c8e90/junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java#L214-L218), but the problem goes deeper in `EventsStreamBuilder` which doesn't the run the BCD update logic (so `BillCycleDayCalculator#calculateBcdForAlignment` logic is wrong).

_Found while debugging https://github.com/killbill/killbill/issues/282._

PR #1067, opened 2018-11-30
Fix for 865

pierre:
Before, we were using the first calculated BCD of the first `BillingEvent` seen. This was a bit arbitrary, especially across bundles (since we weren't looking at the date).

~We now look at the first `BillingEvent` with a non-$0 recurring price and use its effective date to get the account BCD. This matches better our BCD definition: _date of first recurring non-$0 charge_.~

~One side effect is that for pure usage based plans, the account BCD isn't set anymore. For instance, a consumable in arrear usage plan billed on the 1st would have previously set the account BCD to the 1st. Now, it is not updated (unless a recurring price is set to the plan).~

_Edit: see discussion below_

See https://github.com/killbill/killbill/issues/865 and related Ruby test https://github.com/killbill/killbill-integration-tests/pull/35.

This also fixes a catalog caching race condition seen when running integration tests: if the
Analytics plugin is triggered before `updateTenantKeyValue` completes for instance, the wrong catalog could be cached.


junction/src/test/java/org/killbill/billing/junction/plumbing/billing/TestDefaultInternalBillingApi.java
```diff
@@ -50,6 +52,68 @@
...
+                                                        ImmutableList.<PluginProperty>of(),
+                                                        callContext);
+        assertListenerStatus();
+
+        final List<Entitlement> entitlements = entitlementApi.getAllEntitlementsForAccountId(account.getId(), callContext);
+        Assert.assertEquals(entitlements.size(), 2);
+        Assert.assertEquals(entitlements.get(0).getLastActiveProduct().getName(), "Shotgun");
+        Assert.assertEquals(entitlements.get(0).getBillCycleDayLocal(), (Integer) 6);
+        Assert.assertEquals(entitlements.get(1).getLastActiveProduct().getName(), "Cabinet");
+        Assert.assertEquals(entitlements.get(1).getBillCycleDayLocal(), (Integer) 7);
```
sbrossie: This is very weird though... our semantics of what is `getBillCycleDayLocal ()` is so blury. Basically, depending when you look at it, this is different... I am wondering if in case of `ACCOUNT` alignement, returning `null` until the account alignement was computed would not be a better option. As an alternative, forcing this computation, could solve the issue.

pierre:
> Although, even for pure usage based, there is an underlying recurrence -- cadence at which we invoice this usage section, so i am wondering if this is not something we should fix as well.

Fixed via 91a6fd4c42d4b4e66183bc0577981609971e1d3f. I've also added two new tests, one with a recurring in arrear plan, and one with a pure usage in arrear plan.

See https://github.com/killbill/killbill-integration-tests/pull/35/files for Ruby tests changes.

junction/src/test/java/org/killbill/billing/junction/plumbing/billing/TestDefaultInternalBillingApi.java
pierre: Fixed via aec22cc914db60632ea12a49317b00cead221454.

pierre:
> I am wondering if we should not keep an ordered list of such `accountBCDCandidate` and get the first non null one ?

Actually, re-reading the code, `accountBCDCandidate` is never `null`/`0`. I'll check but it looks like I should be able to remove the `if (accountBCDCandidate != 0)` check.

> In addition -- our last conversation on Slack-- we would exclude from that heap/ordered list any billing event associated with non ACCOUNT aligned subscription.

Fixed by 14222cb3a6e23d0cb19ea9d06eff07979fc1b49f.

junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java
```diff
@@ -178,27 +179,63 @@ private void addBillingEventsForBundles(final List<SubscriptionBaseBundle> bundl
...
+            }
+
+            // BCD in the account timezone
+            final int accountBCDCandidate = oldestAccountAlignedBillingEvent.getBillCycleDayLocal();
+            if (accountBCDCandidate != 0) {
+                log.info("Setting account BCD='{}', accountId='{}'", accountBCDCandidate, account.getId());
+                accountApi.updateBCD(account.getExternalKey(), accountBCDCandidate, context);
+
+                // Because we now have computed the real BCD, we need to re-compute the BillingEvents BCD for ACCOUNT alignments (see BillCycleDayCalculator#calculateBcdForAlignment).
+                // The code could maybe be optimized (no need to re-run the full function?), but since it's run once per account, it's probably not worth it.
```
sbrossie: I think this is at least worth a discussion (not strong feeling though): Less about the perf, but more about the maintenance of the code -- this is related to the comment below in  `calculateBcdForTransition`. One thought, is that we could detect first if `currentAccountBCD` is 0 and extract its value from `ACCOUNT` aligned subscriptions, and then proceed with populating billing events without all this complexity.

junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java
```diff
@@ -220,23 +259,19 @@ private void addBillingEventsForSubscription(final ImmutableAccountData account,
...
             }
         }
     }
 
-    private int calculateBcdForTransition(final Catalog catalog, final Map<UUID, Integer> bcdCache, final SubscriptionBase baseSubscription, final SubscriptionBase subscription, final int accountBillCycleDayLocal, final EffectiveSubscriptionInternalEvent transition, final InternalTenantContext internalTenantContext)
-            throws CatalogApiException, AccountApiException, SubscriptionBaseApiException {
-        final BillingAlignment alignment = catalog.billingAlignment(getPlanPhaseSpecifierFromTransition(catalog, transition), transition.getEffectiveTransitionTime(), subscription.getStartDate());
+    private int calculateBcdForTransition(final BillingAlignment realBillingAlignment, final Map<UUID, Integer> bcdCache, final SubscriptionBase baseSubscription, final SubscriptionBase subscription, final int accountBillCycleDayLocal, final InternalTenantContext internalTenantContext) {
+        BillingAlignment alignment = realBillingAlignment;
+        if (alignment == BillingAlignment.ACCOUNT && accountBillCycleDayLocal == 0) {
```
sbrossie: Not super excited about `calculateBcdForTransition` being called twice, one time faking `ACCOUNT` aligned with `SUBSCRIPTION` aligned and then one time with the real thing -- see comment above.

pierre: This is actually unrelated to the recursive call.

Take a look at the change in `BillCycleDayCalculator.java`:

```
 public abstract class BillCycleDayCalculator {

     private static final Logger log = LoggerFactory.getLogger(BillCycleDayCalculator.class);
@@ -39,7 +41,8 @@ public abstract class BillCycleDayCalculator {
         int result = 0;
         switch (alignment) {
             case ACCOUNT:
-                result = accountBillCycleDayLocal != 0 ? accountBillCycleDayLocal : calculateOrRetrieveBcdFromSubscription(bcdCache, subscription, internalTenantContext);
+                Preconditions.checkState(accountBillCycleDayLocal != 0, "Account BCD should be set at this point");
+                result = accountBillCycleDayLocal;
                 break;
             case BUNDLE:
                 result = calculateOrRetrieveBcdFromSubscription(bcdCache, baseSubscription, internalTenantContext);
```

Previously, when the BCD wouldn't be set, `BillCycleDayCalculator.calculateBcdForAlignment` would call `calculateOrRetrieveBcdFromSubscription` for `ACCOUNT` alignments (this is how we compute the account BCD). I've simply moved that logic outside, so that it's explicit (and that codepath should only be exercised by junction now, while today it affects anyone calling `calculateBcdForAlignment` -- hence #865).

junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java
pierre: I agree, and this is what I tried to do originally, but it's harder in practice: we need all billing events, retrieve catalog objects and run the `BillCycleDayCalculator` logic to compute that BCD (we effectively need 2 pass, no matter how you look at it).

That being said, I understand the concern. I'll try to do another pass to see if I'm more inspired today 🤔

PR #1068, opened 2018-12-03
Add support to link usage tracking ids with generated invoices.

sbrossie:
There is a new table, `invoice_tracking_ids`,  that is populated by the invoice code when generating `USAGE` items, and whose goal is to track usage records with invoices.

Due to the nature of our usage record apis, where we can record multiple data points for multiple usage sections and dates in one api call, there is no guarantee  for a 1-1 mapping between  a given `tracking_id` and a given `invoice_id`.

Also, at this point, there is no api to retrieve data from this new `invoice_tracking_ids` table as this would require an api change -- from for `0.20.x` releases.

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
@@ -288,7 +288,7 @@ public class TestConsumableInArrear extends TestIntegrationBase {
 
         final List<UsageRecord> bp1StoneRecords2 = new ArrayList();
         bp1StoneRecords2.add(new UsageRecord(new LocalDate(2012, 4, 23), 10L));
-        // Outside of range for this period -> It's tracking ID spreads across 2 invoices
+        // Outside of range for this period -> Its tracking ID spreads across 2 invoices
         bp1StoneRecords2.add(new UsageRecord(new LocalDate(2012, 5, 1), 101L));
         final SubscriptionUsageRecord bp1UsageRecord2 = new SubscriptionUsageRecord(bp1.getId(), "bp1-tracking-2", ImmutableList.of(new UnitUsageRecord("stones", bp1StoneRecords2)));
         recordUsageData(bp1UsageRecord2, callContext);
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
@@ -18,6 +18,8 @@
 package org.killbill.billing.beatrix.integration.usage;
 
 import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
 
 import org.joda.time.DateTime;
 import org.joda.time.LocalDate;
@@ -30,12 +32,19 @@ import org.killbill.billing.catalog.api.BillingActionPolicy;
 import org.killbill.billing.catalog.api.BillingPeriod;
 import org.killbill.billing.catalog.api.ProductCategory;
 import org.killbill.billing.entitlement.api.DefaultEntitlement;
+import org.killbill.billing.invoice.api.Invoice;
 import org.killbill.billing.invoice.api.InvoiceItemType;
 import org.killbill.billing.payment.api.PluginProperty;
+import org.killbill.billing.usage.api.SubscriptionUsageRecord;
+import org.killbill.billing.usage.api.UnitUsageRecord;
+import org.killbill.billing.usage.api.UsageRecord;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+
+import static org.testng.Assert.assertEquals;
 
 public class TestConsumableInArrear extends TestIntegrationBase {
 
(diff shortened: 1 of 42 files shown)


beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
```diff
@@ -204,44 +219,113 @@ public void testWithNoRecurringPlan() throws Exception {
...
+        final List<UsageRecord> bp1StoneRecords1 = new ArrayList();
+        bp1StoneRecords1.add(new UsageRecord(new LocalDate(2012, 4, 5), 5L));
+        bp1StoneRecords1.add(new UsageRecord(new LocalDate(2012, 4, 15), 10L));
+        bp1StoneRecords1.add(new UsageRecord(new LocalDate(2012, 4, 16), 15L));
+        final SubscriptionUsageRecord bp1UsageRecord1 = new SubscriptionUsageRecord(bp1.getId(), "bp1-tracking-1", ImmutableList.of(new UnitUsageRecord("stones", bp1StoneRecords1)));
+        recordUsageData(bp1UsageRecord1, callContext);
+
+        final List<UsageRecord> bp1StoneRecords2 = new ArrayList();
+        bp1StoneRecords2.add(new UsageRecord(new LocalDate(2012, 4, 23), 10L));
+        // Outside of range for this period -> It's tracking ID spreads across 2 invoices
```
pierre: Nit: `Its` 🤓

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -204,7 +206,7 @@ public void processSubscriptionStartRequestedDate(final RequestedSubscriptionInt
 
             final ImmutableAccountData account = accountApi.getImmutableAccountDataById(accountId, context);
 
-            commitInvoiceAndSetFutureNotifications(account, null, notificationsBuilder.build(), context);
+            commitInvoiceAndSetFutureNotifications(account, null, ImmutableSet.of(), notificationsBuilder.build(), context);
```
pierre: Nit: it seems we could have two signatures:

```
void commitInvoiceAndSetFutureNotifications(final ImmutableAccountData account,
                                            final FutureAccountNotifications futureAccountNotifications,
                                            final InternalCallContext context);

void commitInvoiceAndSetFutureNotifications(final ImmutableAccountData account,
                                            final InvoiceModelDao invoiceModelDao,
                                            final Set<InvoiceTrackingModelDao> trackingIds,
                                            final FutureAccountNotifications futureAccountNotifications,
                                            final InternalCallContext context);
```

invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceTrackingModelDao.java
```diff
@@ -0,0 +1,145 @@
...
+import org.killbill.billing.util.entity.dao.EntityModelDao;
+import org.killbill.billing.util.entity.dao.EntityModelDaoBase;
+
+import com.google.common.base.Objects;
+
+public class InvoiceTrackingModelDao extends EntityModelDaoBase implements EntityModelDao<Entity> {
+
+    private String trackingId;
+    private UUID invoiceId;
+    private UUID subscriptionId;
```
pierre: The mapping `trackingId -> {subscriptionId,unitType,recordDate}` already exists in the usage module. Should we just keep the mapping `trackingId->invoiceId` here?

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceWithMetadata.java
```diff
@@ -80,6 +83,73 @@ public boolean apply(final InvoiceItem input) {
...
+            if (this == o) {
+                return true;
+            }
+            if (!(o instanceof TrackingRecordId)) {
+                return false;
+            }
+            final TrackingRecordId that = (TrackingRecordId) o;
+            // !!! Exclude invoiceId on purpose.
+            //
+            // The Set methods (Sets.difference) is used to exclude usage record already invoiced (on a specified invoiceId),
```
pierre: In the past, for such cases, we did not rely on the native methods but introduced our own (see `InvoiceItem#matches` for instance). In other words, should we keep `equals` as-is and implement our custom `Sets.difference` utility function calling a different method?

I'm just worried about future automatic refactoring. Also, it could introduce other weird behaviors when we implicitly truly expect a well-behaved implementation (for instance, doesn't `HashSet` need a real hash function for the `add` operation?).

beatrix/src/test/java/org/killbill/billing/beatrix/util/InvoiceChecker.java
```diff
@@ -170,6 +180,26 @@ public void checkChargedThroughDate(final UUID entitlementId, final LocalDate ex
         }
     }
 
+
+    public void checkTrackingIds(final Invoice invoice, final Set<String> expectedTrackingIds, final InternalCallContext internalCallContext) {
+        final InvoiceTrackingSqlDao dao = dbi.onDemand(InvoiceTrackingSqlDao.class);
```
pierre: Might be worth creating a bug already for 0.22 to introduce a real API and fix tests.

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceWithMetadata.java
```diff
@@ -80,6 +83,73 @@ public boolean apply(final InvoiceItem input) {
         });
     }
 
+    public Set<TrackingRecordId> getTrackingIds() {
```
pierre: I'm guessing that in `0.22`, we will want to expose these in the `Invoice` object. In the meantime, should we at least pass the information to invoice plugins as a custom/hidden plugin property? That way, it could help folks access that information without having to query the table directly.

Since you have it in `InvoiceWithMetadata`, it seems pretty straightforward to send it to `onSuccessCall` and `onFailureCall`. It's a bit more difficult for the `priorCall` but we could maybe punt that for now.

PR #1087, opened 2019-01-22
Fix for 1086

sbrossie:


invoice/src/main/resources/org/killbill/billing/invoice/migration/V20190121141325__tracking_ids_is_active.sql
```diff
@@ -0,0 +1,4 @@
+alter table invoice_tracking_ids add column is_active boolean default true after record_date;
+alter table invoice_tracking_ids add column updated_by varchar(50) NOT NULL after created_date;
+alter table invoice_tracking_ids add column updated_date datetime NOT NULL DEFAULT '1970-01-01 00:00:00' after updated_by;
+create index invoice_tracking_invoice_id_idx on invoice_tracking_ids(tenant_record_id, invoice_id);
```
pierre: Shouldn't `invoice_id` come first (bigger cardinality)?

invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceTrackingModelDao.java
```diff
@@ -88,6 +90,19 @@ public void setUnitType(final String unitType) {
         this.unitType = unitType;
     }
 
+    // TODO required for jdbi binder
+    public boolean getIsActive() {
+        return isActive;
```
pierre: Nit: could call `isActive()` instead.

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceTrackingSqlDao.sql.stg
```diff
@@ -8,8 +8,11 @@ tableFields(prefix) ::= <<
 , <prefix>subscription_id
 , <prefix>unit_type
 , <prefix>record_date
+, <prefix> is_active
```
pierre: Nit: extra space.

invoice/src/main/resources/org/killbill/billing/invoice/migration/V20190121141325__tracking_ids_is_active.sql
sbrossie: This is not really clear. From this [post](http://colleenmorrow.com/2011/04/07/composite-indexes-does-column-order-matter/):
```
I agree that on composite indexes, the choice of ordering the most selective first doesn’t normally impact performance. This is contrary to what is often given as advice....
```

Besides perf, the other aspect in the index ordering is whether we want to use this index for queries using only one column, in which case, this column needs to be first in the index.

Stepping back, since our `invoiceId` are unique across tenants, i could simplify and only have an index on the `invoice_id`  -- we already have a composite index `tenant_record_id, account_record_id` allowing to search per tenant if required. Thoughts?

Issue #1084, opened 2019-01-15
Unexpected NullPointerException in UsageInvoiceItemGenerator


pierre:
This can happen in the follow scenario:

1. Generate future invoice (assuming subscription has associated usage data)
2. Change plan (different `usageName`)
3. Regenerate invoice

Test scenario in branch [fix-for-1084](https://github.com/killbill/killbill/commit/1da394f2ac302bb88d57a3d5d1e852a9992a8955).

```
2019-01-14T15:15:10,395+0000 lvl='WARN', log='RuntimeExceptionMapper', th='catalina-exec-7', xff='', rId='da4b3566-585a-4c63-b825-873096296815', tok='da4b3566-585a-4c63-b825-873096296815', aRId='595', tRId='1', Unexpected NullPointerException
java.lang.NullPointerException: null
	at org.killbill.billing.invoice.generator.UsageInvoiceItemGenerator$2$1.apply(UsageInvoiceItemGenerator.java:199)
	at org.killbill.billing.invoice.generator.UsageInvoiceItemGenerator$2$1.apply(UsageInvoiceItemGenerator.java:194)
	at com.google.common.collect.Iterators$5.computeNext(Iterators.java:639)
	at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:141)
	at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:136)
	at com.google.common.collect.Iterators$ConcatenatedIterator.hasNext(Iterators.java:1323)
	at org.killbill.billing.invoice.generator.UsageInvoiceItemGenerator.extractPerSubscriptionExistingInArrearUsageItems(UsageInvoiceItemGenerator.java:207)
	at org.killbill.billing.invoice.generator.UsageInvoiceItemGenerator.generateItems(UsageInvoiceItemGenerator.java:87)
	at org.killbill.billing.invoice.generator.DefaultInvoiceGenerator.generateInvoice(DefaultInvoiceGenerator.java:98)
	at org.killbill.billing.invoice.InvoiceDispatcher.generateKillBillInvoice(InvoiceDispatcher.java:667)
	at org.killbill.billing.invoice.InvoiceDispatcher.processAccountWithLockAndInputTargetDate(InvoiceDispatcher.java:556)
	at org.killbill.billing.invoice.InvoiceDispatcher.processAccountWithLock(InvoiceDispatcher.java:353)
[...]

PR #1088, opened 2019-01-23
Fix for 1084

pierre:
This fixes https://github.com/killbill/killbill/issues/1084.


invoice/src/main/java/org/killbill/billing/invoice/generator/UsageInvoiceItemGenerator.java
```diff
@@ -196,7 +196,7 @@ private void updatePerSubscriptionNextNotificationUsageDate(final UUID subscript
                     public boolean apply(final InvoiceItem input) {
                         if (input.getInvoiceItemType() == InvoiceItemType.USAGE) {
                             final Usage usage = knownUsage.get(input.getUsageName());
-                            return usage.getBillingMode() == BillingMode.IN_ARREAR;
+                            return usage != null && usage.getBillingMode() == BillingMode.IN_ARREAR;
```
sbrossie: nit: Fix makes sense and is simple enough - just wondering if a log line would make sense, since this is a case where we previously invoiced some usage, which is not not referenced in the current set of billing events.

pierre: I'm always very wary of adding logging in `invoice`, because if the condition is met, it will always be met for each invoice run in the future. This could be quite noisy for these occasions where the user is actually ok with it (we have the _problem_ today in some Janitor GET paths in `payment`).

PR #1089, opened 2019-01-24
Fix for 1078

sbrossie:
See [usage plugin api proposal](https://github.com/killbill/killbill-plugin-api/blob/work-for-release-0.21.x/usage/src/main/java/org/killbill/billing/usage/plugin/api/UsagePluginApi.java#L25)


invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
```diff
@@ -40,7 +40,7 @@
 import org.killbill.billing.invoice.generator.InvoiceWithMetadata.TrackingRecordId;
 import org.killbill.billing.invoice.usage.ContiguousIntervalUsageInArrear.UsageInArrearItemsAndNextNotificationDate;
 import org.killbill.billing.junction.BillingEvent;
-import org.killbill.billing.usage.RawUsage;
+import org.killbill.billing.usage.plugin.api.RawUsageRecord;
```
pierre: Do we want the internal modules to rely on `plugin-api` objects explicitly? Shouldn't the `usage` module transforms `org.killbill.billing.usage.plugin.api.RawUsageRecord` into a non-plugin version?

For the catalog for instance, only the `catalog` module relies on `VersionedPluginCatalog` (from `plugin-api`) and transforms it into a `DefaultVersionedCatalog`.

usage/src/main/java/org/killbill/billing/usage/api/BaseUserApi.java
```diff
@@ -0,0 +1,74 @@
...
+        for (final String service : allServices) {
+            final UsagePluginApi plugin = pluginRegistry.getServiceForName(service);
+
+            final List<RawUsageRecord> result = plugin.geUsageForAccount(startDate, endDate, tenantContext);
+            // First plugin registered, returns result -- could be empty List if no usage was recorded.
+            if (result != null) {
+
+                for (final RawUsageRecord cur : result) {
+                    if (cur.getDate().compareTo(startDate) < 0 || cur.getDate().compareTo(endDate) >=0) {
+                        logger.warn("Usage plugin returned usage data with date %s, not in the specified range [%s -> %s[",
```
pierre: `%s` should be `{}`.

usage/src/main/java/org/killbill/billing/usage/glue/UsageModule.java
```diff
@@ -46,10 +50,16 @@ protected void installInternalUserApi() {
     }
 
 
+    protected void installCatalogPluginApi() {
```
pierre: Not `catalog` :-)

usage/src/main/java/org/killbill/billing/usage/api/user/DefaultUsageUserApi.java
```diff
@@ -76,27 +83,66 @@ public void recordRolledUpUsage(final SubscriptionUsageRecord record, final Call
...
+                }
             }
             prevDate = curDate;
         }
         return result;
     }
 
+    private List<RolledUpUnit> getRolledUpUnitsForRawPluginUsage(final UUID subscriptionId, @Nullable final String unitType, final List<RawUsageRecord> rawAccountUsage) {
+        final Map<String, Long> tmp = new HashMap<String, Long>();
+        for (RawUsageRecord cur : rawAccountUsage) {
```
pierre: Nit: some `final` qualifiers missing.

invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
sbrossie: Yes, i went back and forth with that. Fully agree it would be cleaner, and was expecting to discuss that with you.

The benefits of using the RawUsageRecord is to avoid copying/converting every single usage record into a new model that is likely to be the exact same one. In the case of the catalog, we had no choice because the catalog objects are not dummy POJO, they have a lot of business logics encapsulated. In the usage case, those are strictly POJO. I think a better approach would be to define RawUsageRecord in our api module and use it both in the plugin api and KB core, but before i embark in this change i wanted to have the conversation.

usage/src/main/java/org/killbill/billing/usage/api/BaseUserApi.java
sbrossie: Will fix!

usage/src/main/java/org/killbill/billing/usage/glue/UsageModule.java
sbrossie: Will fix...

invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
pierre: > In the usage case, those are strictly POJO. I think a better approach would be to define RawUsageRecord in our api module and use it both in the plugin api and KB core, but before i embark in this change i wanted to have the conversation.

Let's talk, but I think it's a cleaner approach indeed.

PR #1090, opened 2019-01-29
invoice: reduce number of queries when creating invoices

pierre:
Instead of fetching existing invoices and invoice items one by one in the `createInvoices` transaction (e.g. `AUTO_INVOICING_REUSE_DRAFT` use case), bulk retrieve all of them.

In the case of regular invoice runs, this will likely speed up things, because we already have the information (we save at least 2 queries per invoice run, linear with the number of invoice items being created). However, in the case of direct API calls (e.g. create adjustment) and HA (bus listener), this will likely slow things down (in most cases, we will retrieve more information than actually needed). Since these are infrequent codepaths, it's probably okay (we could optimize further depending on the codepath, but it would increase the DAO logic complexity).

@sbrossie FYI, this is orthogonal to the caching work we were talking about. Just randomly came across this as I was investigating some perf data.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/dao/ExistingInvoiceMetadata.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/dao/ExistingInvoiceMetadata.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 
+import org.killbill.billing.callcontext.InternalTenantContext;
 import org.killbill.billing.invoice.api.Invoice;
 import org.killbill.billing.invoice.api.InvoiceItem;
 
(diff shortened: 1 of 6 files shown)


invoice/src/main/java/org/killbill/billing/invoice/dao/ExistingInvoiceMetadata.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import org.killbill.billing.callcontext.InternalTenantContext;
+import org.killbill.billing.invoice.api.Invoice;
+import org.killbill.billing.invoice.api.InvoiceItem;
+
+public class ExistingInvoiceMetadata {
```
sbrossie: Nit: the naming does not really convey what this object does, but i know this is difficult... maybe something like `InvoiceCacheLoader` ?

pierre: Indeed, it's not really good - but I cannot think of anything else 😖 However, I don't like `InvoiceCacheLoader` because `*CacheLoader` refers to caching loaders already in Kill Bill (and this is unrelated to caches).

I'll merge as-is for now, but let's keep thinking...

Issue #35, opened 2012-11-10
Revisit who creates ids


pierre:
We create ids in various places today. For example, JAX-RS will create the id for an AccountEmail in AccountEmailJson but account ids are created in DefaultAccountUserApi by AccountModelDao.

We should centralize id generation.

PR #1093, opened 2019-02-08
EntitySqlDaoWrapperInvocationHandler optimizations

pierre:
⚠️ This should be looked at after https://github.com/killbill/killbill/pull/1092 is merged (but no rush).

* Revisit the code to limit the number of queries in various codepaths: by looking at the `ChangeType` and `ObjectType` (i.e. with or without history), we can avoid unnecessary queries in some cases. This makes the code a bit harder to follow, but I've added a few `Preconditions` to ensure my assumptions are correct.
* Switch to bulk queries (e.g. `getByIds`).

To be enjoyed with ☕️... 😼 

FYI, after this, there are two main items left on my TODO list (but we should already start advertising the changes and measuring the performance impact):

* Investigate if we can insert history records in bulk (this might require some `jdbi` changes unfortunately, I'm currently looking into it)
* Fix `InvoiceTrackingSqlDao` auditing (found a bug while refactoring this code)


util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDao.java
```diff
@@ -31,23 +31,27 @@
...
 import org.skife.jdbi.v2.sqlobject.customizers.Define;
 import org.skife.jdbi.v2.sqlobject.mixins.CloseMe;
 import org.skife.jdbi.v2.sqlobject.mixins.Transactional;
+import org.skife.jdbi.v2.util.LongMapper;
 
 @KillBillSqlDaoStringTemplate
 public interface EntitySqlDao<M extends EntityModelDao<E>, E extends Entity> extends AuditSqlDao, HistorySqlDao<M, E>, Transactional<EntitySqlDao<M, E>>, CloseMe {
 
     @SqlUpdate
+    @GetGeneratedKeys(value = LongMapper.class)
```
sbrossie: Slightly tricky though: "If the columns which represent the auto-generated keys were not specified, the JDBC driver implementation will determine the columns which best represent the auto-generated keys."

Also unrelated to this specific PR, but if this is the direction to go, we should then think of [fixing our queue code](https://github.com/killbill/killbill-commons/blob/master/queue/src/main/java/org/killbill/queue/DBBackedQueue.java#L762).

util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDao.java
```diff
@@ -31,23 +31,27 @@
...
     @Audited(ChangeType.INSERT)
     public Object create(@SmartBindBean final M entity,
                          @SmartBindBean final InternalCallContext context);
 
     @SqlBatch
+    // We don't @GetGeneratedKeys here, as it's unclear if the ordering through the batches is respected by all JDBC drivers
```
sbrossie: The JDBI 3 (but 2 is probably no different) [seems to indicate we can use it](http://jdbi.org/#__getgeneratedkeys_3). 

I suppose in the current scheme, the code follows with a get to retrieve all objects and access the auto-generated fields?

util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDaoWrapperInvocationHandler.java
```diff
@@ -210,16 +211,20 @@ public Object execute() throws Throwable {
...
         final ChangeType changeType = auditedAnnotation.value();
+        final boolean isBatchQuery = method.getAnnotation(SqlBatch.class) != null;
 
         // Get the current state before deletion for the history tables
-        final Map<String, M> deletedEntities = new HashMap<String, M>();
+        final Map<Long, M> deletedEntities = new HashMap<Long, M>();
         if (changeType == ChangeType.DELETE) {
             for (final String entityId : entityIds) {
-                deletedEntities.put(entityId, sqlDao.getById(entityId, context));
+                // TODO Switch to getByIds
```
sbrossie: This seems reasonable, but do we even have use case of delete operations in batch?

util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDaoWrapperInvocationHandler.java
```diff
@@ -236,11 +241,79 @@ public Object execute() throws Throwable {
...
 
-        // PERF: override the return value with the reHydrated entity to avoid an extra 'get' in the transaction,
-        // (see EntityDaoBase#createAndRefresh for an example, but it works for updates as well).
-        final List<M> ms = updateHistoryAndAudit(entityIds, deletedEntities, changeType, context);
-        final boolean isBatchQuery = method.getAnnotation(SqlBatch.class) != null;
-        return isBatchQuery ? ms : Iterables.<M>getFirst(ms, null);
+        InternalCallContext context = null;
+        // Retrieve record_id(s) for audit and history tables
+        final List<Long> entityRecordIds = new LinkedList<Long>();
+        if (changeType == ChangeType.DELETE) {
```
sbrossie: Am i understanding well, that all this new code is because we to optimize the rehydration process by using `getByRecordId` instead of `getBydId`?

pierre:
> * What i don't know yet if i like:
>   
>   * Can we measure the improvements -- i know it's unfair to ask for this question at this stage :-) ?

There are really two distinct parts in this work:

* Bulk queries stuff (e.g. invoice with lots of invoice items): I don't think we'll expect a performance boost in all cases. It will depend on a lot of factors, including dataset, type of database engine and jdbc configuration parameters.
* Reducing the number of queries in some codepaths: hard to measure, but we're for sure saving a few round trips 😼 

> * What i like less:
>   
>   * The additional two paths from bulk v.s non bulk -- if we could optimize bulk and treat everything as bulk that would be ideal. I realize this may too much to ask for...

Yup, that's being worked on in the branch `invoice-performance-v5`, but it requires some `jdbi` changes. PRs are coming...

util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDao.java
pierre: > Slightly tricky though: "If the columns which represent the auto-generated keys were not specified, the JDBC driver implementation will determine the columns which best represent the auto-generated keys."

Should be fairly safe though, as it works fine with H2/PostgreSQL/MySQL. Note that I could specify it in the annotation (`@GetGeneratedKeys(columnName = ...)`), but casing is a bit annoying (H2 would want `RECORD_ID` instead of `record_id`).

That being said, it would be nice to completely get rid of `record_id` and only rely on `id` (somewhat related: https://github.com/killbill/killbill/issues/35). This would also unblock the Spanner work (TODO mostly aspirational at this point...).

> Also unrelated to this specific PR, but if this is the direction to go, we should then think of [fixing our queue code](https://github.com/killbill/killbill-commons/blob/master/queue/src/main/java/org/killbill/queue/DBBackedQueue.java#L762).

We could potentially - one thing it keep in mind though is that `LAST_INSERT_ID()` is scoped to a connection while `RETURN_GENERATED_KEYS` is scoped to a `Statement` (would have to refresh my memory on the queue code if it matters in that case).

util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDao.java
pierre: > The JDBI 3 (but 2 is probably no different) [seems to indicate we can use it](http://jdbi.org/#__getgeneratedkeys_3).

Our version of JDBI doesn't support `@GetGeneratedKeys` on batch statements, but I'm currently fixing it.

But the remark here was more about the **order** of keys returned: it is driver specific whether the ordering returned matches the ordering of inserts.

> I suppose in the current scheme, the code follows with a get to retrieve all objects and access the auto-generated fields?

Yes. Note also that for `INSERT` with a history table, we need the full rehydrated object anyways.

util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDaoWrapperInvocationHandler.java
pierre: We don't indeed - just making the code forward compatible 😆

util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDaoWrapperInvocationHandler.java
pierre: Not quite: the gist of it is that we have a matrix of cases `tableName` x `changeType` x `with or without history`. In some cases, we only need the new `record_id`, in others we need the old `entity` and the `new entity`, etc.

The previous code was a bit more generic and wasn't explicitly enumerating these matrix cases, causing extra unnecessary `get` queries to be issued (IIRC, I saw 2 unnecessary extra `getInvoiceItem` queries for each `createInvoiceItem`). I now go in each of these cases to try to optimize as much as possible. Code is a bit more verbose but more obvious IMHO (and I've added `Preconditions` to validate some of my assumptions).

PR #1096, opened 2019-02-12
Invoice performance v5

pierre:
~⚠️ This should be looked at once https://github.com/killbill/killbill/pull/1093 is merged.~

~⚠️ **DO NOT MERGE** https://github.com/killbill/killbill-commons/pull/60 needs to be merged first.~

* Bulk insert history entries
* Optimize Batch inserts codepath: rely on `@GetGeneratedKeys` to avoid extra queries and simplify the branching in `EntitySqlDaoWrapperInvocationHandler` (code review from last PR)

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/dao/HistorySqlDao.java
+++ b/util/src/main/java/org/killbill/billing/util/dao/HistorySqlDao.java
@@ -1,7 +1,7 @@
 /*
  * Copyright 2010-2013 Ning, Inc.
- * Copyright 2014-2017 Groupon, Inc
- * Copyright 2014-2017 The Billing Project, LLC
+ * Copyright 2014-2019 Groupon, Inc
+ * Copyright 2014-2019 The Billing Project, LLC
  *
  * The Billing Project licenses this file to you under the Apache License, version 2.0
  * (the "License"); you may not use this file except in compliance with the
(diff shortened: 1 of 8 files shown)


util/src/main/java/org/killbill/billing/util/dao/HistorySqlDao.java
```diff
@@ -39,8 +41,10 @@
     public List<EntityHistoryModelDao<M, E>> getHistoryForTargetRecordId(@Define("bypassMappingRegistryCache") final boolean bypassMappingRegistryCache,
                                                                          @Bind("targetRecordId") final long targetRecordId,
                                                                          @SmartBindBean InternalTenantContext context);
-    @SqlUpdate
-    @GetGeneratedKeys
-    public Long addHistoryFromTransaction(@EntityHistoryBinder EntityHistoryModelDao<M, E> history,
-                                          @SmartBindBean InternalCallContext context);
+
+    @SqlBatch
+    @BatchChunkSize(1000) // Arbitrary value, just a safety mechanism in case of very large datasets
```
sbrossie: That seems like a large value - i think we will need to test the perf associated with these batches. I have the same concern with my notification work. The main concern i have is that the larger the value i suspect the longer the operation, and i am worried about concurrent operations waiting on the table lock(s). It's more a con cern for the notification work though.

PR #1101, opened 2019-02-19
MDC logging bug fixes

pierre:
See commit by commit for bug descriptions.

### DIFF
--- a/payment/src/main/java/org/killbill/billing/payment/api/DefaultPaymentApi.java
+++ b/payment/src/main/java/org/killbill/billing/payment/api/DefaultPaymentApi.java
@@ -101,9 +101,9 @@ public class DefaultPaymentApi extends DefaultApiBase implements PaymentApi {
         PaymentTransaction paymentTransaction = null;
         PaymentApiException exception = null;
         try {
+            final InternalCallContext internalCallContext = internalCallContextFactory.createInternalCallContext(account.getId(), callContext);
             logEnterAPICall(log, transactionType, account, paymentMethodId, paymentId, null, amount, currency, paymentExternalKey, paymentTransactionExternalKey, null, null);
 
-            final InternalCallContext internalCallContext = internalCallContextFactory.createInternalCallContext(account.getId(), callContext);
             payment = paymentProcessor.createAuthorization(IS_API_PAYMENT, NULL_ATTEMPT_ID, account, paymentMethodId, paymentId, amount, currency, effectiveDate, paymentExternalKey, paymentTransactionExternalKey,
                                                            null, null, SHOULD_LOCK_ACCOUNT, properties, callContext, internalCallContext);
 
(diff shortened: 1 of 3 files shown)


payment/src/main/java/org/killbill/billing/payment/dispatcher/CallableWithRequestData.java
```diff
@@ -72,7 +72,8 @@ public T call() throws Exception {
             UUIDs.setRandom(null);
```
sbrossie: Unrelated to your change but this whole `finally` block is a crazy amount of thread reset state!

payment/src/main/java/org/killbill/billing/payment/dispatcher/CallableWithRequestData.java
```diff
@@ -72,7 +72,8 @@ public T call() throws Exception {
...
             ThreadContext.unbindSecurityManager();
             ThreadContext.unbindSubject();
-            MDC.clear();
+            // Reset the MDC as before the call, in case the plugin fiddle'd with it
+            MDC.setContextMap(mdcContextMap);
```
sbrossie: This 'plugin' thread is going back to the pool - don;t we want to clear the state completely - also how do we know this was the previous state?

payment/src/main/java/org/killbill/billing/payment/api/DefaultPaymentApi.java
```diff
@@ -898,9 +897,9 @@ public Payment createChargebackReversalWithPaymentControl(final Account account,
         PaymentTransaction paymentTransaction = null;
         PaymentApiException exception = null;
         try {
+            final InternalCallContext internalCallContext = internalCallContextFactory.createInternalCallContext(account.getId(), callContext);
```
sbrossie: Very tricky, i had to read this diff several times to realize this issue has to do with the ordering of creating the context  - and calling `populateMDCContext` -- versus doing the first log line. I am wondering if there are not more places where we have this issue, and how we gonna maintain this ?

pierre: > Very tricky, i had to read this diff several times to realize this issue has to do with the ordering of creating the context - and calling `populateMDCContext` -- versus doing the first log line.

That's why I hinted you at looking at my individual commit messages 😃 

> I am wondering if there are not more places where we have this issue, and how we gonna maintain this ?

Yup, I was wondering the same thing unfortunately. The `InternalCallContextFactory` is really **the** centralized place where we resolve the record ids for the log lines, so it kind of makes sense to populate the MDC here. Also, one of the first things we do going from API to internal code is to create the `internal` contexts. So, everything points to the fact that it's the right place/moment to do this sort of things and we are fairly consistent. But, it's definitively not bullet proof...

Sorry, I don't have a good answer here 🙁 I think we need to be due diligent when writing new log lines and inspect them manually. Alternatively, we could try to add some preconditions in the code, but it's non trivial (would have to facade all SLF4J calls probably).

payment/src/main/java/org/killbill/billing/payment/dispatcher/CallableWithRequestData.java
pierre: Absolutely - fixed.

PR #1103, opened 2019-02-22
Fix for 1076

sbrossie:


payment/src/test/java/org/killbill/billing/payment/core/TestPaymentMethodProcessorWithDB.java
```diff
@@ -99,7 +99,6 @@ public void testDeletePaymentMethod() throws Exception {
         Assert.assertEquals(history2.getAccountRecordId(), paymentMethodModelDao.getAccountRecordId());
         Assert.assertEquals(history2.getTenantRecordId(), paymentMethodModelDao.getTenantRecordId());
         Assert.assertEquals(history2.getExternalKey(), paymentMethodModelDao.getExternalKey());
-        // Note: upon deletion, the recorded state is the same as before the delete
```
pierre: See, we knew about it! ;-)

util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDaoWrapperInvocationHandler.java
```diff
@@ -276,7 +258,7 @@ public Object execute() throws Throwable {
             }
         } else {
             // For updates, easiest is to go back to the database
```
pierre: Nit: comment should be updated.

sbrossie: Ack, will do.

payment/src/test/java/org/killbill/billing/payment/core/TestPaymentMethodProcessorWithDB.java
sbrossie: If by 'we' you mean the common pool of knowledge that we both have, then 'yes' !

PR #1108, opened 2019-02-27
Entitlement plugin api fixes

pierre:
This also fixes logging around `targetDates` in invoice.

### DIFF
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultBaseEntitlementWithAddOnsSpecifier.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultBaseEntitlementWithAddOnsSpecifier.java
@@ -1,5 +1,6 @@
 /*
- * Copyright 2016 The Billing Project, LLC
+ * Copyright 2014-2019 Groupon, Inc
+ * Copyright 2014-2019 The Billing Project, LLC
  *
  * The Billing Project licenses this file to you under the Apache License, version 2.0
  * (the "License"); you may not use this file except in compliance with the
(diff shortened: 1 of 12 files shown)


entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultBaseEntitlementWithAddOnsSpecifier.java
```diff
@@ -48,11 +60,19 @@ public UUID getBundleId() {
         return bundleId;
     }
 
+    public void setBundleId(final UUID bundleId) {
```
sbrossie: Mixed reviews... the object was immutable, now it's not. I know it's convenient, but i am wondering if creating another object would not be more maintainable. I am thinking all the payment context for which we have setters, and it's hard to follow the code. if you want to keep it this way, not a strong objection either.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
```diff
@@ -402,7 +402,7 @@ public void uncancelEntitlement(final Iterable<PluginProperty> properties, final
         final WithEntitlementPlugin<Void> uncancelEntitlementWithPlugin = new WithEntitlementPlugin<Void>() {
 
             @Override
-            public Void doCall(final EntitlementApi entitlementApi, final EntitlementContext updatedPluginContext) throws EntitlementApiException {
+            public Void doCall(final EntitlementApi entitlementApi, final DefaultEntitlementContext updatedPluginContext) throws EntitlementApiException {
```
sbrossie: Why can't we keep the interface?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementContext.java
```diff
@@ -117,6 +119,16 @@ public DefaultEntitlementContext(final OperationType operationType,
         this.tenantId = tenantId;
     }
 
+    private static <T> Iterable<T> merge(final Iterable<T> prevValues, final Iterable<T> newValues) {
```
sbrossie: I would expect a `merge` method to merge the two input into one - but looks like instead we either return one or the other ? Isn't more like a `lastNonEmptyOrNull` kind of thing?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultBaseEntitlementWithAddOnsSpecifier.java
pierre: Unfortunately, all our code is structure around:

```
public interface WithEntitlementPlugin<T> {
    T doCall(final EntitlementApi entitlementApi, final DefaultEntitlementContext updatedPluginContext) throws EntitlementApiException;
}
```

`T` could be `UUID`, `List<UUID>`, etc. So the context has to be mutable (passed in as argument).

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
pierre: `EntitlementContext` comes from `plugin-api` and doesn't have the setter.

I could have cast'ed, but I prefer being explicit (I almost renamed `DefaultEntitlementContext` as `MutableEntitlementContext` but the changes would be huge).

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementContext.java
pierre: Indeed - fixed in 30e911ba0ac468d5d9356ce9a5da270d4cadfc9e.

PR #1118, opened 2019-03-12
Fix for 1060

sbrossie:


subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -524,7 +527,7 @@ public long getLastEventOrderedId() {
         return it.hasNext() ? ((SubscriptionBaseTransitionData) it.next()).getTotalOrdering() : -1L;
     }
 
-    public List<SubscriptionBillingEvent> getBillingTransitions() {
+    public List<SubscriptionBillingEvent> getBillingTransitions(final Catalog catalog) throws CatalogApiException {
```
pierre: Worth adding small unit tests just for this (since it's the core of the change and the logic is... subtle 😅)?

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBillingEvent.java
```diff
@@ -79,12 +74,20 @@ public Integer getBcdLocal() {
     public String toString() {
         return "DefaultSubscriptionBillingEvent{" +
                "type=" + type +
-               ", planName='" + planName + '\'' +
-               ", planPhaseName='" + planPhaseName + '\'' +
+               ", plan='" + plan + '\'' +
```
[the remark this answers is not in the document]
sbrossie: Yes, will fix, and print the name only.

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBillingEvent.java
```diff
@@ -79,12 +74,20 @@ public Integer getBcdLocal() {
...
                ", effectiveDate=" + effectiveDate +
                ", totalOrdering=" + totalOrdering +
-               ", lastChangePlanDate=" + lastChangePlanDate +
                ", bcdLocal=" + bcdLocal +
                '}';
     }
+
+    @Override
+    public int compareTo(final SubscriptionBillingEvent o) {
+        if (!getEffectiveDate().equals(o.getEffectiveDate())) {
```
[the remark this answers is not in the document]
sbrossie: Fixed in https://github.com/killbill/killbill/commit/87ea0ec55292e73d08fc19f1c941dc180da8a3d3#diff-6cedcf8423e49fd6497bb402ecd81d9cR90

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
sbrossie: yep... see https://github.com/killbill/killbill/commit/87ea0ec55292e73d08fc19f1c941dc180da8a3d3#diff-52efa6b66884d11316c9bde8c72f90e4R45

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBillingEvent.java
```diff
@@ -79,12 +76,24 @@ public Integer getBcdLocal() {
     public String toString() {
         return "DefaultSubscriptionBillingEvent{" +
                "type=" + type +
-               ", planName='" + planName + '\'' +
-               ", planPhaseName='" + planPhaseName + '\'' +
+               ", plan='" + plan.getName() + '\'' +
```
pierre: Do we need a `null` check here (e.g. for cancellations)?

sbrossie: From construction, `plan` and `planPhase` should never be null -- in case of cancel we take the  prev. transition `plan`

PR #1131, opened 2019-04-14
Work for payment control plugin fixes

sbrossie:


payment/src/main/java/org/killbill/billing/payment/core/sm/control/PaymentStateControlContext.java
```diff
@@ -48,6 +48,9 @@
     private BigDecimal processedAmount;
     private Currency processedCurrency;
 
+    private String originalPluginName;
```
pierre: Maybe rename it to `originalPaymentPluginName`?

PR #1142, opened 2019-06-03
payment: propagate plugin exception for better logging

pierre:


sbrossie:
@pierre Are we gonna release `0.20.x` for this fix?

PR #1148, opened 2019-06-19
server: fix Shiro regression

pierre:
Make sure we use our fix for https://issues.apache.org/jira/browse/SHIRO-540.


pierre:
FYI `integration-tests` are failing because there isn't a corresponding branch in that repo.

PR #1150, opened 2019-06-21
DDL cleanups

pierre:


util/src/test/resources/org/killbill/billing/util/ddl_test.sql
```diff
@@ -1,5 +1,126 @@
 /*! SET default_storage_engine=INNODB */;
 
+DROP TABLE IF EXISTS accounts;
```
sbrossie: Nit: I realize you extracted those hardcoded DDL update from our tests, so this is already better, but fundamentally it sucks that we have to duplicate such tables into our test resources just because they are not available.

pierre:
@sbrossie Once the work on 21.x is stabilized, we should merge this PR.

pierre:
Verified integration tests locally:

```
Finished in 1161.868347 seconds.

185 tests, 10056 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

PR #1162, opened 2019-07-12
Work for 1157

sbrossie:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -168,6 +168,26 @@ public Response getSubscription(@PathParam("subscriptionId") final UUID subscrip
...
 
+    @TimedResource
+    @GET
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Retrieve a subscription by external key", response = SubscriptionJson.class)
+    @ApiResponses(value = {@ApiResponse(code = 404, message = "Subscription not found")})
+    public Response getSubscriptionByKey(@ApiParam(required = true) @QueryParam(QUERY_EXTERNAL_KEY) final String externalKey,
+                                         // The required part is a bit forceful for the user though... but removing would require not using our CatalogInternalApi down the line
+                                         // which was made specifically to ensure that catalog plugin get what they need.
+                                         @ApiParam(required = true) @QueryParam(QUERY_ACCOUNT_ID) final UUID accountId,
```
pierre: I don't think we should require the `accountId`: the point of this API is to retrieve the subscription with only the key (the associated account might not be known).

Take a look at how we do it for `getActiveSubscriptionBundleForExternalKey`: there's a dao call to first get the mapping between the key and the id. From the id, we can then build the right context (`createInternalTenantContext(final UUID objectId, final ObjectType objectType, final TenantContext context)`) and plug into the methods to fully populate the object.

subscription/src/test/java/org/killbill/billing/subscription/engine/dao/TestSubscriptionDao.java
```diff
@@ -312,6 +313,48 @@ public void testWithAuditAndHistory() throws SubscriptionBaseApiException {
 
 
     @Test(groups = "slow")
+    public void testSubscriptionExternalKey() throws SubscriptionBaseApiException, CatalogApiException {
+        final String eternalKey = "6577564455sgwers2";
```
pierre: Typo (`externalKey`).

entitlement/src/main/java/org/killbill/billing/entitlement/logging/EntitlementLoggingHelper.java
```diff
@@ -56,7 +56,7 @@ private static void logCreateEntitlementWithAOs(final StringBuilder logLine,
                                                     final LocalDate entitlementDate,
                                                     final LocalDate billingDate) {
         if (externalKey != null) {
-            logLine.append("key='")
+            logLine.append("bundle key='")
```
pierre: `bundleKey` or `bundleExternalKey` (no space).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -168,6 +168,26 @@ public Response getSubscription(@PathParam("subscriptionId") final UUID subscrip
         return Response.status(Status.OK).entity(json).build();
     }
 
+    @TimedResource
+    @GET
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Retrieve a subscription by external key", response = SubscriptionJson.class)
+    @ApiResponses(value = {@ApiResponse(code = 404, message = "Subscription not found")})
```
pierre: For this to work, I believe `EntitlementApiExceptionMapper` needs to be updated to handle the new error code.

subscription/src/main/resources/org/killbill/billing/subscription/migration/V20190712115638__subscription_external_key.sql
```diff
@@ -0,0 +1,8 @@
+alter table subscriptions add column external_key varchar(255) after bundle_id;
+update subscriptions set external_key=id;
```
pierre: To be truly backward compatible, I'm wondering if this shouldn't be set to the bundle external key instead? The query is very hard to write though (with transfers in mind). Something to call out in the release notes.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
sbrossie: ack, see https://github.com/killbill/killbill/commit/ad582da82dde399c0c60aa3c40a346b6a0551532

subscription/src/main/resources/org/killbill/billing/subscription/migration/V20190712115638__subscription_external_key.sql
sbrossie: ⚠️ Yep, this is currently not backward compatible, and worst than that slightly sneaky because the definition of what `Subscription#externalKey` means has changed. We will for sure need to document it at the very least.

PR #1164, opened 2019-07-17
Work for release 0.21 review

sbrossie:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -218,7 +218,6 @@ public Response getSubscriptionEventAuditLogsWithHistory(@PathParam("eventId") f
     @ApiOperation(value = "Create an subscription", response = SubscriptionJson.class)
     @ApiResponses(value = {@ApiResponse(code = 201, message = "Subscription created successfully")})
     public Response createSubscription(final SubscriptionJson subscription,
-                                       @ApiParam(hidden = true) @Deprecated @QueryParam(QUERY_REQUESTED_DT) final String requestedDate, /* This is deprecated, only used for backward compatibility */
```
pierre: Worth double checking we don't use it ourselves in tests and/or Kaui...

catalog/src/main/resources/org/killbill/billing/catalog/migration/V20190717131708__rename_size_field.sql
```diff
@@ -0,0 +1 @@
+alter table catalog_override_block_definition change `size` `bsize` decimal(15,9) NOT NULL;
```
pierre: Don't use backticks, it's a MySQL-only thing.

catalog/src/main/resources/org/killbill/billing/catalog/migration/V20190717161524__recreate_index_tier_definition.sql
```diff
@@ -0,0 +1,2 @@
+drop index catalog_override_tier_definition_idx on catalog_override_usage_definition;
+create index catalog_override_tier_definition_idx on catalog_override_tier_definition(tenant_record_id, currency);
```
pierre: This index is weird though, no? Cardinality will be quite low in most cases.

PR #1173, opened 2019-07-25
Fix for 1170

sbrossie:


sbrossie:
@pierre for review

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithBCDUpdate.java
```diff
@@ -862,4 +864,32 @@ public void testBCDChangeForConsumableInArrearPlan() throws Exception {
...
+    public void testWithOverrideFixedPriceAndNewBCD() throws Exception {
+
+        final DateTime initialDate = new DateTime(2016, 4, 1, 0, 13, 42, 0, testTimeZone);
+        clock.setTime(initialDate);
+
+        final Account account = createAccountWithNonOsgiPaymentMethod(getAccountData(0));
+        assertNotNull(account);
+
+        // Price override of $10 for fixed and $0 for recurring
+        // Override BCD as well
```
pierre: The account BCD isn't set here though, would it be worth adding a second test where the account has a non zero BCD?

sbrossie: I certainly could, but that seems orthogonal to the issue? The issue is simply that that if there is BCD event in the stream on the same date, we discard fixed price from previous event, and all this is independent of whether account BCD was set or not

pierre: Fair enough 👍

PR #1175, opened 2019-07-27
Work for release 0.21 review

sbrossie:


util/src/main/java/org/killbill/billing/util/customfield/dao/DefaultCustomFieldDao.java
```diff
@@ -57,6 +57,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Preconditions;
```
pierre: Extra `import`.

invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemFactory.java
```diff
@@ -141,7 +141,14 @@ public static InvoiceItem fromModelDaoWithCatalog(final InvoiceItemModelDao invo
...
 
                 if (planName != null) {
-                    final Plan plan = catalog.findPlan(planName, createdDate);
+
+                    // We don't know the creation date of the said subscription, all we know is the time
+                    // of the current event. By specifying KILLBILL_GENESIS, we should at least find a Plan
+                    // and most likely the correct one since we start from the transitionDate and move backward
+                    // Worst case: This is the wrong plan and the prettyName was updated across version, oh well..
+                    //
+                    final DateTime KILLBILL_GENESIS = new DateTime(2019, 10, 28, 0, 0);
```
pierre: ⚠️ Confused by the chosen date? Should it be 2011 instead?

Also,  this should probably be `static final`, and with a timezone.

invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemFactory.java
```diff
@@ -141,7 +141,14 @@ public static InvoiceItem fromModelDaoWithCatalog(final InvoiceItemModelDao invo
...
+                    final Plan plan = catalog.findPlan(planName, transitionDate, KILLBILL_GENESIS);
```
pierre: Should we do the same fix in the Analytics plugin to keep the logic in sync (https://github.com/killbill/killbill-analytics-plugin/blob/f93f95a56c2068037fae6821d78c6df558f4be1e/src/main/java/org/killbill/billing/plugin/analytics/dao/factory/BusinessFactoryBase.java#L331)?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
```diff
@@ -669,9 +669,7 @@ public Response createInstantPayment(@PathParam("invoiceId") final UUID invoiceI
                                          @javax.ws.rs.core.Context final HttpServletRequest request,
                                          @javax.ws.rs.core.Context final UriInfo uriInfo) throws AccountApiException, PaymentApiException {
         verifyNonNullOrEmpty(payment, "InvoicePaymentJson body should be specified");
-        verifyNonNullOrEmpty(payment.getAccountId(), "InvoicePaymentJson accountId needs to be set",
-                             payment.getTargetInvoiceId(), "InvoicePaymentJson targetInvoiceId needs to be set",
-                             payment.getPurchasedAmount(), "InvoicePaymentJson purchasedAmount needs to be set");
```
pierre: I've double checked that passing `null` works as expected: https://github.com/killbill/killbill/blob/381ad0ebfceb5cd660f1fe9bcc05e842a250b0c2/payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java#L677

util/src/main/java/org/killbill/billing/util/customfield/dao/DefaultCustomFieldDao.java
sbrossie: ack will remove

invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemFactory.java
sbrossie: Yes, you are right, this is a typo, was supposed to be 2011, will fix. I can add the timezone, but to match the exact date of our first commit `Fri Oct 28 15:06:57 2011 -0700`. Not sure this part really matters

invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemFactory.java
sbrossie: There is a tack i had opened to track the work on the analytics side: `https://github.com/killbill/killbill-analytics-plugin/issues/99`. I don't know if this is the right fix for this.

This fix was a great way to go around the exception but does not guarantee 100% we are pulling the right version. For a *pretty* string this is good enough but for any price calculation, i am not so sure.

invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemFactory.java
pierre: I meant to say that you should specify `UTC` for the date (`CatalogDateHelper` works in UTC).

PR #1177, opened 2019-08-01
Work for 1174

sbrossie:


invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java
```diff
@@ -378,6 +419,26 @@ public boolean apply(final Tag input) {
...
     }
 
+    private void setParentInvoice(final InvoiceModelDao invoice, final List<Tag> invoicesTags, final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFactory, final InternalTenantContext childContext) {
+        final InvoiceParentChildrenSqlDao invoiceParentChildrenSqlDao = entitySqlDaoWrapperFactory.become(InvoiceParentChildrenSqlDao.class);
+        final List<InvoiceParentChildModelDao> mappings = invoiceParentChildrenSqlDao.getParentChildMappingsByChildInvoiceIds(ImmutableList.of(invoice.getId().toString()), childContext);
+        if (mappings.isEmpty()) {
+            return;
+        }
+
+        Preconditions.checkState(mappings.size() == 1, String.format("Expected only one parent mapping for invoice", invoice.getId()));
```
pierre: ⚠️ Missing format parameter in the `String.format` (also, `Preconditions.checkState` can call `String.format` for you - check the varargs signature one).

invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java
```diff
@@ -227,7 +228,13 @@ public InvoiceItemModelDao createAdjustmentItem(final EntitySqlDaoWrapperFactory
     }
 
     public void populateChildren(final InvoiceModelDao invoice, final List<Tag> invoicesTags, final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFactory, final InternalTenantContext context) {
```
pierre: Do we have HA tests that call this implicitly? Since we have to duplicate the logic, and the HA one is a bit subtle, would be good to make sure we didn't introduce any regression.

sbrossie: This code is pulled anytime there is an invoice generation or a payment associated with an invoice -- just to name those 2 use cases -- so will definitely be part of our test suite for HA. There could be bugs, that we don't see and that are not covered by our tests, but the same could be said of our existing logic.

invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java
sbrossie: Ack will fix.

sbrossie:
> Just concerned about maintainability in the future (if we fix one, the other needs to be fixed as well).

yes i agree, the perf optimization introduces two paths, which will need to be maintained when and if we make code changes in this part of the code; i am also not too thrilled about that :-(

invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java
sbrossie: @pierre Update the PR with additional tests. See 5a10a5fb6027fc5cb

Issue #1183, opened 2019-08-19
Usage invoice item description does not use the pretty name


hendryluk:
This line:
https://github.com/killbill/killbill/blob/e30cdaef205ac51340217679e70dd9d17eb4ae6b/invoice/src/main/java/org/killbill/billing/invoice/model/UsageInvoiceItem.java#L59

Should that be `MoreObjects.firstNotNull(description, prettyUsageName)`?
Currently it's showing the ugly unique-key, which is unhelpful for the customers, and also inconsistent with other implementations of description that's normally in plain english, e.g. https://github.com/killbill/killbill/blob/e30cdaef205ac51340217679e70dd9d17eb4ae6b/invoice/src/main/java/org/killbill/billing/invoice/model/FixedPriceInvoiceItem.java#L61

sbrossie:
@hendryluk I think that makes sense, but please report issues on the mailing list.

PR #1185, opened 2019-08-19
invoice: Make use of prettyUsageName when defined in the item descrip…

sbrossie:
…tion. See #1183


sbrossie:
> Did you double check all other types do the right thing already?

Agreed, see  a6b7e5d454cd5af4cb59bab5ae4ef2574eca0e44

Issue #1193, opened 2019-08-26
callCompletion timeout with AUTO_INVOICING_OFF


pierre:
Reported on the mailing-list: https://groups.google.com/forum/?#!msg/killbilling-users/Lv5EBreXeMY/Gx0DNTseDwAJ (note: overriding the BCD isn't necessary to reproduce the issue).

Related: https://github.com/killbill/killbill/issues/585

PR #1196, opened 2019-08-26
jaxrs: fix callCompletion timeout with AUTO_INVOICING_OFF

pierre:
This fixes https://github.com/killbill/killbill/issues/1193.


pierre:
See https://github.com/killbill/killbill/pull/1234 instead.

PR #1197, opened 2019-08-27
Work for release 0.21 catalog

sbrossie:


catalog/src/main/java/org/killbill/billing/catalog/StandaloneCatalog.java
```diff
@@ -161,7 +162,7 @@ public boolean isTemplateCatalog() {
                (supportedCurrencies == null || supportedCurrencies.length == 0);
     }
 
-    public DefaultPlanRules getPlanRules() {
+    public PlanRules getPlanRules() {
```
pierre: This forces a cast in `DefaultMutableStaticCatalog` - worth the change?

catalog/src/main/java/org/killbill/billing/catalog/StandaloneCatalogWithPriceOverride.java
```diff
@@ -53,16 +55,16 @@
...
         // Initialize from input catalog
         setCatalogName(catalog.getCatalogName());
         setEffectiveDate(catalog.getEffectiveDate());
-        setProducts(catalog.getCurrentProducts());
-        setPlans(catalog.getCurrentPlans());
-        setPriceLists(catalog.getPriceLists());
-        setPlanRules(catalog.getPlanRules());
-        setSupportedCurrencies(catalog.getCurrentSupportedCurrencies());
-        setUnits(catalog.getCurrentUnits());
+        setProducts(((StandaloneCatalog) catalog).getCurrentProducts());
```
pierre: The cast is only to handle the exception? Would it be easier instead to propagate the exception to the constructor?

catalog/src/main/java/org/killbill/billing/catalog/api/DefaultCatalogInternalApi.java
```diff
@@ -20,6 +20,7 @@
 import javax.inject.Inject;
 
 import org.killbill.billing.callcontext.InternalTenantContext;
+import org.killbill.billing.catalog.StandaloneCatalog;
```
pierre: Extra import.

catalog/src/main/java/org/killbill/billing/catalog/api/user/DefaultCatalogUserApi.java
```diff
@@ -89,9 +89,9 @@ public DefaultCatalogUserApi(final CatalogService catalogService,
             return fullCatalog;
         } else {
             final DefaultVersionedCatalog filteredCatalog = new DefaultVersionedCatalog(clock);
-            for (final StandaloneCatalog v : fullCatalog.getVersions()) {
+            for (final StaticCatalog v : fullCatalog.getVersions()) {
                 if (v.getEffectiveDate().compareTo(catalogDateVersion.toDate()) >= 0) {
-                    filteredCatalog.add(v);
+                    filteredCatalog.add((StandaloneCatalog) v);
```
pierre: Does `DefaultVersionedCatalog` need to maintain `StandaloneCatalog` versions just because of JAXB (instead of `StaticCatalog`)? If so, would it be easier to hide these casts inside `DefaultVersionedCatalog` so they are not propagated around in the code?

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestInvoice.java
```diff
@@ -62,7 +62,7 @@
 public class TestInvoice extends TestJaxrsBase {
 
     @Test(groups = "slow", description = "Can search and retrieve invoices with and without items")
-    public void testInvoiceOk() throws Exception {
+        public void testInvoiceOk() throws Exception {
```
pierre: Nit: extra spaces added in this file.

catalog/src/test/java/org/killbill/billing/catalog/TestVersionedCatalog.java
```diff
@@ -44,18 +39,6 @@
     // WeaponsHireSmall-1.xml
     final DateTime dt1 = new DateTime("2010-01-01T00:00:00+00:00");
 
-    // WeaponsHireSmall-2.xml
```
pierre: I saw these catalogs were copied over to `subscription`, but do we still need them here in `catalog`?

subscription/src/test/java/org/killbill/billing/subscription/alignment/TestPlanAligner.java
```diff
@@ -56,6 +59,12 @@ public void beforeClass() throws Exception {
         planAligner = new PlanAligner();
     }
 
+    @BeforeMethod(groups = "fast")
+    public void beforeMethod() throws Exception {
+        super.beforeMethod();
```
pierre: You also need the `if (hasFailed()) {return;}` dance.

catalog/src/main/java/org/killbill/billing/catalog/StandaloneCatalogWithPriceOverride.java
```diff
@@ -33,6 +33,7 @@
 import org.killbill.billing.catalog.api.PlanSpecifier;
 import org.killbill.billing.catalog.api.Product;
 import org.killbill.billing.catalog.api.StaticCatalog;
+import org.killbill.billing.catalog.api.rules.PlanRules;
```
pierre: Extra import.

subscription/src/test/java/org/killbill/billing/subscription/engine/dao/TestSubscriptionDao.java
```diff
@@ -95,7 +95,59 @@ public void beforeMethod() throws Exception {
     @Override // to ignore events
     @AfterMethod(groups = "slow")
     public void afterMethod() throws Exception {
-        if (hasFailed()) {
+        if (hasFailed()) {        final String externalKey = "12345";
```
pierre: Copy/paste typo?

catalog/src/main/java/org/killbill/billing/catalog/StandaloneCatalog.java
sbrossie: Yep, i have no choice as this is now a `StaticCatalog` api. However i am missing the `@Override` annotation that would have made that point more clear. Will add it.

catalog/src/main/java/org/killbill/billing/catalog/StandaloneCatalogWithPriceOverride.java
sbrossie: I have a better idea: `https://github.com/killbill/killbill-api/pull/66/commits/5d8f1a7d5943bf058227ecc7ae56f702f4035a13` (the reason for such exceptions was that both `DefaultVersionedCatalog` was implementing both `Catalog` and `StaticCatalog` and exceptions were needed when we could not find the right catalog version.


I will then look to remove casts related to this.

catalog/src/main/java/org/killbill/billing/catalog/api/DefaultCatalogInternalApi.java
sbrossie: Fixed in latest branch.

catalog/src/main/java/org/killbill/billing/catalog/api/user/DefaultCatalogUserApi.java
sbrossie: I see... you really want me to fix these ugly casts 🤣 

I started to make the changes, and use the `@XmlElement#type()` method to keep the interface and still allow JAXB to work properly; this leads to removing some casts but also adds a whole new set. Mostly our objects export quite a few internal apis (`initialize()`, `validate()`,  optimized versions of our `find` apis, ... This was adding more casts than removing so i undid my changes, but i am ready to have the discussion with you again...

catalog/src/test/java/org/killbill/billing/catalog/TestVersionedCatalog.java
sbrossie: I did not copy them -- i created new versions with just the plans we need inside subscriptions -- but for simplicity kept their same name. 

Those are still in use in `catalog` (e.g `TestVersionedCatalogLoader#testLoad` and also through `beatrix` (e.g `TestWithCatalogPlugin`).

subscription/src/test/java/org/killbill/billing/subscription/alignment/TestPlanAligner.java
sbrossie: ack

catalog/src/main/java/org/killbill/billing/catalog/StandaloneCatalogWithPriceOverride.java
sbrossie: ack

subscription/src/test/java/org/killbill/billing/subscription/engine/dao/TestSubscriptionDao.java
sbrossie: ack, removed.

Issue #1025, opened 2018-07-19
Deserialization failure on old events 


sbrossie:
In #1003 we have removed the field `isNotifiedForInvoices`, so if we try to access existing notifications where this field is present we encounter a deserialization error.

**Example of data:**

```
|      8632 | org.killbill.billing.account.api.user.DefaultAccountCreationEvent | {"data":{"externalKey":"0232811865","name":"Louna Huet","firstNameLength":5,"email":"la.bertrand@vaeur.org","billCycleDayLocal":0,"currency":"EUR","paymentMethodId":null,"timeZone":"America/Chihuahua","locale":"fr","address1":"4981 Place des Francs-Bourgeois","address2":"Apt. 174","companyName":"Paris et Lemoine: Extended empowering help-desk","city":"Dunkerque14","stateOrProvince":"Picardie","postalCode":"25848","country":"MU","phone":"+33 372198446","isMigrated":null,"isNotifiedForInvoices":false},"id":"3c9996bb-acc1-4ead-84b8-54ab3b9ea7f8","searchKey1":1675,"searchKey2":1,"userToken":"5fbfdaa8-fd6c-47ed-ae32-028d69944433"} | 5fbfdaa8-fd6c-47ed-ae32-028d69944433 | 2015-01-02 08:00:43 | 802b3d21d3a1   | 802b3d21d3a1     | 2015-01-02 08:00:43       | PROCESSED        |           0 |        1675 |           1 |
```

**Example of error:**
```
[...]

PR #1206, opened 2019-09-10
Initial for #1025 (step 3 only)

sbrossie:


beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
```diff
@@ -535,10 +535,14 @@ public void testWithFullBlockBilling() throws Exception {
 
         // 2012-04-21 -- block billing effective date is 2012-04-01
```
pierre: `2012-05-21`, right?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
```diff
@@ -535,10 +535,14 @@ public void testWithFullBlockBilling() throws Exception {
...
         clock.addDays(20);
-        busHandler.pushExpectedEvents(NextEvent.BLOCK);
+        busHandler.pushExpectedEvents(NextEvent.BLOCK, NextEvent.INVOICE);
         final BlockingState blockingState1 = new DefaultBlockingState(bpSubscription.getBundleId(), BlockingStateType.SUBSCRIPTION_BUNDLE, "state1", "Service", true, true, true, null);
         subscriptionApi.addBlockingState(blockingState1, new LocalDate(2012, 4, 1), ImmutableList.<PluginProperty>of(), callContext);
         assertListenerStatus();
 
+        invoiceChecker.checkInvoice(account.getId(), 3, callContext,
+                                    new ExpectedInvoiceItemCheck(new LocalDate(2012, 5, 1), new LocalDate(2013, 5, 1), InvoiceItemType.REPAIR_ADJ, new BigDecimal("-2399.95")),
```
pierre: There's no repair on the `2012-04-1` - `2012-04-30` period because it's $0?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
sbrossie: Yes, will fix comment.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
sbrossie: Yep, just a `FIXED` price and regardless of amount, we don't repair `FIXED` items.

sbrossie:
>The fix itself is subtle! 😅

Not super subtle but definitely a bit hidden...

> Did you check where else we call isEmpty?

Unfortunately this is really hard to do, because we are calling `Set#isEmpty` so the answer from IntelliJ is not super useful. But my test shows that after removing these 2 places, the test works as expect -- and all circle-ci tests shows green.

pierre:
> Unfortunately this is really hard to do, because we are calling `Set#isEmpty` so the answer from IntelliJ is not super useful.

I'm sure there's a better way, but here's what I did:

* Add `boolean isEmpty();` to the `BillingEventSet` interface
* Remove `extends SortedSet<BillingEvent>` from `BillingEventSet`
* Search for usage

There are 4 results, 2 in `InvoiceDispatcher` and 2 in `FixedAndRecurringInvoiceItemGenerator` :-)

sbrossie:
@pierre Looking at the use case for `BillingEventSet#isEmpty`, there are 3 left in my branch:

* `FixedAndRecurringInvoiceItemGenerator#processRecurringBillingEvents` and `FixedAndRecurringInvoiceItemGenerator#processFixedBillingEvents` still make sense => when there is no billing events, we should not generate any `proposed` items
* `InvoiceDispatcher#processSubscriptionStartRequestedDateWithLock` is used to handle `RequestedSubscriptionInternalEvent` to compute the notification date, i.e `dryRunNotificationTime`, and therefore it also makes sense to ignore those as there would be nothing to notify

Issue #1061, opened 2018-10-29
Janitor UNKNOWN transition may not correctly update invoice balance


sbrossie:
The issue was reported on the [mailing list](https://groups.google.com/forum/#!searchin/killbilling-users/janitor%7Csort:date/killbilling-users/w1YNtBJ7Vxo/yf9GSTF7CAAJ).

By reading the code it looks like this may be a bug but this requires more investigation.

sbrossie:
Also, what seems like another issue related to Janitor not making transition through state machines [mailing list thread](https://groups.google.com/forum/#!topic/killbilling-users/Z4FEkZymPzk)

sbrossie:
For this original issue, it looks like we should keep the `attemptId` -- if any in the notification -- and when processing the notification, if such  `attemptId` exists, recreate a new attempt, therefore calling the various control plugins.

meden:
I can confirm this behavior: when transitioning from `UNKNOWN` to `SUCCESS` the invoice balance is not updated. I managed to nail it down to this patch (against 0.20.10, needs to be rebased):
```
Index: payment/src/main/java/org/killbill/billing/payment/api/DefaultAdminPaymentApi.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- payment/src/main/java/org/killbill/billing/payment/api/DefaultAdminPaymentApi.java	(date 1554364168000)
+++ payment/src/main/java/org/killbill/billing/payment/api/DefaultAdminPaymentApi.java	(date 1562332274224)
@@ -17,6 +17,7 @@
 
 package org.killbill.billing.payment.api;
 
+import java.math.BigDecimal;
 import java.util.List;
 
 import javax.annotation.Nullable;
@@ -116,7 +117,7 @@
                                                            lastSuccessPaymentState,
                                                            paymentTransaction.getId(),
                                                            transactionStatus,
-                                                           paymentTransaction.getProcessedAmount(),
+                                                           paymentSMHelper.isSuccessState(currentPaymentStateName) ? paymentTransaction.getAmount() : BigDecimal.ZERO,
                                                            paymentTransaction.getProcessedCurrency(),
                                                            paymentTransaction.getGatewayErrorCode(),
                                                            paymentTransaction.getGatewayErrorMsg(),

```
I was willing to make a PR, but I had no tests ready (nor I checked the existing ones :disappointed:), so I was quite sure the PR would have been rejected.

[...]

PR #1216, opened 2019-09-18
Make Janitor notification processing invoke the control state machine

pierre:
This fixes https://github.com/killbill/killbill/issues/1061.

The actual fix is https://github.com/killbill/killbill/commit/cd70793c3a972550fd5ca8ab9d90cb5bf25bef87.

Previous commits refactor the code. Currently, `IncompletePaymentTransactionTask` lives side-by-side with `IncompletePaymentAttemptTask`. This PR ensures that the Janitor is driven by `IncompletePaymentAttemptTask` instead, which now calls `IncompletePaymentTransactionTask` whenever needed, so that we always update both state machines in parallel.

The last commit (https://github.com/killbill/killbill/commit/4c903639096d8cdc8a350366a5569c39a6b84852) fixes an odd behavior reported in https://github.com/killbill/killbill/issues/1061 (not directly related to the Janitor though, but touches similar codepaths).


payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentAttemptTask.java
```diff
@@ -31,35 +35,41 @@
...
- * Task to complete 'partially' incomplete payment attempts. Tis only matters for calls that went through PaymentControl apis.
- * <p/>
- * If the state of the transaction associated with the attempt completed, but the attempt state machine did not,
- * we rerun the retry state machine to complete the call and transition the attempt into a terminal state.
- */
-public class IncompletePaymentAttemptTask extends CompletionTaskBase<PaymentAttemptModelDao> {
+import static org.killbill.billing.payment.core.janitor.IncompletePaymentTransactionTask.TRANSACTION_STATUSES_TO_CONSIDER;
+
+// Janitor implementation for payments going through the control APIs.
+// Invoked on-the-fly or as part of the Janitor notification queue
```
sbrossie: I am slightly confused what ` on-the-fly` means -- it thought if this was from the context of a notification thread, this would not be on the fly?

payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentAttemptTask.java
```diff
@@ -155,18 +256,135 @@ public boolean apply(final PaymentTransactionModelDao input) {
...
+                                                                                                                               paymentTransactionInfoPlugin,
+                                                                                                                               attemptNumber,
+                                                                                                                               userToken,
+                                                                                                                               internalTenantContext);
+        final boolean hasTransactionChanged = latestTransactionStatus == null;
+
+        // Don't insert a notification for the on-the-fly Janitor
+        final boolean shouldInsertNotification = attemptNumber != null;
+        if (!hasTransactionChanged && shouldInsertNotification) {
+            insertNewNotificationForUnresolvedTransactionIfNeeded(paymentTransactionId,
```
sbrossie: Is this really necessary since we have the main `doIteration` loop ?

payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentAttemptTask.java
```diff
@@ -155,18 +256,135 @@ public boolean apply(final PaymentTransactionModelDao input) {
...
+                                                                  internalTenantContext.getTenantRecordId());
+        }
+
+        // If there is a payment attempt associated with that transaction, we need to update it as well
+        boolean hasAttemptChanged = false;
+        final PaymentTransactionModelDao paymentTransactionModelDao = paymentDao.getPaymentTransaction(paymentTransactionId, internalTenantContext);
+        if (paymentTransactionModelDao.getAttemptId() != null) {
+            final PaymentAttemptModelDao paymentAttemptModelDao = paymentDao.getPaymentAttempt(paymentTransactionModelDao.getAttemptId(), internalTenantContext);
+            if (paymentAttemptModelDao != null) {
+                if (hasTransactionChanged || retrySMHelper.getInitialState().getName().equals(paymentAttemptModelDao.getStateName())) {
```
sbrossie: What about the `RETRIED` state?

payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentAttemptTask.java
pierre: `on-the-fly` refers to the payment code calling back the plugin during a `GET` operation (as opposed to doing it via the Janitor notification queue).

`IncompletePaymentAttemptTask` and `IncompletePaymentTransactionTask` do the actual Janitor logic. They are called from two places:
* The `Janitor` class, which orchestrates the notification queue and in-memory task.
* The `PaymentRefresher` and `PaymentControlAwareRefresher` classes, which proxy the `GET` APIs

(See also the implementation details on our wiki - if it's still confusing, I can update it)

payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentAttemptTask.java
pierre: Yes: `doIteration` is only called by the in-memory task (for the control payments stuck in an `INIT` state only).

payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentAttemptTask.java
pierre: I don't think it matters (`RETRIED` is a terminal state): either the transaction changed, and we need to re-run the completion code, or the attempt was stuck in `INIT`.

I have a subsequent PR which will verify the behavior for `ABORTED` though (non-terminal state).

Issue #1205, opened 2019-09-10
Blocking-unblocking scenario triggers invoicing issue


sbrossie:
Consider the following scenario, with expected behavior

1. Start a MONTHLY subscription on `2019-4-27` => `RECURRING` [2019-4-27, 2019-5-27)
2. Block using service `S1` on `2019-5-03` => `REPAIR_ADJ` [2019-5-03, 2019-5-27)
3. Block using service `S2`  on `2019-4-27` => `REPAIR_ADJ` [2019-4-27, 2019-05-03)
4. Block using service `S3`  on `2019-5-17` => nothing (still blocked on 2019-4-27)
5. Unblock using service `S1` on `2019-5-03` => nothing (already blocked on 2019-4-27)
6. Unblock using service `S2` on `2019-4-27` => `RECURRING`  [2019-4-27, 2019-05-17) (blocked up to  `2019-5-17` by `S3`).

The test first fails at step `3` because at that point the full period is blocked and so we don't return any billing events (which is correct), but then invoicing code does not even try to go through invoicing run [here](https://github.com/killbill/killbill/blob/killbill-0.20.12/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L337) and [here](https://github.com/killbill/killbill/blob/killbill-0.20.12/invoice/src/main/java/org/killbill/billing/invoice/generator/DefaultInvoiceGenerator.java#L81) and realize that we need to repair the part [2019-4-27, 2019-05-03).

Issue #667, opened 2016-11-16
Should we generate $0 USAGE items 


sbrossie:
The current code will filter our such items.

sbrossie:
Note that we have changed the behavior of the system in  482663493cee8fc8418aa4dbfacb6d9f70b32b5d to now generate $0 recurring items.

sbrossie:
I am reopening this ticket to see if there is a possibility to add some system property to filter out $0 USAGE items as requested by a few folks.

sbrossie:
After discussing more we decided the property would allow exclude to exclude $0 Usage items but only when `quantity=0`

developer-ivolve:
what is the system property.  Can you link that in the docs somewhere?

PR #1217, opened 2019-09-18
Simplify adjustment item logic and harden invoice code

sbrossie:
Pruning existing items is done outside of the tree, which is simpler and more robust (e.g solves #1205). In order to make it work without too many hacks, the code unifies the logic around invoice adjustment items as described in the proposal.

TODO:
* Should we really generate the $0 REPAIR items  -- they could just be generated conceptually for cleanliness of diagrams but we could decide to not generate them on  disk.
* Add unit tests for InvoicePruner
* Revisit invoice Preconditions : I think there are two kinds:
  1. Some need to be there to avoid mis-billing and will lead into __PARK__ state
  2. Some check for inconsistencies in the data -- but as long we we don't mis-bill we should try to be as lenient as possible (with maybe a `strict` mode)
* Add Unit test (and Preconditions) for NodeInterval#rebalance


pierre:
> * Should we really generate the $0 REPAIR items  -- they could just be generated conceptually for cleanliness of diagrams but we could decide to not generate them on  disk.

One argument in favor of generating them would be to be consistent with our $0 recurring and usage story (which we generate): https://github.com/killbill/killbill/issues/667

PR #1221, opened 2019-09-19
invoice: Filter out $0 REPAIR items.

sbrossie:
In scenarios of full item adjustements with a subscription cancelled prior its EOT
the invoice item tree will generate a REPAIR item for the non billable period but because
everything was already credited through full item adj. such item would be generated with a $0 value.

While this makes a lot of sense from a cleanliness of diagrams, this may end up confusing our users, so
filter them out prior writing on disk.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/tree/AccountItemTree.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/tree/AccountItemTree.java
@@ -18,6 +18,7 @@
 
 package org.killbill.billing.invoice.tree;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


invoice/src/main/java/org/killbill/billing/invoice/tree/AccountItemTree.java
```diff
@@ -160,7 +161,13 @@ public void mergeWithProposedItems(final List<InvoiceItem> proposedItems) {
         for (SubscriptionItemTree tree : subscriptionItemTree.values()) {
             final List<InvoiceItem> simplifiedView = tree.getView();
             if (simplifiedView.size() > 0) {
-                result.addAll(simplifiedView);
+                for (final InvoiceItem i : simplifiedView) {
+                    // Filter out $0 REPAIR_ADJ - these items make a lot of sense conceptually for us but may confuse users...
```
pierre: I went back and forth but I'm now thinking we should keep them after all to be consistent with https://github.com/killbill/killbill/issues/667.

It could be confusing but it's also such a corner case scenario (doing full item adjustment and a repair).

sbrossie:
We decided we would keep the $0 `REPAIR` items, therefore closing the PR.

PR #1232, opened 2019-09-30
Fix for 1207

sbrossie:


subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
```diff
@@ -257,7 +245,22 @@ private boolean doCancelPlan(final Map<DefaultSubscriptionBase, DateTime> subscr
...
+                    throw new SubscriptionBaseApiException(ErrorCode.SUB_CANCEL_BAD_STATE, subscription.getId(), currentState);
+                }
                 final DateTime effectiveDate = subscriptions.get(subscription);
+
+                // If subscription was future cancelled at an earlier date we disallow the operation -- i.e,
+                // user should first uncancel prior trying to cancel again
+                final SubscriptionBaseTransition pendingTransition = subscription.getPendingTransition();
+                if (pendingTransition != null &&
+                    pendingTransition.getTransitionType() == SubscriptionBaseTransitionType.CANCEL &&
+                    pendingTransition.getEffectiveTransitionTime().compareTo(effectiveDate) < 0) {
```
pierre: Aren't we guaranteed that `getPendingTransition` returns future transitions only (`FUTURE_ONLY`)? Otherwise, if the check is really needed, do we need `<=` instead?

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCancel.java
```diff
@@ -519,4 +520,54 @@ public void testCancelPlanOnPendingSubscription3() throws SubscriptionBaseApiExc
...
+        subscriptionInternalApi.setChargedThroughDate(subscription.getId(), newChargedThroughDate, internalCallContext);
+        subscription = (DefaultSubscriptionBase) subscriptionInternalApi.getSubscriptionFromId(subscription.getId(), internalCallContext);
+
+        assertEquals(subscription.getLastActiveProduct().getName(), prod);
+        assertEquals(subscription.getLastActivePriceList().getName(), planSet);
+        assertEquals(subscription.getLastActiveBillingPeriod(), term);
+        assertEquals(subscription.getLastActiveCategory(), ProductCategory.BASE);
+
+        // EARLIER CANCELLATION EOT -> 2012-07-01
+        subscription.cancelWithDate(clock.getUTCNow().plusDays(25), callContext);
```
pierre: Do we already have a test with the opposite scenario (cancel on `2012-07-01` while it is already cancelled on `2012-07-06`)?

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
sbrossie: Updated comment in f7755f9fbb1913efff54a3ae9f4f2326897b295f

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCancel.java
sbrossie: Yep, see `TestUserApiCancel#testWithMultipleCancellationEvent`

Issue #1230, opened 2019-09-27
NPE in DefaultInvoiceDao


pierre:
Seen on `0.21.3`:

```
org.killbill.billing.payment.api.PaymentApiException: Internal payment error : java.lang.NullPointerException
	at org.killbill.billing.payment.dispatcher.PaymentPluginDispatcher.dispatchWithExceptionHandling(PaymentPluginDispatcher.java:68)
	at org.killbill.billing.payment.core.sm.OperationCallbackBase.dispatchWithAccountLockAndTimeout(OperationCallbackBase.java:70)
	at org.killbill.billing.payment.core.sm.control.OperationControlCallback.doOperationCallback(OperationControlCallback.java:84)
	at org.killbill.automaton.DefaultOperation.run(DefaultOperation.java:58)
	at org.killbill.automaton.DefaultState.runOperation(DefaultState.java:95)
	at org.killbill.billing.payment.core.sm.PluginControlPaymentAutomatonRunner.run(PluginControlPaymentAutomatonRunner.java:260)
	at org.killbill.billing.payment.core.sm.PluginControlPaymentAutomatonRunner.run(PluginControlPaymentAutomatonRunner.java:130)
	at org.killbill.billing.payment.core.PluginControlPaymentProcessor.createPurchase(PluginControlPaymentProcessor.java:136)
	at org.killbill.billing.payment.api.svcs.DefaultInvoicePaymentInternalApi.createPurchaseForInvoicePayment(DefaultInvoicePaymentInternalApi.java:114)
	at org.killbill.billing.payment.api.DefaultInvoicePaymentApi.createPurchaseForInvoicePayment(DefaultInvoicePaymentApi.java:79)
Caused by: java.lang.NullPointerException: null
	at org.killbill.billing.invoice.dao.DefaultInvoiceDao$33.inTransaction(DefaultInvoiceDao.java:953)
	at org.killbill.billing.invoice.dao.DefaultInvoiceDao$33.inTransaction(DefaultInvoiceDao.java:926)
	at org.killbill.billing.util.entity.dao.EntitySqlDaoTransactionalJdbiWrapper$JdbiTransaction.inTransaction(EntitySqlDaoTransactionalJdbiWrapper.java:74)
	at org.killbill.billing.util.entity.dao.EntitySqlDaoTransactionalJdbiWrapper$JdbiTransaction.inTransaction(EntitySqlDaoTransactionalJdbiWrapper.java:61)
[...]

PR #1233, opened 2019-10-01
invoice: fix NPE in DefaultInvoiceDao

pierre:
In case a duplicate transaction external key was used, invoice would throw a NPE in the priorCall.

This fixes https://github.com/killbill/killbill/issues/1230.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
@@ -437,9 +437,6 @@ public class DefaultInvoiceDao extends EntityDaoBase<InvoiceModelDao, Invoice, I
                 // Bulk insert the invoice items
                 createInvoiceItemsFromTransaction(transInvoiceItemSqlDao, invoiceItemsToCreate, context);
 
-
-
-
                 // CBA COMPLEXITY...
                 //
                 // Optimized path where we don't need to refresh invoices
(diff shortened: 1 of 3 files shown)


invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -963,8 +959,19 @@ public boolean apply(final InvoicePaymentModelDao input) {
...
+                        if (invoicePayment.getPaymentId() == null) {
+                            // Likely invalid state (https://github.com/killbill/killbill/issues/1230) but go ahead nonetheless to run the payment state machine validations
+                            updatedPaymentId = existingAttempt.getPaymentId();
+                        } else if (existingAttempt.getPaymentId() == null) {
+                            updatedPaymentId = invoicePayment.getPaymentId();
+                        } else {
+                            Preconditions.checkState(invoicePayment.getPaymentId().equals(existingAttempt.getPaymentId()),
+                                                     "PaymentAttempt cannot change paymentId: attemptId=%s, newInvoicePaymentId=%s, existingPaymentId=%s", existingAttempt.getId(), invoicePayment.getPaymentId(), existingAttempt.getPaymentId());
+                            updatedPaymentId = invoicePayment.getPaymentId();
+                        }
```
sbrossie: Wouldn't it be simpler to do:
```
if (invoicePayment.getPaymentId() != null) {
   Preconditions.checkState(....);
    updatedPaymentId = invoicePayment.getPaymentId();
} else { 
  // default to paymentAttempt
  updatedPaymentId = existingAttempt.getPaymentId();
}
```

pierre: Not quite - `Preconditions` is only valid if `existingAttempt.getPaymentId() != null`.

But regardless, I wanted to call out the invalid branch explicitly (source of the bug).

Issue #812, opened 2017-10-30
Tagging a Account as AUTO_PAY_OFF/MANUAL_PAY should also apply


sbrossie:
The discussion originated from the [Kill Bill user mailing] list(https://groups.google.com/forum/#!topic/killbilling-users/rhuEbQ3iEGk).

Currently the [behavior](https://github.com/killbill/killbill/blob/killbill-0.18.14/payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java#L333) is to first check if there is a missing payment method and then check for AUTO_PAY_OFF/MANUAL_PAY.

It would make more sense to reverse the check and abort the payment. This would avoid sending an `INVOICE_PAYMENT_FAILURE` event.

Issue #1211, opened 2019-09-13
callCompletion timeout with unexpected events


pierre:
If an entitlement plugin creates custom fields for instance, `callCompletion` will timeout:

```
java.lang.RuntimeException: Unexpected event type CUSTOM_FIELD_CREATION
	at org.killbill.billing.util.userrequest.CompletionUserRequestBase.onBusEvent(CompletionUserRequestBase.java:143)
	at org.killbill.billing.jaxrs.util.KillbillEventHandler.handleSubscriptionEvents(KillbillEventHandler.java:70)
	at sun.reflect.GeneratedMethodAccessor865.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.google.common.eventbus.Subscriber.invokeSubscriberMethod(Subscriber.java:87)
	at com.google.common.eventbus.Subscriber$1.run(Subscriber.java:72)
	at com.google.common.util.concurrent.MoreExecutors$DirectExecutor.execute(MoreExecutors.java:398)
	at com.google.common.eventbus.Subscriber.dispatchEvent(Subscriber.java:67)
	at com.google.common.eventbus.Dispatcher$ImmediateDispatcher.dispatch(Dispatcher.java:186)
	at com.google.common.eventbus.EventBusThatThrowsException.postWithException(EventBusThatThrowsException.java:65)
	at org.killbill.bus.DefaultPersistentBus.dispatchBusEventWithMetrics(DefaultPersistentBus.java:376)
	at org.killbill.bus.dispatching.BusCallableCallback.dispatch(BusCallableCallback.java:41)
	at org.killbill.bus.dispatching.BusCallableCallback.dispatch(BusCallableCallback.java:30)
	at org.killbill.queue.dispatching.Dispatcher$CallableQueueHandler.call(Dispatcher.java:137)
	at org.killbill.queue.dispatching.Dispatcher$CallableQueueHandler.call(Dispatcher.java:105)
	at org.killbill.commons.concurrent.WrappedCallable.call(WrappedCallable.java:42)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at org.killbill.commons.concurrent.WrappedRunnable.run(WrappedRunnable.java:48)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #1234, opened 2019-10-01
payment: perform AUTO_PAY_OFF check early when paying invoices

pierre:
This fixes https://github.com/killbill/killbill/issues/812, https://github.com/killbill/killbill/issues/1193 and https://github.com/killbill/killbill/issues/1211.

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
@@ -95,13 +95,17 @@ import org.killbill.billing.util.audit.AccountAuditLogs;
 import org.killbill.billing.util.audit.AuditLogWithHistory;
 import org.killbill.billing.util.callcontext.CallContext;
 import org.killbill.billing.util.callcontext.TenantContext;
+import org.killbill.billing.util.tag.ControlTagType;
+import org.killbill.billing.util.tag.Tag;
 import org.killbill.billing.util.userrequest.CompletionUserRequestBase;
 import org.killbill.clock.Clock;
 import org.killbill.commons.metrics.TimedResource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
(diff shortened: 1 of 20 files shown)


pierre:
@sbrossie Assuming CircleCi is green, this should be ready to be reviewed now.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -352,14 +347,27 @@ public Response createSubscriptionsWithAddOnsInternal(final List<BulkSubscriptio
...
+
+            private boolean isImmediateOp = true;
+
             @Override
             public List<UUID> doOperation(final CallContext ctx) throws EntitlementApiException {
+                for (final BaseEntitlementWithAddOnsSpecifier spec: baseEntitlementWithAddOnsSpecifierList) {
+                    if (spec.getBillingEffectiveDate() != null) {
+                        final LocalDate nowInAccountTimeZone = new LocalDate(callContext.getCreatedDate(), spec.getBillingEffectiveDate().getChronology().getZone());
+                        if (spec.getBillingEffectiveDate().isAfter(nowInAccountTimeZone)) {
+                            isImmediateOp = false;
```
sbrossie: I am questioning whether this logic will lead to the right expectation:
* `isImmediateOp` = `false` => this was obtained by checking that `spec#getBillingEffectiveDate` is not null and is  in the future, all good.
* `isImmediateOp` = `true` => obtained as a default for when prev condition is not met. However, prev condition only checked that `spec#getBillingEffectiveDate` is not null and is  in the future, but we could still have `spec.getBillingEffectiveDate().is**Not**After(nowInAccountTimeZone)` and yet have a date that is in the future ?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -351,15 +352,38 @@ public Response createSubscriptionsWithAddOnsInternal(final List<BulkSubscriptio
             baseEntitlementWithAddOnsSpecifierList.add(baseEntitlementSpecifierWithAddOns);
         }
 
+
+        // Retrieve the tags for the ACCOUNT object to correctly implement callCompletion in the simple use-cases.
+        // In reality, this is much more complex though and not all scenarii are supported (e.g. entitlement plugin could add some tags on the fly).
+        final List<Tag> accountTags = tagUserApi.getTagsForAccountType(account.getId(), ObjectType.ACCOUNT, false, callContext);
```
sbrossie: Should we only fetch those when `callCompletion` = `true` ?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
pierre: Good catch! See 60f3dfe3028ba99abc4859c33380042102e1e7cb.

The only downside is that I had to re-use some of the logic internal to `InternalCallContextFactory` inside jaxrs.

PR #1244, opened 2019-11-12
Work for 1134

sbrossie:


invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -207,27 +207,29 @@ public boolean apply(final InvoiceModelDao invoice) {
...
                                                                                                                                  @Override
                                                                                                                                  public boolean apply(final InvoiceModelDao invoice) {
-                                                                                                                                     return !invoice.isMigrated() && invoice.getTargetDate().compareTo(fromDate) >= 0 &&
+                                                                                                                                     return !invoice.isMigrated() &&
+                                                                                                                                            (fromDate == null || invoice.getTargetDate().compareTo(fromDate) >= 0) &&
+                                                                                                                                            (upToDate == null || invoice.getTargetDate().compareTo(upToDate) <= 0) &&
```
pierre: Worth adding the comment from the commit message directly in the code, so we remember without having to `git blame`...

sbrossie: Updated our api doc to reflect the [change of semantics](https://github.com/killbill/killbill-api/pull/72)

Issue #1208, opened 2019-09-12
Wrong argument in catalog API client


pierre:
`catalogApi.getCatalogXml` takes a `LocalDate`. It should be a `DateTime` instead.

sbrossie:
This seems like the same issue from #922, and we need to enhance the [ugly fix](https://github.com/killbill/killbill/commit/c6510a6615de13206af7b591b3fcba3359f02318) to include the `catalogApi.getCatalogXml`.

PR #1245, opened 2019-11-12
jaxrs: Fix CatalogApi#getCatalogXml to take a DateTime. See #1208

sbrossie:


sbrossie:
@pierre Merging, so i can regenerate and release the client

pierre:
👍

PR #1249, opened 2019-12-10
config: reduce default number of threads

pierre:


util/src/main/java/org/killbill/billing/util/config/definition/JaxrsConfig.java
```diff
@@ -26,7 +26,7 @@
 public interface JaxrsConfig extends KillbillConfig {
 
     @Config("org.killbill.jaxrs.threads.pool.nb")
-    @Default("30")
+    @Default("3")
```
sbrossie: Maybe `10` as well ?

pierre: Fixed in 361563e36e236ab9e1a082549e84c24778322683.

PR #1255, opened 2020-01-13
Fix for invoice tree

sbrossie:


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithBCDUpdate.java
```diff
@@ -665,7 +665,7 @@ public void testBCDChangeFromFreePlanToPayingPlanWithTrialAndCHANGE_OF_PLANPolic
         // Change to the paying plan (alignment is CHANGE_OF_PLAN: we end up in TRIAL)
         // Extra NULL_INVOICE event because invoice computes a future notification effective right away
```
pierre: Comment to update

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
```diff
@@ -213,12 +213,15 @@ boolean isSameDayAndSameSubscription(final InvoiceItem prevComputedFixedItem, fi
...
                                                     final InvoiceItemGeneratorLogger invoiceItemGeneratorLogger,
-                                                    final Map<UUID, SubscriptionFutureNotificationDates> perSubscriptionFutureNotificationDate,
+                                                    final Map<UUID, SubscriptionFutureNotificationDates> perSubscriptionFutureNotificationDates,
                                                     final InternalCallContext internalCallContext) throws InvoiceApiException {
 
         try {
             final List<InvoiceItem> items = new ArrayList<InvoiceItem>();
-
+            final LocalDate thisEventEffectiveDate = internalCallContext.toLocalDate(thisEvent.getEffectiveDate());
+            if (thisEventEffectiveDate.compareTo(targetDate) > 0) {
```
pierre: Is this an optimization (unrelated to the bug)?

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
```diff
@@ -272,11 +275,16 @@ boolean isSameDayAndSameSubscription(final InvoiceItem prevComputedFixedItem, fi
                         }
                     }
                     updatePerSubscriptionNextNotificationDate(thisEvent.getSubscriptionId(), itemDataWithNextBillingCycleDate.getNextBillingCycleDate(), items, recurringBillingMode,
-                                                              perSubscriptionFutureNotificationDate);
+                                                              perSubscriptionFutureNotificationDates);
+                }
+            } else {
+                final SubscriptionFutureNotificationDates futureNotificationDates = perSubscriptionFutureNotificationDates.get(thisEvent.getSubscriptionId());
```
pierre: Is my understanding correct that prior to this change, a subscription with a blocked billing period might have had a (NULL) invoice run after the paused date?

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
sbrossie: No, it's part of the big fix: You could have future events in the billing events (past `targetDate`). For instance if you have a future cancelation you want to exit quickly otherwise this will clear the `perSubscriptionFutureNotificationDates` map.

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
sbrossie: I think this is true. I ''ll verify the state of our pause/resume tests for `in-advance` and add one for `in-arrear`.

PR #1258, opened 2020-01-15
jaxrs: pass kbVersion to KPM bundle

pierre:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/NodesInfoResource.java
```diff
@@ -156,7 +156,12 @@ public Response triggerNodeCommand(final NodeCommandJson json,
                                        @HeaderParam(HDR_REASON) final String reason,
                                        @HeaderParam(HDR_COMMENT) final String comment,
                                        @javax.ws.rs.core.Context final HttpServletRequest request,
-                                       @javax.ws.rs.core.Context final UriInfo uriInfo) throws AccountApiException {
+                                       @javax.ws.rs.core.Context final UriInfo uriInfo) {
+        // Add in the broadcast message the Kill Bill version -- this is needed by the KPM plugin for instance
+        final NodeInfo currentNodeInfo = killbillInfoApi.getCurrentNodeInfo();
+        if (currentNodeInfo != null) {
+            json.getNodeCommandProperties().add(new NodeCommandPropertyJson("kbVersion", currentNodeInfo.getKillbillVersion()));
```
sbrossie: Should we check if such key was not already passed?

pierre: I thought about it but the use-case would be to force KPM to see Kill Bill running a different version than the reality. I've found it a bit odd... Easy to change later on if we really need it though.

PR #1265, opened 2020-01-30
Fix for 1264

sbrossie:


payment/src/main/java/org/killbill/billing/payment/retry/BaseRetryService.java
```diff
@@ -91,6 +94,30 @@ public void stop() throws NoSuchNotificationQueue {
         }
     }
 
+    public List<NotificationEventWithMetadata<PaymentRetryNotificationKey>> getRetryAttempts(final UUID attemptId, final InternalCallContext context) {
```
pierre: If this is just for testing, should we instead implement this in our tests?

payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentTransactionTask.java
```diff
@@ -127,8 +127,8 @@ TransactionStatus updatePaymentAndTransactionIfNeeded(final UUID accountId,
                                                           final InternalTenantContext internalTenantContext) throws LockFailedException {
         // In the GET case, make sure we bail as early as possible (see PaymentRefresher)
         if (currentTransactionStatus != null && !TRANSACTION_STATUSES_TO_CONSIDER.contains(currentTransactionStatus)) {
-            // Nothing to do
-            return null;
```
pierre: Shouldn't this be changed too?

https://github.com/killbill/killbill/blob/ce2ace3039250f4c87ac2afa16a51a498c045360/payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentTransactionTask.java#L159

payment/src/main/java/org/killbill/billing/payment/retry/BaseRetryService.java
sbrossie: sure, i 'll remove it altogether as my test actually does not really need it since we already return the pending attempts in the Payment#attempts

payment/src/main/java/org/killbill/billing/payment/core/janitor/IncompletePaymentTransactionTask.java
sbrossie: Yes, i think so too, will fix.

PR #1274, opened 2020-02-05
invoice, payment: Add paymentAttemptId into the INVOICE_PAYMENT_SUCCE…

sbrossie:
…SS json (InvoicePaymentMetadata)


payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
```diff
@@ -642,6 +652,7 @@ public boolean apply(final PaymentTransactionModelDao input) {
                                                           successfulTransaction.getCurrency(),
                                                           successfulTransaction.getProcessedCurrency(),
                                                           successfulTransaction.getPaymentId(),
+                                                          incompleteInvoicePayment.getId(),
```
pierre: Is that correct?

sbrossie: Good catch - fixed in https://github.com/killbill/killbill/pull/1274/commits/c9a71e60bb54318b67cb4cfb570f80592240977f. My test is just checking that value is not null, but i did verify this time this is correct.

Issue #1275, opened 2020-02-06
Unrecognized reported usage units ignored silently


pierre:
If usage data is reported for units that aren't present in the catalog, these aren't invoiced and ignored silently. These could happen for instance if a third party provider is responsible for reporting these units, but is not in control of the catalog configuration.

This can also be confusing if subsequent versions of the catalog know about these unit types: depending on which catalog version a given subscription is on (e.g. grandfathering), some of these units can be invoices while some are ignored.

Could Kill Bill be more defensive about this?

* Validating upon usage reporting might be hard, as the usage module would have to pull the catalog versions
* Another idea could be to park the account when invoicing detects such inconsistencies (we might have to introduce a config property for backward compatibility)

Issue #1273, opened 2020-02-05
Performance issue in getInvoiceTags


pierre:
See https://github.com/killbill/killbill/blob/b74f63d7e6e085155a266fc4f2cbad4db4ed892e/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L959

We shouldn't pull the full invoice object to retrieve the account id.

PR #1278, opened 2020-02-07
0.20 maintenance fixes

pierre:
Fixes for https://github.com/killbill/killbill/issues/1273 and https://github.com/killbill/killbill/issues/1275.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
@@ -176,13 +176,13 @@ public class SubscriptionUsageInArrear {
                                        new ContiguousIntervalConsumableUsageInArrear(usage, accountId, invoiceId, rawSubscriptionUsage, existingTrackingIds, targetDate, rawUsageStartDate, usageDetailMode, invoiceConfig, internalTenantContext);
 
                     inFlightInArrearUsageIntervals.put(usageKey, existingInterval);
-                    allSeenUnitTypesForBillingEvent.addAll(existingInterval.getUnitTypes());
                 }
                 // Add billing event for that usage interval
                 existingInterval.addBillingEvent(event);
                 // Remove usage interval for toBeClosed set
                 toBeClosed.remove(usageKey);
 
+                allSeenUnitTypesForBillingEvent.addAll(existingInterval.getUnitTypes());
                 contiguousIntervalsUsageInArrear.add(existingInterval);
             }
 
@@ -196,6 +196,7 @@ public class SubscriptionUsageInArrear {
                 final ContiguousIntervalUsageInArrear interval = inFlightInArrearUsageIntervals.remove(usageKey);
                 if (interval != null) {
                     interval.addBillingEvent(event);
+                    interval.addAllSeenUnitTypesForBillingEvent(event, allSeenUnitTypesForBillingEvent);
                     usageIntervals.add(interval.build(true));
                 }
             }
--- a/invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
@@ -29,7 +29,6 @@ import java.util.UUID;
 
 import org.joda.time.DateTime;
 import org.joda.time.LocalDate;
-import org.killbill.billing.ErrorCode;
 import org.killbill.billing.callcontext.InternalTenantContext;
 import org.killbill.billing.catalog.api.BillingMode;
 import org.killbill.billing.catalog.api.CatalogApiException;
@@ -148,9 +147,6 @@ public class SubscriptionUsageInArrear {
 
         final Set<UsageKey> allSeenUsage = new HashSet<UsageKey>();
 
-        // Will contain all unit types that each ContiguousIntervalUsageInArrear has looked at, as defined in the catalog
-        final Set<String> allSeenUnitTypes = new HashSet<String>();
-
         for (final BillingEvent event : subscriptionBillingEvents) {
             // Extract all in arrear /consumable usage section for that billing event.
             final List<Usage> usages = findUsageInArrearUsages(event);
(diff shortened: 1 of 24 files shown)


invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
```diff
@@ -164,6 +176,7 @@ public UsageKey apply(final Usage input) {
                                        new ContiguousIntervalConsumableUsageInArrear(usage, accountId, invoiceId, rawSubscriptionUsage, existingTrackingIds, targetDate, rawUsageStartDate, usageDetailMode, internalTenantContext);
 
                     inFlightInArrearUsageIntervals.put(usageKey, existingInterval);
+                    allSeenUnitTypes.addAll(existingInterval.getUnitTypes());
```
sbrossie: Shouldn't this be reset for each billing event though?

Let's say we have 2 billing events with different usage types:
*` efftDt1`: b1 {t1, t2}
* `efftDt2`: b2 {t2}

->  If we record usage type t1, after `efftDt2` -> park?

invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
```diff
@@ -180,6 +193,21 @@ public UsageKey apply(final Usage input) {
                 }
             }
         }
+
+        // Sanity check: https://github.com/killbill/killbill/issues/1275
+        for (final RawUsage rawUsage : rawSubscriptionUsage) {
```
sbrossie: This code could probably be moved inside `ContiguousIntervalUsageInArrear#getRolledUpUsage` where we already iterate across all `rawSubscriptionUsage`  elements?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
```diff
@@ -999,8 +1000,9 @@ public Response getTags(@PathParam(ID_PARAM_NAME) final UUID invoiceId,
                             @QueryParam(QUERY_AUDIT) @DefaultValue("NONE") final AuditMode auditMode,
                             @javax.ws.rs.core.Context final HttpServletRequest request) throws TagDefinitionApiException, InvoiceApiException {
         final TenantContext tenantContext = context.createTenantContextNoAccountId(request);
-        final Invoice invoice = invoiceApi.getInvoice(invoiceId, tenantContext);
-        return super.getTags(invoice.getAccountId(), invoiceId, auditMode, includedDeleted, tenantContext);
+        // See https://github.com/killbill/killbill/issues/1273
+        final UUID accountId = AuditLevel.NONE.equals(auditMode.getLevel()) ? null : invoiceApi.getInvoice(invoiceId, tenantContext).getAccountId();
+        return super.getTags(accountId, invoiceId, auditMode, includedDeleted, tenantContext);
```
sbrossie: Yeah, the change makes sense from a perf point of view - handling this Nullable `accountId` is not great though, but what to do...

invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
pierre: I agree there is a bug in the scenario you mentioned above (I was able to reproduce it in a test case), but I'm struggle coming up with a fix.

First, to move the logic into `getRolledUpUsage`, this logic has to be moved to: https://github.com/killbill/killbill/blob/19b338ffba1e8998ff80ac7d894a0f0672e7fe80/invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java#L435-L443

If I do so though, as I'm iterating through the usage data in `getRolledUpUsage`, it's hard the filter the recorded usage by unit type as it's unclear if the unknown recorded data is either not defined in the catalog, or simply for another usage section (one billing event maps to several `ContiguousIntervalUsageInArrear`).

The only solution I can find would be to move my logic inside the `for` loop of `computeInArrearUsageInterval` (meaning I would need to expose `getRolledUpUsage` publicly and call it there to filter the usage by date).

Might be worth for you refreshing your memory on this and talk through it in more details, in case I'm missing something...

sbrossie: I looked at it, and i think one issue is that code from `0.20` and `0.22` has diverged quite a bit: In particular, i had added a mechanism to link a `transitionTime` with its `billingEvent`: `https://github.com/killbill/killbill/commit/48681d8821db97a0952482e6abf8a308dea2f918#diff-3e43568d6c03fc887b25fd5883cb0c3aR99`

So, what i had in mind was:
1. Compute the `allSeenUnitTypes` per `billingEvent` (similar to `toBeClosed`)
2. Pass the `allSeenUnitTypes`  to `interval.addBillingEvent(event)` and leverage the `TransitionTime` structure to also attach the `allSeenUnitTypes`

=> At this point in `ContiguousIntervalUsageInArrear` we have the mapping between a given 
`transition` -> `billingEvent` -> `allSeenUnitTypes` and we should be able as we go through the `RolledUpUsage` and check wether we see entries contains unit that are not part of the `allSeenUnitTypes`  (for that specific `billingEvent`)

But... this code is not available on `0.20` so that makes things a bit more complex...

pierre: > => At this point in `ContiguousIntervalUsageInArrear` we have the mapping between a given `transition` -> `billingEvent` -> `allSeenUnitTypes` and we should be able as we go through the `RolledUpUsage` and check wether we see entries contains unit that are not part of the `allSeenUnitTypes` (for that specific `billingEvent`)

This is not enough though because of the `filterInputRawUsage` I mentioned above: each interval knows about the billing events but not about other intervals (unknown unit type could be caused either by a missing entry in the catalog or because it is owned by another interval -- typically when you have both capacity and consumable for the same plan).

I tried a few approaches and finally settled on a mix of backport of your patch + removing the logic `filterInputRawUsage`. To be reviewed...

pierre:
Created https://github.com/killbill/killbill/pull/1279 with a cleaner history.

PR #1279, opened 2020-02-11
0.20 maintenance fixes

pierre:
Fixes for https://github.com/killbill/killbill/issues/1273 and https://github.com/killbill/killbill/issues/1275.

See https://github.com/killbill/killbill/pull/1278.

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestUnknownUsageUnits.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestUnknownUsageUnits.java
@@ -217,4 +217,83 @@ public class TestUnknownUsageUnits extends TestIntegrationBase {
                                                  new ExpectedInvoiceItemCheck(new LocalDate(2012, 8, 2), new LocalDate(2012, 9, 1), InvoiceItemType.USAGE, new BigDecimal("0")));
         invoiceChecker.checkTrackingIds(curInvoice, ImmutableSet.of("tracking-11"), internalCallContext);
     }
+
+    @Test(groups = "slow", description = "https://github.com/killbill/killbill/pull/1279")
+    public void testWithUnknownUsageAndSeveralPeriods() throws Exception {
+        // We take april as it has 30 days (easier to play with BCD)
+        // Set clock to the initial start date - we implicitly assume here that the account timezone is UTC
+        clock.setDay(new LocalDate(2012, 4, 1));
+
+        final AccountData accountData = getAccountData(1);
+        final Account account = createAccountWithNonOsgiPaymentMethod(accountData);
+        accountChecker.checkAccount(account.getId(), accountData, callContext);
+
+        Assert.assertFalse(parkedAccountsManager.isParked(internalCallContext));
+
+        // Enable strict mode
+        invoiceConfig.setShouldParkAccountsWithUnknownUsage(true);
+
+        // Catalog-v1.xml
+        final DefaultEntitlement bpSubscription = createBaseEntitlementAndCheckForCompletion(account.getId(), "bundleKey", "Server", ProductCategory.BASE, BillingPeriod.MONTHLY, NextEvent.CREATE, NextEvent.BLOCK, NextEvent.NULL_INVOICE);
+        // Check bundle after BP got created otherwise we get an error from auditApi.
+        subscriptionChecker.checkSubscriptionCreated(bpSubscription.getId(), internalCallContext);
+
+        // Record known usage for April
+        recordUsageData(bpSubscription.getId(), "tracking-1", "server-hourly-type-1", new LocalDate(2012, 4, 1), 99L, callContext);
+        recordUsageData(bpSubscription.getId(), "tracking-2", "bandwidth-type-1", new LocalDate(2012, 4, 15), 100L, callContext);
+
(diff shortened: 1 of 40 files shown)


invoice/src/main/java/org/killbill/billing/invoice/usage/SubscriptionUsageInArrear.java
```diff
@@ -160,26 +174,39 @@ public UsageKey apply(final Usage input) {
...
+                contiguousIntervalUsageInArrear.addAllSeenUnitTypesForBillingEvent(event, allSeenUnitTypesForBillingEvent);
             }
 
             // Build the usage interval that are no longer referenced
             for (final UsageKey usageKey : toBeClosed) {
                 final ContiguousIntervalUsageInArrear interval = inFlightInArrearUsageIntervals.remove(usageKey);
                 if (interval != null) {
                     interval.addBillingEvent(event);
+                    // No usage section for CANCEL events
+                    interval.addAllSeenUnitTypesForBillingEvent(event, usages.isEmpty() ? allSeenUnitTypesForPrevBillingEvent : allSeenUnitTypesForBillingEvent);
```
sbrossie: Confused about that one.. I would expect that we pass an empty set for any cancelation, ... I would also expect that for such events `usages` is empty (all the time)

beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestUnknownUsageUnits.java
```diff
@@ -0,0 +1,188 @@
...
+    @Override
+    protected KillbillConfigSource getConfigSource(final Map<String, String> extraProperties) {
+        final Map<String, String> allExtraProperties = new HashMap<String, String>(extraProperties);
+        allExtraProperties.putAll(DEFAULT_BEATRIX_PROPERTIES);
+        allExtraProperties.put("org.killbill.catalog.uri", "catalogs/testUnknownUsageUnits");
+        return getConfigSource(null, allExtraProperties);
+    }
+
+    @Test(groups = "slow", description = "https://github.com/killbill/killbill/issues/1275")
+    public void testWithUnknownUsage() throws Exception {
```
sbrossie: Looks like this test relies on multiple catalog versions, it would make it (much) easier to point in the test which catalog version is used at a certain time (and possibly valid units)

pierre:
@sbrossie Updated as discussed (FYI it turns out I was already testing the case where usage is deleted in my new test).

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
```diff
@@ -81,7 +82,8 @@
 
     protected final List<TransitionTime> transitionTimes;
     protected final List<BillingEvent> billingEvents;
-    protected final Map<BillingEvent, Set<String>> allSeenUnitTypes;
+    // Ordering is important here!
+    protected final LinkedHashMap<BillingEvent, Set<String>> allSeenUnitTypes;
```
sbrossie: Do we really need this ordering? It seems all you care about is the ability to find the last billing event in the method `addAllSeenUnitTypesFromPrevBillingEvent` ? Shouldn't such last billing event be retrievable from `billingEvents` ?

sbrossie:
@pierre I think we are close enough, that if you want to start looking at the merge, this should be fine...

pierre:
> i have been thinking about this last billing event `CANCEL` and why there is a dissymmetry...

It's not so much that the `CANCEL` event is dissymmetric (my original comment was wrong to point it out explicitly), it's more that the code is dissymmetric:

* in the branch where `existingInterval == null`, we are creating a new interval because we discovered a new usage section. At this point, we start tracking the unit types seen.
* in the `for` loop `final UsageKey usageKey : toBeClosed`, we are closing intervals which aren't referenced anymore by this billing event. This could be because of a `CANCEL` event (no more usage), or because we have a catalog version change (even if the usage hasn't changed, a new interval is created as `UsageKey` takes into account the catalog effective date). In either case, we need to look at the usage as defined by the previous billing event.

Does this make sense? If so, let me know if your concerned is addressed or if I should add a new test case to be safe (I didn't quite follow your example though 😬).

> @pierre I think we are close enough, that if you want to start looking at the merge, this should be fine...

I'm going to release 0.20.15 as to unblock folks impacted by this, but I'll keep this PR open until we finally converge 😄

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java
pierre: Yes, we could look it up that way. But in that case, the ordering here needs to be swapped: https://github.com/killbill/killbill/pull/1279/files#diff-050672ee3314d0f9a29f5065b853673bR200 (call `addAllSeenUnitTypesFromPrevBillingEvent` before `addBillingEvent`). This felt a bit more sneaky...

pierre:
@sbrossie I've added the following (passing) test: https://github.com/killbill/killbill/pull/1279/commits/f184f37d1ddbd68bf371f71a5d52ade7397d2c96. Let me know if this is aligned with what you had in mind.

PR #1280, opened 2020-02-11
Fix for 1276

sbrossie:


payment/src/main/java/org/killbill/billing/payment/core/sm/control/ControlPluginRunner.java
```diff
@@ -85,25 +85,27 @@ public PriorPaymentControlResult executePluginPriorCalls(final Account account,
...
+                                                                                                      paymentId,
+                                                                                                      paymentExternalKey,
+                                                                                                      paymentTransactionId,
+                                                                                                      paymentTransactionExternalKey,
+                                                                                                      paymentApiType,
+                                                                                                      transactionType,
+                                                                                                      hppType,
+                                                                                                      inputAmount,
+                                                                                                      inputCurrency,
+                                                                                                      null,
```
pierre: Why this change?

FYI at least `getProcessedCurrency` is used in the invoice control plugin prior call (`recordPaymentAttemptCompletion`) -- are we sure `null` doesn't have any side effect?

sbrossie: I don't see where `getProcessedCurrency` is used in the invoice control plugin prior call -- i think the code you are [pointing at](https://github.com/killbill/killbill/blob/master/payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java#L652) is using the values from the `successfulTransaction`, not the context?

Stepping back: the processed currency/amount is the one that is used after the payment was made, so it makes little sense to pass it to the control plugin in the `priorCall`, hence the change.

That being said, i realize that this change is somewhat orthogonal to what i wanted to change (and therefore could be removed), i was taking the opportunity to clean up things. I can change it back.

Issue #1288, opened 2020-04-02
NPE in /1.0/kb/invoices/:invoiceId/payments


pierre:
Aborted `payment_attempts` can have a corresponding row in `invoice_payments` with a `NULL` `payment_id` (note that the `payment_cookie_id` i.e. the `payment_attempts. transaction_external_key` won't be `NULL` though):

```
MySQL [killbill]> select * from invoice_payments order by record_id desc limit 1\G
*************************** 1. row ***************************
                record_id: 3
                       id: e7f6b1d0-3823-4c5b-9226-12327af64492
                     type: ATTEMPT
               invoice_id: 742a1a60-58f8-4ed2-b9db-b1816a741f44
               payment_id: NULL
             payment_date: 2012-05-02 00:37:59
                   amount: 0.000000000
                 currency: USD
       processed_currency: USD
        payment_cookie_id: bac67833-ac7c-458d-8e2a-659c0cdfb0b6
linked_invoice_payment_id: NULL
                  success: 0
               created_by: PaymentRequestProcessor
             created_date: 2012-05-02 00:37:59
        account_record_id: 1
         tenant_record_id: 0
1 row in set (0.00 sec)
```

This can trigger a NPE here: https://github.com/killbill/killbill/blob/1be24a4a144e80abcf45a5982deb7c1f2e22aa3b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L623

```
java.lang.NullPointerException: null
    at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:877)
    at com.google.common.collect.ImmutableSet$Builder.add(ImmutableSet.java:483)
    at com.google.common.collect.ImmutableSet.copyOf(ImmutableSet.java:248)
    at com.google.common.collect.ImmutableSet.copyOf(ImmutableSet.java:230)
    at org.killbill.billing.jaxrs.resources.InvoiceResource.getPaymentsForInvoice(InvoiceResource.java:620)
```

PR #1289, opened 2020-04-02
jaxrs: fix NPE in getPaymentsForInvoice

pierre:
Handle the case of invoice payments with a NULL payment id (i.e. when the payment is aborted by a payment control plugin). These won't be returned by the API.

For testing, I had to add a new (undocumented) query parameter to the createInstantPayment API, in order to add an additional payment control plugin to the chain.

This fixes https://github.com/killbill/killbill/issues/1288.

### DIFF
--- a/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPayment.java
+++ b/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPayment.java
@@ -1,6 +1,6 @@
 /*
- * Copyright 2014-2015 Groupon, Inc
- * Copyright 2014-2015 The Billing Project, LLC
+ * Copyright 2014-2020 Groupon, Inc
+ * Copyright 2014-2020 The Billing Project, LLC
  *
  * The Billing Project licenses this file to you under the Apache License, version 2.0
  * (the "License"); you may not use this file except in compliance with the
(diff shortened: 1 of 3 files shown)


profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestPayment.java
```diff
@@ -219,6 +219,50 @@ public void testWithTimeoutPayment() throws Exception {
...
+        final InvoicePayment invoicePaymentNull = invoiceApi.createInstantPayment(failedInvoice.getInvoiceId(), invoicePayment, true, null, inputOptions);
+        Assert.assertNull(invoicePaymentNull);
+
+        // Verify new state
+        final Payments updatedPayments = accountApi.getPaymentsForAccount(account.getAccountId(), true, false, ImmutableMap.<String, String>of(), AuditLevel.NONE, requestOptions);
+        Assert.assertEquals(updatedPayments.size(), 1);
+        Assert.assertEquals(updatedPayments.get(0).getPaymentId(), initialPayments.get(0).getPaymentId());
+        final InvoicePayments updatedInvoicePayments = invoiceApi.getPaymentsForInvoice(failedInvoice.getInvoiceId(), requestOptions);
+        Assert.assertEquals(updatedInvoicePayments.size(), 1);
+        Assert.assertEquals(updatedInvoicePayments.get(0).getPaymentId(), updatedPayments.get(0).getPaymentId());
```
sbrossie, line 263: If I understand correctly this returns the first payment that failed (with a pending retry), and the one made manually that got aborted is correctly discarded from the list -- and before it was causing the NPE ?

sbrossie, line 263: If I understand correctly this returns the first payment that failed (with a pending retry), and the one made manually that got aborted is correctly discarded from the list -- and before it was causing the NPE ?

pierre: Yup!

PR #1290, opened 2020-04-07
invoice: clarify behavior for skipping overlapping usage items

pierre:
* Revisit log line to make it clearer and easier to debug
* Adjust test to highlight system behavior

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
@@ -477,7 +477,7 @@ public class TestConsumableInArrear extends TestIntegrationBase {
         final DefaultEntitlement aoSubscription = addAOEntitlementAndCheckForCompletion(bpSubscription.getBundleId(), "Bullets", ProductCategory.ADD_ON, BillingPeriod.NO_BILLING_PERIOD, NextEvent.CREATE, NextEvent.BLOCK, NextEvent.NULL_INVOICE);
 
         recordUsageData(aoSubscription.getId(), "tracking-1", "bullets", new LocalDate(2012, 4, 1), 99L, callContext);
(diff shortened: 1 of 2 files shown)


beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestConsumableInArrear.java
```diff
@@ -477,7 +477,7 @@ public void testWithPartialBlockBilling() throws Exception {
         final DefaultEntitlement aoSubscription = addAOEntitlementAndCheckForCompletion(bpSubscription.getBundleId(), "Bullets", ProductCategory.ADD_ON, BillingPeriod.NO_BILLING_PERIOD, NextEvent.CREATE, NextEvent.BLOCK, NextEvent.NULL_INVOICE);
 
         recordUsageData(aoSubscription.getId(), "tracking-1", "bullets", new LocalDate(2012, 4, 1), 99L, callContext);
-        recordUsageData(aoSubscription.getId(), "tracking-2", "bullets", new LocalDate(2012, 4, 15), 100L, callContext);
+        recordUsageData(aoSubscription.getId(), "tracking-2", "bullets", new LocalDate(2012, 4, 18), 100L, callContext);
```
sbrossie, line 480: Any reason for changing the date from `15` -> `18` ?

Oh I see, you are changing this test, to make sure  some of the usage points are past the effective blocking date. However, the `Ignoring usage...` happens regardless of this date, so I suppose the goal is check that such usage data would not be repaired -- which is true as we never repair usage?

pierre: > Oh I see, you are changing this test, to make sure some of the usage points are past the effective blocking date. 

Correct.

> However, the `Ignoring usage...` happens regardless of this date, so I suppose the goal is check that such usage data would not be repaired -- which is true as we never repair usage?

Yes: I wanted a test to highlight this behavior.

Note also that the content of `ru` is also very different when hitting the `Ignoring usage...` line:

* Before: 1->16, with 199 units
* After: 1->16, with 99 units

This further highlights that the code is looking at dates only.

Issue #277, opened 2015-02-25
Ability to override paymentMethod per subscription/bundle


sbrossie:
Today, when an invoice needs to be paid, the payment code extracts the default payment method associated with the account. An invoice contains potentially several items, each of which can map to a specific bundle/subscription. A Payment matches a specific invoice. Associating a payment method with a specific subscription or bundle will only work if the invoice associated with the payment contains only one subscription/bundle (or they all share the same overridden payment method).

Fortunately we have a mechanism in catalog to group certain subscription/bundle into one invoice. This is called `billingAlignment`, and we support 3 cases:
* ACCOUNT (all subscriptions will appear on one invoice)
* BUNDLE (all subscriptions for a given bundle will appear on one invoice)
* SUBSCRIPTION (only a specific subscription will appear on one given invoice)

 In order to override that behavior and override payment method per subscription/bundle, there are several possible designs:

1. Extend the current `InvoicePaymentRoutingPluginApi` to implement that functionality. The `InvoicePaymentRoutingPluginApi` implements the `PaymentRoutingPluginApi`, and that api is called from `RetryOperationCallback` to potentially break the flow or override values such as amount to be paid, currency to use, but also payment method that should be be used.

So, that code could be extended in the following way:
* Add a new table (state) to capture the paymentMethodId associated to a given bundle/subscription
* Export a new endpoint to register the paymentMethodId associated to a given bundle/subscription
* Have the `InvoicePaymentRoutingPluginApi`  check whether the payment method should be overridden by looking in the new table.

The issue is that this code (`InvoicePaymentRoutingPluginApi`) has not been extracted as a first class citizen plugin and is still part of Kill Bill core.

2. Modification of the previous proposal but in a custom plugin

The system property `org.killbill.payment.invoice.plugin` allows to specify an order of `PaymentRoutingPluginApi` that will be called by `RetryOperationCallback`  in the `priorCall` method.
This one will now need to contain "__INVOICE_PAYMENT_CONTROL_PLUGIN__, <name_of_new_plugin>", in that order.

The invoiceId that will be needed can be extracted using the property `PROP_IPCD_INVOICE_ID`. See for instance how this is done in `InvoicePaymentRoutingPluginApi`:
`final PluginProperty invoiceProp = getPluginProperty(paymentRoutingContext.getPluginProperties(), PROP_IPCD_INVOICE_ID);`

chris13524:
Any news on this? This would be a great feature to have.

pierre:
[...]

PR #1311, opened 2020-06-05
Invoice grouping feature

pierre:
Prepare ☕ and 🍫... 😄


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -572,90 +607,154 @@ private InvoiceWithFutureNotifications processAccountWithLockAndInputTargetDate(
...
             }
 
-            if (!isDryRun) {
-                // Compute whether this is a new invoice object (or just some adjustments on an existing invoice), and extract invoiceIds for later use
-                final Set<UUID> uniqueInvoiceIds = getUniqueInvoiceIds(invoice);
-                final boolean isRealInvoiceWithItems = uniqueInvoiceIds.remove(invoice.getId());
-                final Set<UUID> adjustedUniqueOtherInvoiceId = uniqueInvoiceIds;
-
-                logInvoiceWithItems(account, invoice, actualTargetDate, adjustedUniqueOtherInvoiceId, isRealInvoiceWithItems);
+            // For non dry-run scenarii, the CBA will be recomputed in the DAO layer (https://github.com/killbill/killbill/issues/637)
```
sbrossie, line 660: ... unless description has changed and then DAO would recompute the missing part (e.g TAX) as needed.

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -572,90 +607,154 @@ private InvoiceWithFutureNotifications processAccountWithLockAndInputTargetDate(
...
+                    invoiceModelDaos.add(invoiceModelDao);
+                }
 
+                if (invoiceModelDaos.isEmpty()) {
+                    setFutureNotifications(account, futureAccountNotifications, internalCallContext);
+                } else {
+                    // Note: we're demultiplexing the tracking ids here (same tracking ids for all invoices generated), as we don't
+                    // really know how the plugin(s) did the grouping. If multiple invoice items were generated for a a given subscription
+                    // for instance (e.g. catch-up scenario), they might now end up on different invoices.
+                    final Set<InvoiceTrackingModelDao> trackingIds = new HashSet<InvoiceTrackingModelDao>();
```
sbrossie, line 690: That part bothers me a little bit though... 

At the very least, I think we will need to change the `TrackingRecordId#equals` and `TrackingRecordId#hashCode` to not take into account the `invoiceId` (so essentially we could reuse the already existing `TrackingRecordId#isSimilarRecord`) so as to not have duplicate in our set [here](https://github.com/killbill/killbill/blob/killbill-0.22.9/invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalUsageInArrear.java#L499)

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
```diff
@@ -160,54 +163,84 @@ private void onCompletionCall(final boolean isSuccess,
...
+            }
 
-    private boolean updateOriginalInvoiceWithPluginInvoiceItems(final DefaultInvoice originalInvoice, final Collection<InvoiceItem> additionalInvoiceItems) {
-        if (additionalInvoiceItems.isEmpty()) {
-            return false;
+            // Note that here, the clonedInvoice is just used as a container object, it doesn't actually represent the final invoice.
+            // For instance, the plugin could return an ITEM_ADJ pointing to an existing invoice: the clonedInvoice object would
+            // not include the items from the original invoice. This doesn't matter anyways, as the dao will create the appropriate entries
+            // and we will re-hydrate the objects before returning (at least in the non dry-run path, see comments in processAccountWithLockAndInputTargetDate).
+            final DefaultInvoice clonedInvoice = (DefaultInvoice) originalInvoice.clone();
```
sbrossie, line 211: Do we really need to clone -- seems like we remove pretty much everything?

invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java
```diff
@@ -269,23 +269,32 @@ public Invoice getInvoiceByInvoiceItem(final UUID invoiceItemId, final TenantCon
...
-        final Invoice result = dispatcher.processAccount(true, accountId, targetDate, dryRunArguments, false, internalContext);
-        if (result == null) {
+        final Iterable<Invoice> results = dispatcher.processAccount(true, accountId, targetDate, dryRunArguments, false, internalContext);
+        if (results == null) {
             throw new InvoiceApiException(ErrorCode.INVOICE_NOTHING_TO_DO, accountId, targetDate != null ? targetDate : "null");
         } else {
-            return result;
+            if (Iterables.<Invoice>size(results) != 1) {
+                // If we change the API one day, we need to fix the implementation (see comments in processAccountWithLockAndInputTargetDate)
+                throw new UnsupportedOperationException("Invoices were generated but this version of Kill Bill doesn't support returning grouped invoices");
```
sbrossie, line 278: Could we instead return one Invoice with all item (non grouped). The `invoiceId` that is returned in dry-run is bogus anyways. We would lose the grouping but still have the correct set of items

invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java
```diff
@@ -269,23 +269,32 @@ public Invoice getInvoiceByInvoiceItem(final UUID invoiceItemId, final TenantCon
...
 
-        final Invoice result = dispatcher.processAccount(true, accountId, targetDate, null, false, internalContext);
-        if (result == null) {
+        final Iterable<Invoice> results = dispatcher.processAccount(true, accountId, targetDate, null, false, internalContext);
+        if (results == null) {
             throw new InvoiceApiException(ErrorCode.INVOICE_NOTHING_TO_DO, accountId, targetDate != null ? targetDate : "null");
         } else {
-            return result;
+            if (Iterables.<Invoice>size(results) != 1) {
+                throw new UnsupportedOperationException("Invoices were generated but this version of Kill Bill doesn't support returning grouped invoices");
```
sbrossie, line 294: This one is trickier: Perhaps we could return one invoice with a `null` `invoiceId` which means that several invoices were created as part of grouping feature. Unexposed functionality...

Doing this would allow to write a custom endpoint from plugin that would call this api and then re-apply its grouping logic?

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
pierre: Happy to improve the comment, but I'm not sure I follow your suggestion.

At this point, the temporary CBA item has been removed if needed (i.e. if the description wasn't changed). Regardless, we need to recompute the CBA. This is either done manually (dry-run scenario) or automatically in the transaction (DAO layer): this is what this comment refers to (and why there is a `if` around the CBA call).

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
pierre: I actually went back and forth on this :-) I ended up keeping the `clone()` call to hide the HA complexity (different constructor) but agreed we could go both ways.

invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java
pierre: I find it confusing: dry-run wouldn't reflect the non-dry-run case (grouping).

invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java
pierre: This might break Kaui though: the `Location` header wouldn't be able to be created.

Let's chat IRL about both use-cases, on one hand, I would prefer to be explicit. On the other, I agree that it's a huge limitation to block these calls.

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
pierre: Tracking ids are unique per subscription and record date. One thing I could do is add some amount of logic to parse the invoice items, find which tracking date falls within which service period, and dispatch these ids manually?

pierre:
@sbrossie FYI, I've pushed 6f7d3723f21a829cfac0f2e0a905df380320b7c3

pierre:
@sbrossie Merging into `emb-project` -- I'll open a new PR to address comments.

pierre:
@sbrossie Check the above in the context of https://github.com/killbill/killbill/issues/277.

PR #1318, opened 2020-07-02
jaxrs: rewrite BundleJson transformation

pierre:
Throwing a `RuntimeException` like it was written doesn't work well: the original exception (`CatalogApiException`) was not only masked, but it would also trigger a RST (because the exception was thrown during the Jackson serialization).


sbrossie:
@pierre I know we have used this pattern quite a bit, wondering if there are more instances to fix?

pierre:
> @pierre I know we have used this pattern quite a bit, wondering if there are more instances to fix?

I looked around, it seems we are good for JAX-RS.

PR #1327, opened 2020-07-18
Fix for plan changes that eventually specify an effectiveDateForExistingSubscriptions value

msqr:
This fix is to address the situation I find myself in, where I have many existing subscriptions and existing catalog versions, but only the most recent catalog version specifies an `effectiveDateForExistingSubscriptions` value. My expectation was that this would cause the plan to be applied to existing subscriptions starting on that date, but it was not. Upon investigation, I found a loop that stopped looking at catalog versions if no `effectiveDateForExistingSubscriptions` was specified. Thus it stopped looking too soon and never found the catalog version that _did_ have that date specified. After applying this fix, my invoices started picking up the most recent catalog prices, as expected.

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
@@ -624,7 +624,7 @@ public class DefaultSubscriptionBase extends EntityBase implements SubscriptionB
                         // Iterate through all more recent version of the catalog to find possible effectiveDateForExistingSubscriptions transition for this Plan
                         Plan nextPlan = catalog.getNextPlanVersion(currentPlan);
                         while (nextPlan != null ) {
-                            if ( nextPlan.getEffectiveDateForExistingSubscriptions() != null) {
+                            if (nextPlan.getEffectiveDateForExistingSubscriptions() != null) {
                                 final DateTime nextEffectiveDate = new DateTime(nextPlan.getEffectiveDateForExistingSubscriptions()).toDateTime(DateTimeZone.UTC);
                                 final PlanPhase nextPlanPhase = nextPlan.findPhase(planPhase.getName());
 
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
@@ -623,15 +623,17 @@ public class DefaultSubscriptionBase extends EntityBase implements SubscriptionB
 
                         // Iterate through all more recent version of the catalog to find possible effectiveDateForExistingSubscriptions transition for this Plan
                         Plan nextPlan = catalog.getNextPlanVersion(currentPlan);
-                        while (nextPlan != null && nextPlan.getEffectiveDateForExistingSubscriptions() != null) {
-                            final DateTime nextEffectiveDate = new DateTime(nextPlan.getEffectiveDateForExistingSubscriptions()).toDateTime(DateTimeZone.UTC);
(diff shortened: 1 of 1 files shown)


msqr:
My investigation started from an [old email thread](https://groups.google.com/g/killbilling-users/c/Hmc_CBl79og/m/-DhagicqAwAJ).

sbrossie:
@msqr Could you either write a test case -- or if too difficult, provide a working version of a catalog with the scenario so we can repro and write the test ourselves?

msqr:
Sorry I was trying to find how to write a good test case, and got as far as trying to write something like `org.killbill.billing.subscription.api.user.TestSubscriptionBillingEvents` with a different catalog, but am running out of time to devote to the effort. Here are 3 catalog versions I created to show what I mean, which is essential versions like:

1. Initial catalog
2. New catalog with price changes (no plan changes) and *no* `effectiveDateForExistingSubscriptions` set
3. New catalog with further price changes *and* `effectiveDateForExistingSubscriptions` set

The point being that the issue I faced was that catalog version 3 was never used, because this loop stopped looking for catalog changes after version 2 because of the missing `effectiveDateForExistingSubscriptions` value.

[catalog-versions.zip](https://github.com/killbill/killbill/files/4944497/catalog-versions.zip)

sbrossie:
@msqr I understand what your catalog is now, but I still have no idea about the scenario you ran and also your expectations:

1. When did you start the subscription?
2. What is the BCD associated with it?
3. What is the CTD (chargeThroughDate) -- how many invoices do you have?
3. When do you expect to see the price change -- and which action would trigger it?

msqr:
The subscription is started after catalog #1 and before catalog #2. The BCD is 1. There would be invoices already created up to the billing period just before catalog #3's effective date. I'd expect to see the price change for any invoice created after the `effectiveDateForExistingSubscriptions` date passes.

If you imagine at a higher level, I have subscribers that have been getting invoiced over a long period of time. There have been several catalog changes over that period of time, but `effectiveDateForExistingSubscriptions` was never set on the plan. At some point, I've changed the price in the usage tiers (in my case, they decrease) and uploaded a new catalog with those changes *and* a `effectiveDateForExistingSubscriptions` set. For usage collected after the point in time specified in `effectiveDateForExistingSubscriptions`, I want the subscribers to be charged the updated prices, whenever an invoice happens after that date.

sbrossie:
@msqr We have a [test case](https://github.com/killbill/killbill/blob/killbill-0.22.10/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogWithEffectiveDateForExistingSubscriptions.java#L149) which seems to test exactly what you have:
* [3 catalog versions](https://github.com/killbill/killbill/tree/killbill-0.22.10/beatrix/src/test/resources/catalogs/testCatalogWithEffectiveDateForExistingSubscriptions)
* Plan is usage based
* We test that `effectiveDateForSubscriptions` is taken into account as we move through the correct period.

I can certainly write another test, if you point out the difference -- if not, I am not sure what to write. I think an exact script would be useful. E.g Given catalog xxx,
* On date D1 create subscription
* On date D2 ...
* ...

=> Expect XXX

Ideally, I would ask you to write the test, but I realize our test framework needs practice, so I can take care of it, if the scenario is crystal clear.

msqr:
The difference is that your [catalog 2 specifies the `effectiveDateForExistingSubscriptions`](https://github.com/killbill/killbill/blob/killbill-0.22.10/beatrix/src/test/resources/catalogs/testCatalogWithEffectiveDateForExistingSubscriptions/WaterUtility-v2.xml#L61) while my catalog 2 does not. It is not until catalog 3 that `effectiveDateForExistingSubscriptions` appears.

sbrossie:
Was able to reproduce the issue, and the code fix looks really good 🎉 Merging into integration branch `fix-for-1329` for further testing.

PR #1330, opened 2020-07-21
Fix for 1329

sbrossie:


msqr:
Excellent, thanks for the test case!

PR #1331, opened 2020-07-22
Add action "ddl" to db-helper

aiham:
I needed to use the `create_ddl_file` function from db-helper without actually connecting to a database so I've added a `ddl` action to db-helper to expose it.


sbrossie:
@aiham Do we really need this flag. Can't you do: `db-helper -a dump > /tmp/killbill.ddl` ?

aiham:
You're right! Don't know how I missed the dump action. Thanks.

PR #1338, opened 2020-07-31
Performance fixes

pierre:


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -836,8 +836,15 @@ private void logInvoiceWithItems(final ImmutableAccountData account, final Invoi
             tmp.append(String.format("Adjusting existing invoiceId='%s', numberOfItems='%d', accountId='%s', targetDate='%s':\n",
                                      adjustedInvoices, invoice.getNumberOfItems(), account.getId(), targetDate));
         }
+        int n = 0;
         for (final InvoiceItem item : invoice.getInvoiceItems()) {
+            if (n > 50) {
```
sbrossie, line 841: Might be worth extracting as `private final const` and intuitively I would say a default less than 50. Perhaps 20?

invoice/src/main/java/org/killbill/billing/invoice/notification/DefaultNextBillingDatePoster.java
```diff
@@ -110,15 +109,32 @@ private void insertNextBillingFromTransactionInternal(final EntitySqlDaoWrapperF
...
             }
 
+            int n = 0;
+            final StringBuilder subscriptionIdsAsStringBuilder = new StringBuilder();
+            for (final UUID subscriptionId: subscriptionIds) {
+                if (n > 0) {
+                    subscriptionIdsAsStringBuilder.append(",");
+                }
+
+                if (n > 50) {
```
sbrossie, line 119: Same const to share ?

pierre:
> Share the `50` and 👍

I actually had not shared it on purpose as it's not obvious both should be equal. No strong feeling though, fixed in d659364e96c4c1f66f390ea334e5d45c1067c629.

PR #1358, opened 2020-09-16
Index on users.username should be unique

cab:
Without this, two requests can come in at the same time and result in duplicate users.

Discussed on mailing list [here](https://groups.google.com/g/killbilling-users/c/KBTGCF3_AI4).

### DIFF
--- a/util/src/main/resources/org/killbill/billing/util/migration/V20200916171757__recreate_users_index_as_unique.sql
+++ b/util/src/main/resources/org/killbill/billing/util/migration/V20200916171757__recreate_users_index_as_unique.sql
@@ -1,0 +1,2 @@
+drop index users_username on users;
+create unique index users_username ON users(username);
--- a/util/src/main/resources/org/killbill/billing/util/ddl.sql
+++ b/util/src/main/resources/org/killbill/billing/util/ddl.sql
@@ -265,7 +265,7 @@ CREATE TABLE users (
     updated_by varchar(50) DEFAULT NULL,
     PRIMARY KEY(record_id)
 ) /*! CHARACTER SET utf8 COLLATE utf8_bin */;
-CREATE INDEX users_username ON users(username);
+CREATE UNIQUE INDEX users_username ON users(username);
 
 
 DROP TABLE IF EXISTS user_roles;


sbrossie (changes requested):
Thanks for the PR.

The fix looks good but we will need also some migration files under `https://github.com/killbill/killbill/tree/master/util/src/main/resources/org/killbill/billing/util/migration`.

For info about creating these files, please see our [database migration guide](https://docs.killbill.io/latest/database_migrations.html)

util/src/main/resources/org/killbill/billing/util/migration/V20200916171757__recreate_users_index_as_unique.sql
```diff
@@ -0,0 +1,2 @@
+drop index users_username on users;
+create unique inex users_username ON users(username);
```
sbrossie, line 2: Typo: `inex`.

Could you run it, make sure it works?

sbrossie (changes requested):

util/src/main/resources/org/killbill/billing/util/migration/V20200916171757__recreate_users_index_as_unique.sql
cab: Apologies, I did run it and noticed the typo, but forgot had already committed. My bad.

```sql
> drop index users_username on users;
Query OK, 0 rows affected (0.009 sec)
Records: 0  Duplicates: 0  Warnings: 0

> create unique index users_username ON users(username);
Query OK, 0 rows affected (0.015 sec)
Records: 0  Duplicates: 0  Warnings: 0

> show index from users;
+-------+------------+----------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table | Non_unique | Key_name       | Seq_in_index | Column_name | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+-------+------------+----------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| users |          0 | PRIMARY        |            1 | record_id   | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
| users |          0 | record_id      |            1 | record_id   | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
| users |          0 | users_username |            1 | username    | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
+-------+------------+----------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
3 rows in set (0.001 sec)
```

sbrossie (approved):

sbrossie:
Tanks @cab 🎉

PR #1360, opened 2020-09-16
Fix issue when using AUTO_INVOICING_REUSE_DRAFT and with (subscriptio…

sbrossie:
…n) dates in the past.

The scenario is one where we have AUTO_INVOICING_DRAFT/AUTO_INVOICING_REUSE_DRAFT set on the account
and we create some subscriptions in the past.

There will be an initial bus event for the start date creating the invoice and adding the first RECURRING period (from test example: 8/7 -> 9/1)
Then, a notification be picked up immediately for the next RECURRING period as its effective date is in the past (9/1 -> 10/1)

Unfortunately prior the fix the targetDate which is computed as the max(event date, all invoices targetDate) is wrong because targetDate from (resused) invoice
was not updated.

This fix will now update the targetDate associated with the ongoing (reused) invoice and therefore avoid all the REPAIR we were seeing.

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationBase.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationBase.java
@@ -20,15 +20,18 @@ package org.killbill.billing.beatrix.integration;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Named;
 
+import org.awaitility.Awaitility;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
@@ -131,6 +134,10 @@ import org.killbill.billing.util.nodes.KillbillNodesApi;
 import org.killbill.billing.util.tag.ControlTagType;
 import org.killbill.bus.api.PersistentBus;
 import org.killbill.bus.api.PersistentBus.EventBusException;
+import org.killbill.notificationq.api.NotificationEvent;
+import org.killbill.notificationq.api.NotificationEventWithMetadata;
+import org.killbill.notificationq.api.NotificationQueue;
+import org.killbill.notificationq.api.NotificationQueueService;
 import org.skife.config.ConfigurationObjectFactory;
 import org.skife.config.TimeSpan;
 import org.skife.jdbi.v2.Handle;
@@ -318,6 +325,9 @@ public class TestIntegrationBase extends BeatrixTestSuiteWithEmbeddedDB implemen
 
     @Inject
     protected PaymentDao paymentDao;
+
+    @Inject
+    protected NotificationQueueService notificationQueueService;
 
     protected ConfigurableInvoiceConfig invoiceConfig;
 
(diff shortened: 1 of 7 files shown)


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationBase.java
```diff
@@ -1000,6 +1010,43 @@ public Void withHandle(final Handle handle) throws Exception {
...
+                }
+            } finally {
+                // Go through all results to close the connection
+                while (iterator.hasNext()) {
+                    iterator.next();
+                }
+            }
+        }
+        if (nbNotifications != 0) {
+            log.info(": {} queue(s) with more notification(s) to process", nbNotifications);
```
pierre, line 1033: Nit: the comment is incorrect, it's not the number of queues, but the number of notifications.

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -395,8 +395,10 @@ public void createInvoice(final InvoiceModelDao invoice,
                                 billingEventSqlDao.create(new InvoiceBillingEventModelDao(invoiceModelDao.getId(), BillingEventSerializer.serialize(billingEvents), context.getCreatedDate()), context);
                             }
                             createdInvoiceIds.add(invoiceModelDao.getId());
-                        } else if (invoiceOnDisk.getStatus() == InvoiceStatus.DRAFT && invoiceModelDao.getStatus() == InvoiceStatus.COMMITTED) {
-                            invoiceSqlDao.updateStatus(invoiceModelDao.getId().toString(), InvoiceStatus.COMMITTED.toString(), context);
+                        } else if (invoiceOnDisk.getStatus() != invoiceModelDao.getStatus() || /* Update if status is different */
+                                   (invoiceModelDao.getTargetDate() != null && /* Update if target date is specified and prev was either null or different */
+                                    (invoiceOnDisk.getTargetDate() == null || invoiceOnDisk.getTargetDate().compareTo(invoiceModelDao.getTargetDate()) != 0))) {
```
pierre, line 400: I'm wondering if you need to explicitly check that `invoiceOnDisk.getTargetDate().compareTo(invoiceModelDao.getTargetDate()) < 0`, to make sure we never go back in time.

Could you trigger a target date bill run in the past in your test, before committing the final invoice, and verify the resulting target date?

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -395,8 +395,10 @@ public void createInvoice(final InvoiceModelDao invoice,
                                 billingEventSqlDao.create(new InvoiceBillingEventModelDao(invoiceModelDao.getId(), BillingEventSerializer.serialize(billingEvents), context.getCreatedDate()), context);
                             }
                             createdInvoiceIds.add(invoiceModelDao.getId());
-                        } else if (invoiceOnDisk.getStatus() == InvoiceStatus.DRAFT && invoiceModelDao.getStatus() == InvoiceStatus.COMMITTED) {
-                            invoiceSqlDao.updateStatus(invoiceModelDao.getId().toString(), InvoiceStatus.COMMITTED.toString(), context);
+                        } else if (invoiceOnDisk.getStatus() != invoiceModelDao.getStatus() || /* Update if status is different */
```
pierre, line 398: `Update if status is different` -> I'm a tad nervous about this, as there is nothing preventing the code now to do other transitions, such as `COMMITTED` -> `VOID`.

pierre (changes requested):

sbrossie:
@pierre Agree with all your comments. See https://github.com/killbill/killbill/pull/1360/commits/7514c6fe36b4f39482eb07c3b03cd51b716ce376

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -395,11 +396,22 @@ public void createInvoice(final InvoiceModelDao invoice,
...
                             createdInvoiceIds.add(invoiceModelDao.getId());
-                        } else if ((invoiceOnDisk.getStatus() == InvoiceStatus.DRAFT && invoiceModelDao.getStatus() == InvoiceStatus.COMMITTED) || /* Update if status is needed */
-                                   (invoiceModelDao.getTargetDate() != null && /* Update if target date is specified and prev was either null or prior to new date */
-                                    (invoiceOnDisk.getTargetDate() == null || invoiceOnDisk.getTargetDate().compareTo(invoiceModelDao.getTargetDate()) < 0))) {
-                            invoiceSqlDao.updateStatusAndTargetDate(invoiceModelDao.getId().toString(), invoiceModelDao.getStatus().toString(), invoiceModelDao.getTargetDate(), context);
-                            committedReusedInvoiceId.add(invoiceModelDao.getId());
+                        } else {
+                            // Transition to COMMITTED or keep current status
+                            final InvoiceStatus newStatus = (InvoiceStatus.COMMITTED == invoiceModelDao.getStatus() && InvoiceStatus.COMMITTED != invoiceOnDisk.getStatus()) ? InvoiceStatus.COMMITTED : invoiceOnDisk.getStatus();
```
pierre, line 401: `InvoiceStatus.COMMITTED != invoiceOnDisk.getStatus()` -> `invoiceOnDisk.getStatus() == InvoiceStatus.DRAFT` to be strictly backward compatible (e.g. prevent transition from `VOID` to `COMMITTED`).

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -28,6 +28,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
```
pierre, line 31: Fancy!

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -395,11 +396,22 @@ public void createInvoice(final InvoiceModelDao invoice,
...
+
+                            // Update if target date is specified and prev targetDate was either null or prior to new date
+                            final LocalDate newTargetDate = (invoiceModelDao.getTargetDate() != null &&
+                                                             (invoiceOnDisk.getTargetDate() == null ||
+                                                              invoiceOnDisk.getTargetDate().compareTo(invoiceModelDao.getTargetDate()) < 0)) ?
+                                                            invoiceModelDao.getTargetDate() :
+                                                            invoiceOnDisk.getTargetDate();
+
+                            // If either newStatus or newTargetDate was updated perform the update
+                            if (newStatus != invoiceOnDisk.getStatus() || !Objects.equals(newTargetDate, invoiceOnDisk.getTargetDate())) {
```
pierre, line 411: Does that work with dates (it uses `compareTo`)?

pierre (approved):

PR #1368, opened 2020-10-01
invoice, usage: Add info and debug logs around usage records and dates

sbrossie:


usage/src/main/java/org/killbill/billing/usage/api/svcs/DefaultInternalUserApi.java
```diff
@@ -54,6 +58,8 @@ public DefaultInternalUserApi(final RolledUpUsageDao rolledUpUsageDao,
     @Override
     public List<RawUsageRecord> getRawUsageForAccount(final LocalDate startDate, final LocalDate endDate, final InternalTenantContext internalTenantContext) {
 
+        log.info("GetRawUsageForAccount startDate='{}', endDate='{}'", startDate, endDate);
```
pierre, line 61: `debug`?

usage/src/main/java/org/killbill/billing/usage/api/BaseUserApi.java
```diff
@@ -73,18 +76,75 @@ public BaseUserApi(final OSGIServiceRegistration<UsagePluginApi> pluginRegistry)
...
             }
         }
         // All registered plugins returned null
         return null;
     }
 
+    private static class DebugMap {
+
+        private final Map<UUID, List<RawUsageRecord>> perSubscriptionRecords;
+        private final Logger logger;
```
pierre, line 98: You might not have to pass the `logger` -- I would think it would be visible in here.

usage/src/main/java/org/killbill/billing/usage/api/svcs/DefaultInternalUserApi.java
sbrossie: Was thinking of keeping this one as INFO on purpose (this is just a high level trace to see which calls we make per account-recordId

usage/src/main/java/org/killbill/billing/usage/api/BaseUserApi.java
sbrossie: I think because the class is `static` I need to -- but I have not checked.

PR #1369, opened 2020-10-02
Fix for 667

sbrossie:


invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceWithMetadata.java
```diff
@@ -69,11 +86,63 @@ private void build() {
...
         }
     }
 
+
+    public static  final Map<UUID, DateTime> computeChargedThroughDates(final Invoice invoice, final InternalCallContext context)  {
+        final Map<UUID, DateTime> chargedThroughDates;
+        if (invoice != null &&
+            invoice.getStatus() == InvoiceStatus.COMMITTED) /* See https://github.com/killbill/killbill/issues/1296 */ {
+            chargedThroughDates = new HashMap<>();
+                for (final InvoiceItem item : invoice.getInvoiceItems()) {
```
pierre, line 121: Maybe keep the original comment:

```
// Don't use invoice.getInvoiceItems(final Class<T> clazz) as some items can come from plugins
```

(it's an API we should probably remove at some point)

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -651,17 +653,8 @@ private InvoiceWithFutureNotifications processAccountWithLockAndInputTargetDate(
                 // Commit invoice on disk
                 final ExistingInvoiceMetadata existingInvoiceMetadata = new ExistingInvoiceMetadata(existingInvoices);
                 commitInvoiceAndSetFutureNotifications(account, invoiceModelDao, billingEvents, trackingIds, futureAccountNotifications, existingInvoiceMetadata, internalCallContext);
+                setChargedThroughDatesNoExceptions(invoiceWithMetadata.getChargeThroughDates(), internalCallContext);
```
pierre, line 656: Should we really ignore the exception here?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/usage/TestWithoutZeroUsageItems.java
```diff
@@ -0,0 +1,219 @@
...
+
+public class TestWithoutZeroUsageItems extends TestIntegrationBase {
+
+
+    @Inject
+    protected InvoiceDao invoiceDao;
+
+    @BeforeClass(groups = "slow")
+    public void beforeClass() throws Exception {
+        super.beforeClass();
```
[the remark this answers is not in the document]
sbrossie: We should talk, I am not sure I really understand this.

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceItemGenerator.java
```diff
@@ -37,6 +38,10 @@
 import org.killbill.billing.junction.BillingEventSet;
 import org.slf4j.Logger;
 
+import com.google.common.base.Predicate;
```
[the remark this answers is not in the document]
sbrossie: ack, will remove.

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
sbrossie: There is no good solution. My 2021 view is, that if we come from api, we should throw, but if we come from bus event, we should really park the account as the state becomes inconsistent (but this is not what we do prior my change I think, so as far as I can tell, I tied to current behavior). Let's talk?

pierre (approved):

Issue #1370, opened 2020-10-07
Sub-millisecond precision in catalog date can lead to duplicate invoice items


pierre:
If the catalog effective date is set to `2020-09-16T10:34:25.348646Z` for instance, the DDL must support sub-millisecond precision (by default, it doesn't).

PR #1371, opened 2020-10-07
invoice: ignore sub-second precision when matching items

pierre:
This fixes https://github.com/killbill/killbill/issues/1370.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemCatalogBase.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemCatalogBase.java
@@ -25,6 +25,7 @@ import javax.annotation.Nullable;
 
 import org.joda.time.DateTime;
 import org.joda.time.LocalDate;
+import org.joda.time.Seconds;
 import org.killbill.billing.catalog.api.Currency;
 import org.killbill.billing.invoice.api.InvoiceItem;
 import org.killbill.billing.invoice.api.InvoiceItemType;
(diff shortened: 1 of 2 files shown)


invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemCatalogBase.java
```diff
@@ -127,7 +128,11 @@ public boolean matches(final Object o) {
             return false;
         }
         final InvoiceItemCatalogBase that = (InvoiceItemCatalogBase) o;
-        if (catalogEffectiveDate != null && that.catalogEffectiveDate != null && catalogEffectiveDate.compareTo(that.catalogEffectiveDate) != 0) {
+        if (catalogEffectiveDate != null &&
+            that.catalogEffectiveDate != null &&
+            catalogEffectiveDate.compareTo(that.catalogEffectiveDate) != 0 &&
+            // https://github.com/killbill/killbill/issues/1370
+            Math.abs(Seconds.secondsBetween(catalogEffectiveDate, that.catalogEffectiveDate).getSeconds()) != 0) {
```
sbrossie, line 135: This seems fine, but could be simplified I think, by concatening line 133 and 135: 

`Math.abs(Seconds.secondsBetween(catalogEffectiveDate, that.catalogEffectiveDate).getSeconds())  > 1`

sbrossie (approved):

invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemCatalogBase.java
pierre: True, but I find the current version more explicit (the first statement is the actual logic, the second is a workaround for DB engines limitations).

Issue #1381, opened 2020-10-27
Memory leak when retrieving tags with audit logs


sbrossie:
See [email thread](https://groups.google.com/g/killbilling-users/c/ZZLIAuswcFQ)

sbrossie:
Tried to reproduce the issue, but nothing obvious showed up. I ran the following `cURL` in a row to fetch account tags (with full audit logs):

```
curl -v \
    -u admin:password \
    -H "X-Killbill-ApiKey: bob" \
    -H "X-Killbill-ApiSecret: lazar" \
    -H "Accept: application/json" \
    'http://127.0.0.1:8080/1.0/kb/accounts/21ab8c9d-d86f-4ac1-8fb1-a577d2cefb85/tags?audit=FULL'
```

In between each iteration I check the `main.pool` connections (`/1.0/metrics` endpoint), but I don't see any increment in the # of `TotalConnections`, it remains the same after each iteration (the opposite of what one would expect if there was a leak.)

fbleite:
Have you tried retrieving the tags for an account that does not have any tags associated with it? 
The issue in our case happened after we had already added and removed the `AUTO_INVOICING_OFF` tag from the account, then when we tried to get the account tags for the said account, that at that moment does not have any tags associated with it, then the leak happens.

sbrossie:
@fbleite Ok, yes, that does it... The issue is that we [allocate the iterator for the audit logs](https://github.com/killbill/killbill/blob/killbill-0.22.16/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java#L207) prior we enter the loop to return tags, and so if there is no tags (got added and then removed), we still end up with an iterator that has audit logs but it is never closed.

PR #1383, opened 2020-11-10
Fix connection leak when retrieving tags with audit logs. See #1381

sbrossie:


pierre (approved):
I re-read the implementation of `AccountAuditLogsForObjectType`: don't we also have an issue in the `MINIMAL` case, when the tag is created and deleted (we stop at the `INSERT` and don't consume all entries in `AccountAuditLogsForObjectType#getAuditLogs`)? The bug might depend on the database fetch size though...

Wondering if we shouldn't mirror the `FULL` implementation in `AccountAuditLogsForObjectType#getAuditLogs` for safety?

Note that the full optimization we had in mind initially was never implemented anyways unfortunately (check TODO in `DefaultAccountAuditLogs`).

sbrossie:
@pierre Updated the PR -- there was indeed a leak for the `MINIMAL` type.

PR #1388, opened 2020-11-20
Fix for 1387

sbrossie:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
```diff
@@ -389,18 +389,8 @@ public Response generateDryRunInvoice(@Nullable final InvoiceDryRunJson dryRunSu
...
-            } else if (DryRunType.SUBSCRIPTION_ACTION.equals(dryRunSubscriptionSpec.getDryRunType()) && dryRunSubscriptionSpec.getEffectiveDate() != null) {
-                inputDate = dryRunSubscriptionSpec.getEffectiveDate();
-            } else {
-                inputDate = toLocalDate(targetDate);
-            }
-        } else {
-            inputDate = toLocalDate(targetDate);
-        }
+        final LocalDate inputDate = (dryRunSubscriptionSpec != null && DryRunType.UPCOMING_INVOICE.equals(dryRunSubscriptionSpec.getDryRunType())) ?
+                                     null : toLocalDate(targetDate);
```
pierre, line 393: Since `dryRunSubscriptionSpec.getEffectiveDate()` isn't used anymore for `SUBSCRIPTION_ACTION`, do we need to update the Slate docs to reflect this change?

pierre (approved):

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
sbrossie: No because [it is still used](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L1076). This is the whole point, we need 2 distinct dates, one for the action effective date and one for the target invoice date.

sbrossie: Actually, i do need to update the doc as it is wrong: `Target date, if dryRunType is TARGET_DATE. Otherwise not needed`. Will do...

Issue #1408, opened 2021-03-03
Guice injection issue


sbrossie:
Based on some recent code changes, it seems that we have a latent issue with our overall injection. The exception thrown is a `java.lang.ClassCastException` (and not something related to some missing bindings or circular dependencies).

```
com.google.inject.CreationException: Unable to create injector, see the following errors:

1) Error in custom provider, java.lang.ClassCastException: com.sun.proxy.$Proxy86 cannot be cast to javax.inject.Provider
  at org.killbill.billing.util.glue.CacheModule.configure(CacheModule.java:72) (via modules: org.killbill.billing.server.modules.KillbillServerModule -> org.killbill.billing.util.glue.CacheModule)
  while locating javax.cache.CacheManager
    for the 1st parameter of org.killbill.billing.util.cache.CacheControllerDispatcherProvider.<init>(CacheControllerDispatcherProvider.java:46)
  while locating org.killbill.billing.util.cache.CacheControllerDispatcherProvider
  at org.killbill.billing.util.glue.CacheModule.configure(CacheModule.java:76) (via modules: org.killbill.billing.server.modules.KillbillServerModule -> org.killbill.billing.util.glue.CacheModule)
  while locating org.killbill.billing.util.cache.CacheControllerDispatcher
    for the 1st parameter of org.killbill.billing.util.config.tenant.CacheConfig.<init>(CacheConfig.java:43)
  at org.killbill.billing.util.glue.ConfigModule.configure(ConfigModule.java:42) (via modules: org.killbill.billing.server.modules.KillbillServerModule -> org.killbill.billing.util.glue.ConfigModule)
  while locating org.killbill.billing.util.config.tenant.CacheConfig
    for the 2nd parameter of org.killbill.billing.util.config.definition.MultiTenantEventConfig.<init>(MultiTenantEventConfig.java:37)
  while locating org.killbill.billing.util.config.definition.MultiTenantEventConfig
  at org.killbill.billing.util.glue.ConfigModule.configure(ConfigModule.java:48) (via modules: org.killbill.billing.server.modules.KillbillServerModule -> org.killbill.billing.util.glue.ConfigModule)
  while locating org.killbill.billing.util.config.definition.EventConfig
    for the 2nd parameter of org.killbill.billing.util.optimizer.BusOptimizerOn.<init>(BusOptimizerOn.java:48)
  while locating org.killbill.billing.util.optimizer.BusOptimizerOn
[...]

PR #1409, opened 2021-03-04
Use different Guice Key for the shiro CacheManager injection. See #1408

sbrossie:


profiles/killbill/src/main/java/org/killbill/billing/server/modules/KillBillShiroWebModule.java
```diff
@@ -91,9 +92,9 @@ public String getString(final String propertyName) {
 
         // Magic provider to configure the cache manager
         if (redisCacheConfig.isRedisCachingEnabled()) {
-            bind(CacheManager.class).toProvider(RedisShiroManagerProvider.class).asEagerSingleton();
+            bind(CacheManager.class).annotatedWith(Names.named(SHIRO_CACHE_MANAGER)).toProvider(RedisShiroManagerProvider.class).asEagerSingleton();
         } else {
-            bind(CacheManager.class).toProvider(EhcacheShiroManagerProvider.class).asEagerSingleton();
+            bind(CacheManager.class).annotatedWith(Names.named(SHIRO_CACHE_MANAGER)).toProvider(EhcacheShiroManagerProvider.class).asEagerSingleton();
```
pierre, line 97: This is a `org.apache.shiro.cache.CacheManager`, not a `javax.cache.CacheManager` -- I don't think I understand how it causes #1408.

Are you sure it's not related to the fact that `MultiTenantEventConfig` is a just-in-time binding. Did you try `bind(EventConfig.class).to(MultiTenantEventConfig.class).asEagerSingleton()`?

sbrossie: That's precisely the problem: We have two bindings for the same interface `CacheManager` that bind to 2 different classes -> `ClassCastException`. Fails consistently before fix , and works fine afterwards.

sbrossie:
Closing as this seems like the wrong fix.

PR #1414, opened 2021-03-20
Fix optimized invoice dryrun

sbrossie:


invoice/src/main/java/org/killbill/billing/invoice/optimizer/InvoiceOptimizerBase.java
```diff
@@ -53,6 +53,7 @@ public InvoiceOptimizerBase(final InvoiceDao invoiceDao,
         protected final LocalDate cutoffDate;
         protected final List<Invoice> invoices;
 
+        // Test only
```
pierre, line 56: I usually mark these as `@VisibleForTesting`.

pierre (approved):

invoice/src/main/java/org/killbill/billing/invoice/optimizer/InvoiceOptimizerBase.java
sbrossie: Ack - fixed in https://github.com/killbill/killbill/pull/1414/commits/b7cf7e4fcea7a37475f8eeb1e86b91c472706cbd

PR #1417, opened 2021-03-27
Introduce a mechanism to reschedule invoices on lock contention

sbrossie:
* Cleanup pass in https://github.com/killbill/killbill/pull/1417/commits/225b37f6cd8875295c5175023f2ddc11a2c736cd
* New feature in https://github.com/killbill/killbill/pull/1417/commits/b1e3dff65ac281cb9e14e94b7e551364ff97ae85

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
@@ -284,6 +284,7 @@ public class InvoiceDispatcher {
         return processAccount(false, accountId, targetDate, dryRunArguments, isRescheduled, context);
     }
 
+
     public Invoice processAccount(final boolean isApiCall,
                                   final UUID accountId,
                                   @Nullable final LocalDate targetDate,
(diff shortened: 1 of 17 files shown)


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -311,6 +301,24 @@ public Invoice processAccount(final boolean isApiCall,
             log.warn("Unable to determine parking state for accountId='{}'", accountId);
         }
 
+        if (invoiceOptimizer.isOn()) {
+            final TimeSpan timeSpan = invoiceConfig.getRescheduleIntervalOnLock(context);
+            // Anything below 1sec, we would ignore
+            final int delaySec = (int) TimeUnit.SECONDS.convert(timeSpan.getMillis(), TimeUnit.MILLISECONDS);
+            if (!isApiCall &&
+                !locker.isFree(LockerType.ACCNT_INV_PAY.toString(), accountId.toString()) &&
```
pierre, line 309: I'm fine with the opportunistic approach, but wouldn't it be _more solid_ to simply reschedule in the `LockFailedException` catch block below?

sbrossie: Both approaches would work, but why have every single thread wait in a lock if they can simply figure out right away that the lock is held?

pierre: I'm just thinking about the race condition: you are not guaranteed to reschedule with that opportunistic approach (but happy either way).

sbrossie: I think the race condition would be covered with this but we can talk about it, if you are not convinced -- and if I am wrong, I should definitely change the code!

sbrossie: Fixed in https://github.com/killbill/killbill/pull/1417/commits/c2503c6619274e99caa2908ec6e7ba2a57617ba9

sbrossie: Also added https://github.com/killbill/killbill/pull/1417/commits/0fadbfb4f89cec2d47c958994978448969d166d4 as I think it was a mistake to swallow the exception when coming from api.

pierre (approved):

PR #1424, opened 2021-04-13
util: Fix issue in string template leading to a WARN log

sbrossie:


util/src/main/resources/org/killbill/billing/util/entity/dao/EntitySqlDao.sql.stg
```diff
@@ -375,8 +375,8 @@ auditTableValues() ::= <<
 
 getHistoryForTargetRecordId() ::= <<
 select
-  <idField("")>
-, <prefix>record_id as history_record_id
+  <idField()>
```
pierre, line 378: This should be `<idField("t.")>`.

util/src/main/resources/org/killbill/billing/util/entity/dao/EntitySqlDao.sql.stg
```diff
@@ -375,8 +375,8 @@ auditTableValues() ::= <<
...
+, record_id as history_record_id
```
pierre, line 379: `t.record_id`

pierre (changes requested):

PR #1425, opened 2021-04-15
subscription: Fix issue in PlanAligner leading to exception during un…

sbrossie:
…cancel

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/alignment/PlanAligner.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/alignment/PlanAligner.java
@@ -158,11 +158,7 @@ public class PlanAligner extends BaseAligner {
                                                                                pendingOrLastPlanTransition.getNextPlan(),
                                                                                pendingOrLastPlanTransition.getNextPhase().getPhaseType(),
                                                                                catalog,
-                                                                               // Use the catalog version at subscription creation time: this allows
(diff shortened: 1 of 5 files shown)


subscription/src/main/java/org/killbill/billing/subscription/alignment/PlanAligner.java
```diff
@@ -158,11 +158,7 @@ public TimedPhase getNextTimedPhase(final DefaultSubscriptionBase subscription,
                                                                                pendingOrLastPlanTransition.getNextPlan(),
                                                                                pendingOrLastPlanTransition.getNextPhase().getPhaseType(),
                                                                                catalog,
-                                                                               // Use the catalog version at subscription creation time: this allows
-                                                                               // for PHASE events and uncancel operations for plans/products/pricelists already retired
```
pierre, line 162: Do you know if we already have a test for this? If not, could you remove `pistol-discount-monthly` in a v3 catalog in your new test to verify this?

pierre (approved):

subscription/src/main/java/org/killbill/billing/subscription/alignment/PlanAligner.java
sbrossie: See https://github.com/killbill/killbill/pull/1425/commits/945654822c5a7bb44ae14f4278ef58d0934e9f76

PR #1429, opened 2021-04-22
invoice: Attempt to fix an edge case with invoice optimizer (IN_ARREA…

sbrossie:
…R + back cancelation)

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/optimizer/InvoiceOptimizerExp.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/optimizer/InvoiceOptimizerExp.java
@@ -58,6 +58,7 @@ public class InvoiceOptimizerExp extends InvoiceOptimizerBase {
     private static Logger logger = LoggerFactory.getLogger(InvoiceOptimizerExp.class);
 
     private static Period UNSPECIFIED_PERIOD = new Period(InvoiceConfig.DEFAULT_NULL_PERIOD);
+
     @Inject
     public InvoiceOptimizerExp(final InvoiceDao invoiceDao,
                                final Clock clock,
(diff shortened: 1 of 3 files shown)


sbrossie:
Tests were [successful](https://app.circleci.com/pipelines/github/killbill/killbill/342/workflows/4ca0a2b4-15cc-42f8-b77a-f304b3dd304e), then i amended a comment and canceled the suite.

invoice/src/main/java/org/killbill/billing/invoice/optimizer/InvoiceOptimizerExp.java
```diff
@@ -155,22 +157,45 @@ public boolean apply(final InvoiceItem invoiceItem) {
...
+                                    @Override
+                                    public boolean apply(final InvoiceItem item) {
+                                        return (item.getInvoiceItemType() == InvoiceItemType.RECURRING &&
+                                                item.getSubscriptionId().equals(invoiceItem.getSubscriptionId()) &&
+                                                item.getStartDate().compareTo(invoiceItem.getStartDate()) == 0);
+                                    }
+                                }).orNull();
+                                if (existingItem != null) {
+                                    if (existingItem.getEndDate().compareTo(invoiceItem.getEndDate()) != 0 ||
+                                        !existingItem.getPlanName().equals(invoiceItem.getPlanName()) ||
```
pierre, line 183: Shouldn't we check the phase too?

Thinking out loud: is there any overlap with the `matches` logic? https://github.com/killbill/killbill/blob/00d9ace820bc7ddf6961a16a2727abe17b198bf9/invoice/src/main/java/org/killbill/billing/invoice/model/InvoiceItemCatalogBase.java#L125

sbrossie: Added the check on phaseName as discussed in https://github.com/killbill/killbill/pull/1429/commits/5b280fc088678ff8acdf06647b29e1cabbd8deb1

Agree the logic is different than `InvoiceItemCatalogBase#matches` - I know for sure I need the end date in my case as I have no tree to create the boundaries, but *It seems like* checking on other changes for that same item (i.e same start date) regarding plan make sense, but this path is untested 🤷‍♂️

PR #1431, opened 2021-04-27
util: convert account procedures to PostgreSQL

pierre:
There isn't an easy way to share the code unfortunately.


sbrossie (approved):
Why is it split in 2?

PR #1433, opened 2021-05-19
Update against latest java client + enhance AccountResource#payAllInvoices

sbrossie:
The new behavior for AccountResource#payAllInvoices was implemented in https://github.com/killbill/killbill/pull/1433/commits/74c4f3930fa9267e7c38d0de997e4f9879e6a142

Note: ci tests canceled until we release the new java client -- see https://github.com/killbill/killbill-client-java/pull/89

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
@@ -750,6 +750,7 @@ public class AccountResource extends JaxRsResourceBase {
     @Path("/{accountId:" + UUID_PATTERN + "}/" + INVOICE_PAYMENTS)
     @ApiOperation(value = "Trigger a payment for all unpaid invoices", response = InvoiceJson.class, responseContainer = "List")
     @ApiResponses(value = {@ApiResponse(code = 201, message = "Successful operation"),
+                           @ApiResponse(code = 204, message = "Nothing to pay"),
                            @ApiResponse(code = 404, message = "Invalid account id supplied")})
     public Response payAllInvoices(@PathParam("accountId") final UUID accountId,
                                    @QueryParam(QUERY_PAYMENT_METHOD_ID) final UUID inputPaymentMethodId,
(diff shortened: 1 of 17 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -740,8 +748,8 @@ public Response getInvoicePayments(@PathParam("accountId") final UUID accountId,
     @Produces(APPLICATION_JSON)
     @Consumes(APPLICATION_JSON)
     @Path("/{accountId:" + UUID_PATTERN + "}/" + INVOICE_PAYMENTS)
-    @ApiOperation(value = "Trigger a payment for all unpaid invoices")
-    @ApiResponses(value = {@ApiResponse(code = 204, message = "Successful operation"),
+    @ApiOperation(value = "Trigger a payment for all unpaid invoices", response = InvoiceJson.class, responseContainer = "List")
+    @ApiResponses(value = {@ApiResponse(code = 201, message = "Successful operation"),
```
pierre, line 752: We should update Slate docs.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -779,6 +790,19 @@ public Response payAllInvoices(@PathParam("accountId") final UUID accountId,
                                              null :
                                              (inputPaymentMethodId != null ? inputPaymentMethodId : account.getPaymentMethodId());
                 createPurchaseForInvoice(account, invoice.getId(), amountToPay, paymentMethodId, externalPayment, null, null, pluginProperties, callContext);
+
+                if (filterMinDate == null || filterMinDate.isAfter(invoice.getTargetDate())) {
```
pierre, line 794: Is it really the target date we want to look at here? Or the invoice date?

pierre (approved):
:shipit: Just one question below.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
sbrossie: > We should update Slate docs.

Yes, but not quite ready

I have some second thoughts, I think we should keep the `204` for when there was no payment made, similar to what we do [here](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L383). Also, currently, if no invoice were paid the `Location` link returned is wrong as it would create a follow GET that would fetch *all* invoices

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
sbrossie: Should be the targetDate -- see https://github.com/killbill/killbill/blob/master/invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg#L90

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
sbrossie: Fixed in https://github.com/killbill/killbill/pull/1433/commits/e01dd5cea57974231a38c781b1a5fa972bda2ff7

Issue #1438, opened 2021-06-10
Usage points at cancellation day not taken into account


pierre:
In Kill Bill, by convention, the upper limit of an interval is always open. In practice, this means that when we bill from 2021-07-01 to 2021-08-01, we effectively bill for [2021-07-01, 2021-08-01), i.e. the last day being billed is 2021-07-31.

If a subscription starts at 3pm on 2021-07-01, we effectively bill for [2021-07-01T15:00:00, 2021-08-01T15:00:00), even though the invoice doesn't reflect the hour of the day.

When rolling up the raw usages, we are currently ignoring the last day, so any usages between [2021-08-01T00:00:00, 2021-08-01T15:00:00) are discarded (while these should be billed).

PR #1439, opened 2021-06-10
invoice: take into account usage points at cancellation day

pierre:
This fixes https://github.com/killbill/killbill/issues/1438.

Suggestions for the reviewer:

* Verify expectations in the new Beatrix tests in `TestUsageInArrear.java`. It's always the same scenario (yearly base subscription with usage-based add-on), the main differences are operations taken on the add-on subscription (various cancellation dates or cancel/uncancel).
* Notice how I had to change the `getRawUsageForAccount` query to match the expected behavior of a usage plugin (last day is inclusive).
* Let me know if you think of other regression scenarii to write?


sbrossie (approved):
All good, but I suggest we add this `https://github.com/killbill/killbill/pull/1442`

PR #1445, opened 2021-06-16
invoice: add test for usage sub cancelled at start of term

pierre:


sbrossie:
@pierre Added https://github.com/killbill/killbill/pull/1445/commits/abc78b184f959e2510e7c1525a906c39eea1915a for the fix and matching beatrix use case

pierre:
:shipit: 

(I cannot approve it as I opened the PR initially).

Issue #1453, opened 2021-07-09
Completion of a invoice (payment) refund leads to ClassCastException


sbrossie:
Scenario: In a scenario where the REFUND is first marked as `PENDING`, and later transitioned to `SUCCESS` through the `paymentApi#notifyPendingTransactionOfStateChangedWithPaymentControl` api, we see the following stack trace:

```
Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to java.util.UUID
	at org.killbill.billing.invoice.dao.InvoiceDaoHelper.computeItemAdjustments(InvoiceDaoHelper.java:103)
	at org.killbill.billing.invoice.dao.DefaultInvoiceDao$27.inTransaction(DefaultInvoiceDao.java:881)
	at org.killbill.billing.invoice.dao.DefaultInvoiceDao$27.inTransaction(DefaultInvoiceDao.java:878)
	at org.killbill.billing.util.entity.dao.EntitySqlDaoTransactionalJdbiWrapper$JdbiTransaction.inTransaction(EntitySqlDaoTransactionalJdbiWrapper.java:74)
	at org.killbill.billing.util.entity.dao.EntitySqlDaoTransactionalJdbiWrapper$JdbiTransaction.inTransaction(EntitySqlDaoTransactionalJdbiWrapper.java:61)
	at org.skife.jdbi.v2.sqlobject.InTransactionHandler$1.inTransaction(InTransactionHandler.java:41)

```

PR #1456, opened 2021-07-12
payment: Fix deserialization for refund+item adj. See #1453

sbrossie:


payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
```diff
@@ -459,7 +460,37 @@ private PriorPaymentControlResult getPluginCreditResult(final PaymentControlCont
...
+            }
+
+            final Object v = m.get(k);
+            BigDecimal val;
+            if (v instanceof BigDecimal) {
+                val = (BigDecimal) v;
+            } else if (v instanceof String) {
+                val = new BigDecimal((String) v);
+            } else if (v instanceof Integer) {
+                val = BigDecimal.valueOf((Integer) v);
```
pierre, line 484: Not sure this does what we want. Would it be easier to do something like?
```
try {
  val = new BigDecimal(v.toString());
} catch () { // throw }
```

payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
```diff
@@ -459,7 +460,37 @@ private PriorPaymentControlResult getPluginCreditResult(final PaymentControlCont
...
         }
-        return (Map<UUID, BigDecimal>) prop.getValue();
+        // The deserialization may not recreate the map we expect, i.e Map<UUID, BigDecimal>, so we convert each key/value by hand
+        // See https://github.com/killbill/killbill/issues/1453
+        final Map<UUID, BigDecimal> res = new HashMap<>();
+        final Map m = (Map) prop.getValue();
+        for (final Object k : m.keySet()) {
+            UUID uuid;
+            if (k instanceof String) {
+                uuid = UUID.fromString((String) k);
```
pierre, line 470: You could maybe be a bit more generic:

```
try {
  uuid = UUID.fromString(v.toString());
} catch () { // throw }
```

That way, we'd handle other types as well (e.g. `CharSequence`).

payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
sbrossie: Actually, if the amount is an integer, it gets deserialized as an `Integer` and so I need this `else if` for this.

sbrossie: Fixed in https://github.com/killbill/killbill/pull/1456/commits/a915db3c09ad7c6c384f7c3fe8e0848f36eda12b

payment/src/main/java/org/killbill/billing/payment/invoice/InvoicePaymentControlPluginApi.java
sbrossie: The only use case we really have is one when we ned to convert String -> UUID - i'd rather keep the code as it is.

PR #1461, opened 2021-07-21
util: Add a mechanism to not return old node_infos entries

sbrossie:
The cleanup of the node_info entries relies on the KB shutdown sequence.

Hiwever, there is no guarantee that such shutdown sequence got initiated (e.g OOM JVM crash)
or that it terminated cleanly (k8s KILL -9 pod is takes too long), so with this change we introduce
an executor that refreshes the `updated_date` and the code has been modified to only return recent values
 (updated within last 3 minutes).

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/nodes/DefaultKillbillNodesService.java
+++ b/util/src/main/java/org/killbill/billing/util/nodes/DefaultKillbillNodesService.java
@@ -18,6 +18,8 @@
 package org.killbill.billing.util.nodes;
 
 import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 import javax.inject.Inject;
 
@@ -32,6 +34,7 @@ import org.killbill.billing.util.nodes.dao.NodeInfoModelDao;
 import org.killbill.billing.util.nodes.json.NodeInfoModelJson;
 import org.killbill.billing.util.nodes.json.PluginInfoModelJson;
 import org.killbill.clock.Clock;
+import org.killbill.commons.concurrent.Executors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


util/src/main/java/org/killbill/billing/util/nodes/DefaultKillbillNodesService.java
```diff
@@ -44,20 +47,27 @@
 
     private static final Logger logger = LoggerFactory.getLogger(DefaultKillbillNodesService.class);
 
+    private final static int TERMINATION_TIMEOUT_SEC = 5;
```
pierre, line 50: Could we make these configurable at least through System properties (just in case...)?

util/src/main/java/org/killbill/billing/util/nodes/DefaultKillbillNodesService.java
```diff
@@ -96,6 +121,21 @@ public void start() {
...
         logger.info("Deleting nodeInfo for {}", CreatorName.get());
         nodeInfoDao.delete(CreatorName.get());
+        if (isStopped) {
+            return;
+        }
+        try {
+            nodeInfoExecutor.shutdown();
+            final boolean success = nodeInfoExecutor.awaitTermination(TERMINATION_TIMEOUT_SEC, TimeUnit.SECONDS);
+            if (!success) {
+                logger.warn("NodeInfoExecutor failed to complete termination within " + TERMINATION_TIMEOUT_SEC + "sec");
```
pierre, line 131: You might want to try to `shutdownNow` here.

util/src/test/java/org/killbill/billing/util/nodes/dao/TestNodeInfoDao.java
```diff
@@ -67,6 +67,15 @@ public void testBasic() throws Exception {
         assertEquals(all.size(), 2);
         assertEquals(all.get(0), newNode1);
         assertEquals(all.get(1), newNode2);
+
+        now = clock.getUTCNow();
+        Thread.sleep(1001);
```
pierre, line 72: Could you just `addDeltaFromReality` instead of sleeping?

util/src/main/java/org/killbill/billing/util/nodes/DefaultKillbillNodesService.java
sbrossie: Not against it, but these termination timeout are harcoded all over our code already, e.g [here](https://github.com/killbill/killbill/blob/master/tenant/src/main/java/org/killbill/billing/tenant/api/TenantCacheInvalidation.java#L65) or [here](https://github.com/killbill/killbill-commons/blob/master/queue/src/main/java/org/killbill/queue/DefaultQueueLifecycle.java#L126), ...

util/src/main/java/org/killbill/billing/util/nodes/DefaultKillbillNodesService.java
sbrossie: We could, but again, this not not what we do in other part of the codes, e.g [here](https://github.com/killbill/killbill/blob/master/tenant/src/main/java/org/killbill/billing/tenant/api/TenantCacheInvalidation.java#L113). Also, if trying to shutdown cleanly fails (times-out), then is it important to try to force a shutdown of the pool knowing that shutdown sequence will continue -- and mostly timeout as a result and be aborted?

util/src/test/java/org/killbill/billing/util/nodes/dao/TestNodeInfoDao.java
sbrossie: 👍 https://github.com/killbill/killbill/pull/1461/commits/2d1f8aabb29086530fae3e433747c48ecf9c198e

PR #1466, opened 2021-08-04
shiro: initial support for Auth0

pierre:
Add support for Auth0 authentication and authorization (both username/password and Bearer token).

Unfortunately, I had to copy over some of https://github.com/apache/shiro/pull/129 because of https://issues.apache.org/jira/browse/SHIRO-679 (tracked by https://github.com/killbill/killbill-oss-parent/issues/36): `BearerToken` and `BearerHttpAuthenticationFilter` (no specific logic though, can be swapped with upstream version eventually).

The Bearer filter will now be invoked first, and fallback to the Basic Auth one if there is no token.

I'll also submit a PR to update the RBAC docs.

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/security/shiro/realm/KillBillAuth0Realm.java
+++ b/util/src/main/java/org/killbill/billing/util/security/shiro/realm/KillBillAuth0Realm.java
@@ -50,7 +50,9 @@ import org.apache.shiro.authz.AuthorizationException;
 import org.apache.shiro.authz.AuthorizationInfo;
 import org.apache.shiro.authz.SimpleAuthorizationInfo;
 import org.apache.shiro.realm.AuthorizingRealm;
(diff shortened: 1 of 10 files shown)


util/src/main/java/org/killbill/billing/util/security/shiro/realm/KillBillAuth0Realm.java
```diff
@@ -0,0 +1,591 @@
...
+        public static final ECParameterSpec SECP256K1;
+        public static final ECParameterSpec P_384;
+        public static final ECParameterSpec P_521;
+
+        private EcCurve() {}
+
+        static {
+            // Values obtained from org.bouncycastle.jce.ECNamedCurveTable
+
+            P_256 = new ECParameterSpec(
```
sbrossie, line 523: Black magic...

sbrossie (approved):

Issue #1476, opened 2021-08-11
AUTO_INVOICING_OFF should not prevent manual invoice run


sbrossie:
The account tag `AUTO_INVOICING_OFF` is meant to prevent the system to *automatically* generates invoices (e.g bus events, notifications).

However the current implementation will prevent [manual invoice](https://killbill.github.io/slate/#invoice-trigger-an-invoice-run) runs to also happen.

PR #1478, opened 2021-08-11
invoice, junction. Enables api invoicing when AUTO_INVOICING_OFF is s…

sbrossie:
…et. See #1476

Prior to this, any invoicing attempt would abort if AUTO_INVOICING_OFF was set -- this included dry-run operations.

With this fix, we now allow invoice generation through api even if AUTO_INVOICING_OFF is set. This reflects the true
nature of the tag, i.e to disable 'auto' invoicing from system bus events or notifications.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
@@ -313,7 +313,7 @@ public class InvoiceDispatcher {
             // Grab lock unless we do a dry-run
             final boolean isDryRun = dryRunArguments != null;
             lock = !isDryRun ? locker.lockWithNumberOfTries(LockerType.ACCNT_INV_PAY.toString(), accountId.toString(), invoiceConfig.getMaxGlobalLockRetries()) : null;
-            return processAccountInternal(parkedAccount, accountId, targetDate, dryRunArguments, isRescheduled, context);
+            return processAccountInternal(isApiCall, parkedAccount, accountId, targetDate, dryRunArguments, isRescheduled, context);
         } catch (final LockFailedException e) {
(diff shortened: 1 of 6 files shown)


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -348,15 +349,11 @@ private Invoice processAccountInternal(final boolean parkedAccount,
         try {
             // Make sure to first set the BCD if needed then get the account object (to have the BCD set)
             final BillingEventSet billingEvents = billingApi.getBillingEventsForAccountAndUpdateAccountBCD(accountId, dryRunArguments, context);
-            if (billingEvents.isAccountAutoInvoiceOff()) {
+            if (!isApiCall && billingEvents.isAccountAutoInvoiceOff()) {
                 return null;
             }
 
-            // Avoid pulling all invoices when AUTO_INVOICING_OFF is set since we will disable invoicing later
-            // (Note that we can't return right away as we send a NullInvoice event)
```
pierre, line 356: It looks like we were returning right away before. Was this a stale comment? 🤔

pierre (approved):

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
sbrossie: Yep, I am also having trouble with the comment, as it does not make sense considering [we were returning right away on the line before](https://github.com/killbill/killbill/blob/killbill-0.22.25/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L352) when `AUTO_INVOICING_OFF` was set.

PR #1484, opened 2021-08-20
Account BCD update feature pass 1

pierre:
See individual commits.

☕ Invoice bug fix: ☕
* c6c569d98d097301824604ac323b16cd6032df15: remove a precondition and implement a split in the tree. I would recommend re-reading the `addNode` implementation and the modified test cases, before looking at the diff.
* 978fdf6bdf74beaca2cad014831a53475548514f: new test to highlight what the above enables.

🍺 New features: 🍺 
* 9fa48f06dc6f879832cea2d5e792c6c0472401b0: ability to disable select caches via the new property `org.killbill.cache.disabled`.
* cf6d24b3504ee28393285c53e78f2e13674b851f: new feature to allow account BCD updates (get familiar with `TestWithAccountBCDUpdate#setupScenario`, more tests coming up using that exact scenario).

🍸 _Trivial_ changes: 🍸 
* 07cd841ed20e658a2cb8242fcbb814d4571b3d6f
* 377169c115510c831af56b1b5fd3024430dd4111
* 31da803bee5a6e8b9a00d2dc22a3c1cc4a101e5f
* 4cbbad8fbd8bcd965850632fa7fb142d00302502: not _that_ trivial technically, as the `@Override` broke the `testAddOverlapNode3` test. Fixed subsequently: the test wasn't testing what we thought it was testing anyways (the state exception came from the `DummyNodeInterval` relying on the `super` implementation and missing items, not from the `addNode` insertion).
* bc2a52e3eccb357c6cec9b222514a8f59d421cc0

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithAccountBCDUpdate.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithAccountBCDUpdate.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright 2020-2021 Equinix, Inc
+ * Copyright 2014-2021 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.killbill.billing.beatrix.integration;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.joda.time.DateTime;
+import org.joda.time.LocalDate;
+import org.joda.time.Period;
+import org.killbill.billing.account.api.Account;
+import org.killbill.billing.account.api.DefaultMutableAccountData;
+import org.killbill.billing.api.TestApiListener.NextEvent;
+import org.killbill.billing.beatrix.util.InvoiceChecker.ExpectedInvoiceItemCheck;
+import org.killbill.billing.catalog.api.BillingPeriod;
+import org.killbill.billing.catalog.api.ProductCategory;
+import org.killbill.billing.entitlement.api.DefaultEntitlement;
+import org.killbill.billing.invoice.api.Invoice;
+import org.killbill.billing.invoice.api.InvoiceItemType;
+import org.killbill.billing.platform.api.KillbillConfigSource;
+import org.killbill.billing.util.config.definition.InvoiceConfig;
+import org.killbill.billing.util.features.KillbillFeatures;
+import org.mockito.Mockito;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+
+public class TestWithAccountBCDUpdate extends TestIntegrationBase {
+
+    @BeforeClass(groups = "slow")
+    public void beforeClass() throws Exception {
+        if (hasFailed()) {
+            return;
+        }
+
+        this.killbillFeatures = Mockito.spy(KillbillFeatures.class);
+        Mockito.when(killbillFeatures.allowAccountBCDUpdate()).thenReturn(true);
+
+        super.beforeClass();
+    }
+
+    @Override
+    protected KillbillConfigSource getConfigSource(final Map<String, String> extraProperties) {
+        final Map<String, String> allExtraProperties = new HashMap<String, String>(extraProperties);
+        allExtraProperties.putAll(DEFAULT_BEATRIX_PROPERTIES);
+        allExtraProperties.put(KillbillFeatures.PROP_FEATURE_INVOICE_OPTIMIZATION, "true");
+        allExtraProperties.put("org.killbill.cache.disabled", "account-bcd");
+        return getConfigSource(null, allExtraProperties);
+    }
+
+    @BeforeMethod(groups = "slow")
+    public void beforeMethod() throws Exception {
+        if (hasFailed()) {
+            return;
+        }
+        super.beforeMethod();
+
+        invoiceConfig.setMaxInvoiceLimit(new Period(InvoiceConfig.DEFAULT_NULL_PERIOD));
+    }
+
+    @Test(groups = "slow")
+    public void testAccountBCDChangeWithNoOptimization() throws Exception {
+        final Account account = setupScenario();
+
+        // 2016-7-17: BCD alignment. We repair earliest period and generate trailing pro-ration.
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.INVOICE_PAYMENT, NextEvent.PAYMENT);
+        clock.addDays(27);
+        assertListenerStatus();
+
+        final List<ExpectedInvoiceItemCheck> expectedInvoices = new ArrayList<ExpectedInvoiceItemCheck>();
+        List<Invoice> invoices = invoiceUserApi.getInvoicesByAccount(account.getId(), false, false, callContext);
+        expectedInvoices.add(new ExpectedInvoiceItemCheck(new LocalDate(2016, 5, 17), new LocalDate(2016, 6, 17), InvoiceItemType.RECURRING, new BigDecimal("249.95")));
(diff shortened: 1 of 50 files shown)


util/src/main/java/org/killbill/billing/util/cache/NoOpCacheController.java
```diff
@@ -0,0 +1,98 @@
...
+        return ImmutableList.<K>of();
+    }
+
+    @Override
+    public boolean isKeyInCache(final K key) {
+        return false;
+    }
+
+    @Override
+    public V get(final K key, final CacheLoaderArgument cacheLoaderArgument) {
```
sbrossie, line 47: I don't understand fully the impl. of this class, and why for instance it is not enough to return `null` (nothing in the cache for the key)

sbrossie (commented):
Looks great, left a few comments for discussion

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithAccountBCDUpdate.java
```diff
@@ -0,0 +1,170 @@
...
+        Mockito.when(killbillFeatures.allowAccountBCDUpdate()).thenReturn(true);
+
+        super.beforeClass();
+    }
+
+    @Override
+    protected KillbillConfigSource getConfigSource(final Map<String, String> extraProperties) {
+        final Map<String, String> allExtraProperties = new HashMap<String, String>(extraProperties);
+        allExtraProperties.putAll(DEFAULT_BEATRIX_PROPERTIES);
+        allExtraProperties.put(KillbillFeatures.PROP_FEATURE_INVOICE_OPTIMIZATION, "true");
```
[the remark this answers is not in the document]
pierre: I saw that, but I'd like to add a few more scenarios specifically around Account BCD update, some of them with the optimization turned on.

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -187,6 +187,31 @@ public InvoiceDispatcher(final InvoiceGenerator generator,
...
+            if (billingEvents.isEmpty()) {
+                return;
+            }
+
+            final ImmutableAccountData account = accountApi.getImmutableAccountDataById(accountId, internalCallContext);
+
+            // Don't trigger an invoice now, or we might repair the current period: in practice, we expect the BCD change to apply at the billing cycle
+            DateTime rescheduleDate = clock.getUTCNow().withDayOfMonth(newBCD);
+            if (rescheduleDate.isBefore(clock.getUTCNow())) {
+                rescheduleDate = rescheduleDate.plusMonths(1);
```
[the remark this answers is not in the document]
pierre: Let's say an account is invoiced on the 20th. On 06/20, the customer asks for his BCD to be the 17 moving forward. We would want the next invoice to be generated on 07/17. The only way to do that is to enqueue a notification on the 17th (this month or next month, depending on which day it is today).

invoice/src/main/java/org/killbill/billing/invoice/InvoiceListener.java
```diff
@@ -176,6 +180,21 @@ public void run(final RequestedSubscriptionInternalEvent event) {
                                                  dispatcher.processSubscriptionStartRequestedDate(event, context);
                                              }
                                          });
+        subscriberQueueHandler.subscribe(AccountChangeInternalEvent.class,
+                                         new SubscriberAction<AccountChangeInternalEvent>() {
+                                             @Override
+                                             public void run(final AccountChangeInternalEvent event) {
+                                                 for (final ChangedField changedField : event.getChangedFields()) {
+                                                     if ("billCycleDayLocal".equals(changedField.getFieldName()) &&
```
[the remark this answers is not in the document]
pierre: I don't think it's needed since without the feature on, the code path cannot be exercised (BCD immutable). That's why no test needed to be changed!

invoice/src/main/java/org/killbill/billing/invoice/tree/NodeInterval.java
```diff
@@ -111,26 +114,67 @@ public boolean addNode(final NodeInterval newNode, final AddNodeCallback callbac
...
         while (curChild != null) {
-
-            // newNode is contained, we go deeper in the tree
+            // newNode is contained (i.e. fits into curChild interval), we go deeper in the tree
             if (curChild.isItemContained(newNode)) {
                 return curChild.addNode(newNode, callback);
             }
 
-            // newNode overlaps, we have to rebalance
+            // newNode overlaps (i.e. curChild is contained in newNode), we have to rebalance:
```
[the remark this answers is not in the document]
pierre: Funny: for me, having the tree printed out vertically helps a ton 😄 How about you add some comments here that _speak_ to you? I don't think we can have too many comments in this part of the code...

invoice/src/main/java/org/killbill/billing/invoice/tree/NodeInterval.java
```diff
@@ -111,26 +114,67 @@ public boolean addNode(final NodeInterval newNode, final AddNodeCallback callbac
...
+                    curChild.getParent().addNode(newNodes[0], callback);
+                    curChild.getParent().addNode(newNodes[1], callback);
+                    return true;
+                }
 
                 // We have not implemented all cases so adding preconditions
-                Preconditions.checkState(newNode.getEnd().compareTo(curChild.getStart()) <= 0,
-                                         "Failed to insert new node %s, end date overlaps with right child %s", newNode, curChild);
-
                 Preconditions.checkState(prevChild == null || newNode.getStart().compareTo(prevChild.getEnd()) >= 0,
```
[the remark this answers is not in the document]
pierre: Sounds good! That was indeed on my TODO list...

util/src/main/java/org/killbill/billing/util/cache/NoOpCacheController.java
pierre: This isn't the cache: this is the `CacheController`. Because we have implemented the Read-through pattern (https://www.ehcache.org/documentation/3.3/caching-patterns.html#read-through), we never really see the caches directly, but instead a `CacheController` (which knows when to delegate to the loader).

pierre:
I've merged https://github.com/killbill/killbill/pull/1486 into this. Two new commits to review independently: 🙏 

* cd841fb3e71c872dd083605009067454d97306ea: behavior change in `testAccountBCDChangeWithNoOptimization` after your fix (as expected: additional repairs are not generated anymore)
* 858ac5da19414b0f4684fd42108906b2d3795fb9: behavior change in `testAccountBCDChangeWithNoOptimization` after triggering an invoice run when the BCD is changed, instead of fiddling with notification dates directly. Take a look at the commit comment for details

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceWithMetadata.java
```diff
@@ -155,6 +154,16 @@ public boolean apply(final InvoiceItem input) {
         });
     }
 
+    private boolean hasItemsForDate(final LocalDate date) {
```
sbrossie, line 157: Don't really understand what this is about, and why this was missing 🤔

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithAccountBCDUpdate.java
```diff
@@ -175,11 +175,18 @@ private Account setupScenario() throws Exception {
         invoiceChecker.checkInvoice(invoices.get(3).getId(), callContext, expectedInvoices);
         expectedInvoices.clear();
 
-        // Move the BCD back to the 17 -- no repair happens upon the BCD change
+        // Move the BCD back to the 17 -- the 3 days between 2016-07-17 and 2016-07-20 are repaired
+        busHandler.pushExpectedEvents(NextEvent.INVOICE);
         accountUserApi.updateAccount(account.getId(), new DefaultMutableAccountData(null, null, null, 0, null, null, null, 17, null, null, null, null, null, null, null, null, null, null, null, null, null, false), callContext);
```
sbrossie, line 180: Nit: One thought would be to enhance our `TestApiListener` and create a new `NextEvent.ACCOUNT_BCP_UPDATE` that we could expect here.

sbrossie (commented):
Looks good, I like the approach better. Just one thing with the fix on the `InvoiceWithMetadata` I 'd like to understand.

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceWithMetadata.java
pierre: When only a `REPAIR_ADJ` is generated, we were clearing the notifications entirely. As a side effect, we were not generating the invoice at the next BCD (2016-7-17) in `testAccountBCDChangeWithNoOptimization` -- we were waiting until the 20th.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithAccountBCDUpdate.java
pierre: I had looked into it -- this should really be `NextEvent.ACCOUNT_CHANGE` but that probably requires changing a lot of tests!

PR #1486, opened 2021-08-24
Test for unnecessary repairs with fixes

sbrossie:
The test `TestSubscriptionItemTree#testWithBCDChange4` should now pass as expecting and all invoice tests pass but `ci` still unhappy because of `TestWithAccountBCDUpdate#testAccountBCDChangeWithNoOptimization`, and I was not sure what the status was.

Note: There is more work to properly implement a generic tree where we can split and rebalance arbitrary items at arbitrary levels. I have some thoughts but this would require a good amount of bandwidth...


sbrossie:
I believe this can be closed -- commits were integrated into [other PR](https://github.com/killbill/killbill/pull/1484)

PR #1493, opened 2021-08-29
Re-implement NodeInterval#addNode in a generic way

sbrossie:
**Overview**

The method `NodeInterval#addNode` required to construct our invoice item tree based on `existing` and then `proposed` items was initially implemented  with simple use cases where various invoice items represented as `NodeInterval` were not supposed to overlap.

Starting from [this commit](https://github.com/killbill/killbill/commit/ba81d9b53173f0b09ace71cdbef018f05f914a56), we introduced the ability to split `NodeInterval` and because the tree was not generic we added preconditions in a lot of places to deal with non implemented use cases.

Then, more recently we had to implement [more cases](https://github.com/killbill/killbill/commit/c6c569d98d097301824604ac323b16cd6032df15), and then the logic to split started [to leak into the code where we were rebalancing items](https://github.com/killbill/killbill/pull/1484/commits/07eb86f3fbf2fd132886c07d8e79ba254afcc681), and not only we were left with a lot of cases to implement, but it also became clear this was not the right implementation.

This PR re-implements this whole method in a much simpler and generic way. However, by doing so, it changes the way the tree is constructed and this led to some changes in a few tests:
* All high level `profile`, `beatrix`, and `integration-tests` were left untouched 🎉 
* A few invoice unit tests had to be modified, and those will be detailed below.

The main code change is encapsulated in this https://github.com/killbill/killbill/pull/1493/commits/6338109f8cdc7f21eccdc4d9ac1aad79382db232 along with this https://github.com/killbill/killbill/pull/1493/commits/51e87c300db4d30851d7a615b6194efe3d5eedbf for more documentation. Probably easier to read the [code of the one resulting method](https://github.com/killbill/killbill/blob/invoice-tree-fix/invoice/src/main/java/org/killbill/billing/invoice/tree/NodeInterval.java#L107) ☕ 

**Invoice Unit Tests**

* https://github.com/killbill/killbill/pull/1493/commits/7876bae255d703f1380b23087ee82720319fde2c: Remove invalid test cases (node rebalancing don't exist anymore)
* https://github.com/killbill/killbill/pull/1493/commits/abd65eb042e499ad8ec44ebb01b1a02e781bdf90: An edge case scenario with a partial invoice item adjustment led to a different behavior. Adding a beatrix level test to reproduce the issue with old behavior showed that code was unstable and would lead to a 'double invoicing' exception. With new behavior, it produces a fine result and no exception. 🎉 
* The following are all edge cases where we have existing invalid recurring items and some invoice code running to make to repair and re-invoicing. New beatrix tests have been introduced first with the old behavior, and then by ingesting the data from this old behavior with the desire effect of showing that new behavior would lead a no-op. 🎉 
  * https://github.com/killbill/killbill/pull/1493/commits/47cad58d1c26ade31031634317ab1235fe1ffe2a
  * https://github.com/killbill/killbill/pull/1493/commits/4207ad2f26f25ed5b57481eaec95f428312d5dd0
* Finally there are some invoice unit tests that allowed for inserting some items into the tree out of sequence that would now produce different results. Because the use cases don't make sense in a real scenario, such assertions have been removed:
  * https://github.com/killbill/killbill/pull/1493/commits/dd3e1c64774906ecda080b114f11ed84ad2b1e17
  * https://github.com/killbill/killbill/pull/1493/commits/9b30fb3bd9d041daac77e1102c2c43bd1294e2d2

**Risks**

Because we change the way we construct our invoice item tree, there is a certain amount of risk with folks having old data. However we have seen that none of our high level tests were modified (and we have some fairly complicated use cases at this level) so for most real use cases this should not be an issue; for edge cases, we have also spent great amount of efforts showing that everything will be correctly handled (no-op) or lead to a better result.

Happy review! 🍻

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/tree/NodeInterval.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/tree/NodeInterval.java
@@ -80,6 +80,7 @@ public class NodeInterval {
         return;
     }
 
+    // Convenient (static) method to insert a node that belongs below parent, right to prevNode (if not null) and before nextNode (if not null)
     private static boolean insertNode(final NodeInterval parentNode, @Nullable final NodeInterval prevNode, @Nullable final NodeInterval nextNode, final NodeInterval newNode, final AddNodeCallback callback) {
         if (!callback.shouldInsertNode(parentNode, (ItemsNodeInterval) newNode)) {
             return false;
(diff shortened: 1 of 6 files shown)


pierre (approved):
Thank you for a well crafted PR, it definitively helped the reviewer! 🙏 

I ran code coverage analysis because I wanted to make sure all branches are now tested. I believe this is the case for the main `addNode` method, but a few other methods don't have 100% coverage. I suspect this is because it's old methods that are either unused and / or only used in tests (`isThisItemEncompassed`, `isThisItemOverlaped`, `isThisItemOverlapedLeft`, `isThisItemOverlapedRight`, `removeChild`, etc.).

Could you clean these up?

For instance (see red markings on the left):

image

invoice/src/main/java/org/killbill/billing/invoice/tree/NodeInterval.java
```diff
@@ -223,7 +225,8 @@ private void adjustRightMostChildSibling(final NodeInterval curNode) {
      * @param callback   custom logic to decide if a given node is a match
      * @return the found node or null if there is nothing.
      */
-    public NodeInterval findNode(final LocalDate targetDate, final SearchCallback callback) {
+    @VisibleForTesting
+    NodeInterval findNode(final LocalDate targetDate, final SearchCallback callback) {
```
pierre, line 229: Can't we just delete it? It seems to be used only in `testSearch`, which I believe tests it.

invoice/src/main/java/org/killbill/billing/invoice/tree/NodeInterval.java
```diff
@@ -179,7 +180,8 @@ public boolean addNode(final NodeInterval newNode, final AddNodeCallback callbac
         return insertNode(this, prevChild, null, newNode, callback);
     }
 
-    public void removeChild(final NodeInterval toBeRemoved) {
+    @VisibleForTesting
+    void removeChild(final NodeInterval toBeRemoved) {
```
pierre, line 184: Same question: it looks like it's only used in `testRemove*` which test it.

sbrossie: See https://github.com/killbill/killbill/pull/1493/commits/0509ee20a4f8cd2ea83910d6ea4f3a566f970548

invoice/src/main/java/org/killbill/billing/invoice/tree/NodeInterval.java
sbrossie: See https://github.com/killbill/killbill/pull/1493/commits/0509ee20a4f8cd2ea83910d6ea4f3a566f970548

PR #1525, opened 2021-10-29
Invoice html fix

reshmabidikar:
- Fixed TestInvoice#testInvoiceOk for new HTML template but disabled this test as it produces different output in Java 8 and Java 11. 
- Fixed test_ddl.sql

### DIFF
--- a/util/src/test/resources/org/killbill/billing/util/ddl_test.sql
+++ b/util/src/test/resources/org/killbill/billing/util/ddl_test.sql
@@ -65,12 +65,13 @@ CREATE TABLE bundles (
     tenant_record_id bigint /*! unsigned */ not null default 0,
     PRIMARY KEY(record_id)
 ) /*! CHARACTER SET utf8 COLLATE utf8_bin */;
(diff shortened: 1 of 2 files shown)


util/src/test/resources/org/killbill/billing/util/ddl_test.sql
```diff
@@ -65,12 +65,13 @@ CREATE TABLE bundles (
     tenant_record_id bigint /*! unsigned */ not null default 0,
     PRIMARY KEY(record_id)
 ) /*! CHARACTER SET utf8 COLLATE utf8_bin */;
-
+CREATE UNIQUE INDEX bundles_external_key ON bundles(external_key, tenant_record_id);
```
sbrossie, line 68: I suppose this is because of [this](https://github.com/killbill/killbill/blob/master/subscription/src/main/resources/org/killbill/billing/subscription/ddl.sql#L125) ?

sbrossie (approved):

Issue #1465, opened 2021-08-04
Additional catalog validation for phase types and duration


pierre:
`FIXEDTERM` phase types must have a non-`UNLIMITED` `Duration` specified, while `EVERGREEN` phases are assumed to be `UNLIMITED`.

While we cannot easily capture that logic inside the XSD, we could add additional validation as part of our `ValidatingConfig` runtime validation framework.

Issue #1481, opened 2021-08-16
The upload catalog code allows uploading a bad catalog


reshmabidikar:
The upload catalog code does not run full versioned catalog validations and allows uploading a bad catalog. For example, if there is a plan with a phase type as `EVERGREEN`, the upload catalog code allows uploading a catalog that changes the phase type of the plan to `FIXEDTERM`.  This should not be allowed, that is it should not be possible to change the phases type of a plan across catalog versions. 

1. Add full validation when the catalog is uploaded
2. Add a test to verify the behavior above

Relevant code:

https://github.com/killbill/killbill/blob/99f7102c83cefe892027f4ac0d1ab4da37dd517b/catalog/src/main/java/org/killbill/billing/catalog/api/user/DefaultCatalogUserApi.java#L97

 https://github.com/killbill/killbill/blob/99f7102c83cefe892027f4ac0d1ab4da37dd517b/catalog/src/main/java/org/killbill/billing/catalog/io/VersionedCatalogLoader.java#L98

sbrossie:
Fixed as https://github.com/killbill/killbill/pull/1518

PR #1529, opened 2021-11-17
Fix for #1465

tejinderghatore:
Fix for https://github.com/killbill/killbill/issues/1465

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
@@ -68,11 +68,13 @@ public class TestCatalogValidation extends TestIntegrationBase {
     @Test(groups = "slow", description = "https://github.com/killbill/killbill/issues/1465")
     public void testUploadCatalogPlanValidation() throws Exception {
 	/*
-	* This is the test case for #1465. CatalogValidation-v3.xml is added to the DB even if the validation fails for 1465. So to test this #1465 validation, 2nd catalog CatalogValidation-v4.xml needs to be updated and it validates those validations.
+	* This is the test case for #1465.
 	*/
+    	int catalogUploaded = 0;
 	try {
 	    uploadCatalog("CatalogValidation-v3.xml");
 	    assertListenerStatus();
+	    catalogUploaded = catalogUploaded + 1;
 	} catch (CatalogApiException cApiException) {
 		assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
 	}
@@ -80,10 +82,12 @@ public class TestCatalogValidation extends TestIntegrationBase {
 	try {
 	    uploadCatalog("CatalogValidation-v4.xml");
 	    assertListenerStatus();
+	    catalogUploaded = catalogUploaded + 1;
 	} catch (CatalogApiException cApiException) {
 		assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
 	}
-
+	//Since out of the above 2 catalogs only 1 is uploaded, so catalogUploaded should be 1.
+	assertEquals(catalogUploaded, 1);
     }
     private void uploadCatalog(final String name) throws CatalogApiException, IOException {
         catalogUserApi.uploadCatalog(Resources.asCharSource(Resources.getResource("catalogs/testCatalogValidation/" + name), Charsets.UTF_8).read(), testCallContext);
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
@@ -66,23 +66,23 @@ public class TestCatalogValidation extends TestIntegrationBase {
     }
     
     @Test(groups = "slow", description = "https://github.com/killbill/killbill/issues/1465")
-	public void testUploadCatalogPlanValidation() throws Exception {
-		try {
-			uploadCatalog("CatalogValidation-v2.xml");
-			assertListenerStatus();
-		} catch (CatalogApiException cApiException) {
-			assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
-		}
-
-		try {
-			uploadCatalog("CatalogValidation-v3.xml");
-			assertListenerStatus();
-		} catch (CatalogApiException cApiException) {
-			//assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
-			assertNotEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
-		}
-
+    public void testUploadCatalogPlanValidation() throws Exception {
+	try {
+	    uploadCatalog("CatalogValidation-v2.xml");
+	    assertListenerStatus();
+	} catch (CatalogApiException cApiException) {
+		assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
 	}
+
+	try {
+	    uploadCatalog("CatalogValidation-v3.xml");
+	    assertListenerStatus();
+	} catch (CatalogApiException cApiException) {
+		//assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
+		assertNotEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
+	}
+
+    }
     private void uploadCatalog(final String name) throws CatalogApiException, IOException {
         catalogUserApi.uploadCatalog(Resources.asCharSource(Resources.getResource("catalogs/testCatalogValidation/" + name), Charsets.UTF_8).read(), testCallContext);
     }
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
@@ -67,19 +67,21 @@ public class TestCatalogValidation extends TestIntegrationBase {
     
     @Test(groups = "slow", description = "https://github.com/killbill/killbill/issues/1465")
     public void testUploadCatalogPlanValidation() throws Exception {
+	/*
+	* This is the test case for #1465. CatalogValidation-v3.xml is added to the DB even if the validation fails for 1465. So to test this #1465 validation, 2nd catalog CatalogValidation-v4.xml needs to be updated and it validates those validations.
+	*/
 	try {
-	    uploadCatalog("CatalogValidation-v2.xml");
+	    uploadCatalog("CatalogValidation-v3.xml");
 	    assertListenerStatus();
 	} catch (CatalogApiException cApiException) {
 		assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
 	}
 
 	try {
-	    uploadCatalog("CatalogValidation-v3.xml");
+	    uploadCatalog("CatalogValidation-v4.xml");
 	    assertListenerStatus();
 	} catch (CatalogApiException cApiException) {
-		//assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
-		assertNotEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
+		assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
 	}
 
     }
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
@@ -18,16 +18,14 @@
 package org.killbill.billing.beatrix.integration;
 
 import static org.testng.Assert.assertEquals;
-
+import static org.testng.Assert.assertNotEquals;
 import java.io.IOException;
 
 import org.joda.time.DateTime;
 import org.killbill.billing.ErrorCode;
 import org.killbill.billing.account.api.Account;
 import org.killbill.billing.catalog.api.CatalogApiException;
-import org.killbill.billing.catalog.api.VersionedCatalog;
 import org.killbill.billing.util.callcontext.CallContext;
-import org.testng.Assert;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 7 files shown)


catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
```diff
@@ -190,6 +190,21 @@ private void validatePlanShape(final Plan plan, final Plan targetPlan, final Val
...
+        }
+    }
+    
+    private void validatePlanDuration(final Plan plan, final PlanPhase cur, final ValidationErrors errors) {
+        if(cur.getPhaseType().name().equals(PhaseType.EVERGREEN.name()) && !cur.getDuration().getUnit().name().equals(TimeUnit.UNLIMITED.name())) {
+            errors.add(new ValidationError(String.format("EVERGREEN Phase '%s'for plan '%s' in version '%s'  must have duration as UNLIMITED'",
+                    cur.getName(), plan.getName(), plan.getCatalog().getEffectiveDate()),
+                    DefaultVersionedCatalog.class, "")); 
+        }
+        if(cur.getPhaseType().name().equals(PhaseType.FIXEDTERM.name()) && cur.getDuration().getUnit().name().equals(TimeUnit.UNLIMITED.name())) {
```
sbrossie, line 204: I know this was not specified in the description of #1465 but it would probably make sense to also add this same validation for `PhaseType#TRIAL` and `PhaseType#DISCOUNT`. @pierre Any thoughts?

catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
```diff
@@ -190,6 +190,21 @@ private void validatePlanShape(final Plan plan, final Plan targetPlan, final Val
                                                              cur.getName(), plan.getName(), plan.getCatalog().getEffectiveDate(), targetPlan.getCatalog().getEffectiveDate()),
                                                DefaultVersionedCatalog.class, ""));
             }
+	    //Fix for https://github.com/killbill/killbill/issues/1465
+            validatePlanDuration(plan, cur, errors);
+        }
+    }
+    
+    private void validatePlanDuration(final Plan plan, final PlanPhase cur, final ValidationErrors errors) {
+        if(cur.getPhaseType().name().equals(PhaseType.EVERGREEN.name()) && !cur.getDuration().getUnit().name().equals(TimeUnit.UNLIMITED.name())) {
```
sbrossie, line 199: Nit: `if (` -> missing space.

sbrossie (commented):
Fix looks good, but I don't see any changes for the test `TestCatalogValidation#testUploadCatalog` so I don't really understand how we have now test logic in place to verify the new behavior?

catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
```diff
@@ -190,6 +190,21 @@ private void validatePlanShape(final Plan plan, final Plan targetPlan, final Val
...
+            errors.add(new ValidationError(String.format("EVERGREEN Phase '%s'for plan '%s' in version '%s'  must have duration as UNLIMITED'",
```
pierre, line 200: Missing space before `for plan`.
Double spaces (extra space) before `must`.

catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
pierre: Yup, that makes sense.

tejinderghatore: @sbrossie Can you tell me the exact validations to be put for PhaseType#TRIAL and PhaseType#DISCOUNT .

catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
tejinderghatore: Fixed.

catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
tejinderghatore: Fixed.

tejinderghatore:
@sbrossie  : I have not made any changes in `TestCatalogValidation#testUploadCatalog` for this, as this new validation will be called from testUploadCatalog() test case only. Do let me know if I need to write a separate testCase for testing this validation only.

catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
sbrossie: Validation for `TRIAL`  and `DISCOUNT` should be similar to `FIXEDTERM`

sbrossie:
> I have not made any changes in `TestCatalogValidation#testUploadCatalog` for this, as this new validation will be called from testUploadCatalog() test case only. Do let me know if I need to write a separate testCase for testing this validation only.

If the test already existed prior the fix and if it has not been updated (and assuming it still passes), it means it is not testing for the new code. We just a test assertion somewhere, whether this existing test or a new one that validates the new code.

catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
tejinderghatore: This is done.

beatrix/src/test/resources/catalogs/testCatalogValidation/CatalogValidation-v2.xml
```diff
@@ -45,7 +45,7 @@
             <product>Standard</product>
             <initialPhases>
             </initialPhases>
-            <finalPhase type="FIXEDTERM">
+            <finalPhase type="EVERGREEN">
```
sbrossie, line 48: ⚠️  Will changing this not affect the test that was written for https://github.com/killbill/killbill/issues/1481 - we need two version of the same Plan, one with `EVERGREEN` and the other with `FIXEDTERM` ?

catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java
```diff
@@ -190,6 +190,31 @@ private void validatePlanShape(final Plan plan, final Plan targetPlan, final Val
...
+        }
+    }
+    
+    private void validatePlanDuration(final Plan plan, final PlanPhase cur, final ValidationErrors errors) {
+        if (cur.getPhaseType().name().equals(PhaseType.EVERGREEN.name()) && !cur.getDuration().getUnit().name().equals(TimeUnit.UNLIMITED.name())) {
+            errors.add(new ValidationError(String.format("EVERGREEN Phase '%s' for plan '%s' in version '%s' must have duration as UNLIMITED'",
+                    cur.getName(), plan.getName(), plan.getCatalog().getEffectiveDate()),
+                    DefaultVersionedCatalog.class, "")); 
+        }
+        if (cur.getPhaseType().name().equals(PhaseType.FIXEDTERM.name()) && cur.getDuration().getUnit().name().equals(TimeUnit.UNLIMITED.name())) {
```
sbrossie, line 204: This is 3 times the same check, you could greatly simplify this with:

```
if (cur.getPhaseType().name().equals(PhaseType.EVERGREEN.name()) && !cur.getDuration().getUnit().name().equals(TimeUnit.UNLIMITED.name())) { /* EVERGREEN cannot have UNLIMITED duration */
...
} else if (!cur.getPhaseType().name().equals(PhaseType.EVERGREEN.name())) && cur.getDuration().getUnit().name().equals(TimeUnit.UNLIMITED.name())) {  /* FIXEDTERM, TRIAL, DISCOUNT need to have UNLIMITED duration */
}
```

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
```diff
@@ -67,6 +65,24 @@ public void testUploadCatalog() throws Exception {
     	}
     }
     
+    @Test(groups = "slow", description = "https://github.com/killbill/killbill/issues/1465")
+    public void testUploadCatalogPlanValidation() throws Exception {
+	try {
+	    uploadCatalog("CatalogValidation-v2.xml");
+	    assertListenerStatus();
```
sbrossie, line 72: Is this for testing #1481, if not what are we testing?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
```diff
@@ -67,6 +65,24 @@ public void testUploadCatalog() throws Exception {
...
+	} catch (CatalogApiException cApiException) {
+		assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
+	}
+
+	try {
+	    uploadCatalog("CatalogValidation-v3.xml");
+	    assertListenerStatus();
+	} catch (CatalogApiException cApiException) {
+		//assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
```
sbrossie, line 81: Nit: Remove extra comment

sbrossie (changes requested):

sbrossie:
@tejinderghatore Also your tests are failing: In terms of ordering, I would rather you *first* fix the tests and then prompt us for code review, this would save us some time 🙏  - unless this is a partial review to discuss something specific.

```2021-11-18T13:07:09.4558658Z [ERROR] Failures: 
2021-11-18T13:07:09.4560776Z [ERROR]   TestCatalogValidation>GuicyKillbillTestSuite.run:209->testUploadCatalogPlanValidation:82 did not expect [2080] but found [2080]
2021-11-18T13:07:09.4562320Z [INFO] 
2021-11-18T13:07:09.4562962Z [ERROR] Tests run: 314, Failures: 1, Errors: 0, Skipped: 116```

File: catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java (hunk 5 above)
Comment (tejinderghatore) [reply catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java]: Fixed.

File: beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java (hunk 6 above)
Comment (tejinderghatore) [reply beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java:75]: Yes you are right. I have reverted this change.

File: beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java (hunk 7 above)
Comment (tejinderghatore) [reply beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java]: Done.

File: beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java [hunk 8]
```diff
@@ -67,6 +65,24 @@ public void testUploadCatalog() throws Exception {
...
+    public void testUploadCatalogPlanValidation() throws Exception {
+	try {
+	    uploadCatalog("CatalogValidation-v2.xml");
+	    assertListenerStatus();
+	} catch (CatalogApiException cApiException) {
+		assertEquals(cApiException.getCode(), ErrorCode.CAT_INVALID_FOR_TENANT.getCode());
+	}
+
+	try {
+	    uploadCatalog("CatalogValidation-v3.xml");
```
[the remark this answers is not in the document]
tejinderghatore: Added a comment for this.

tejinderghatore:
> @tejinderghatore Also your tests are failing: In terms of ordering, I would rather you _first_ fix the tests and then prompt us for code review, this would save us some time 🙏 - unless this is a partial review to discuss something specific.
> 
> ```
> 2021-11-18T13:07:09.4560776Z [ERROR]   TestCatalogValidation>GuicyKillbillTestSuite.run:209->testUploadCatalogPlanValidation:82 did not expect [2080] but found [2080]
> 2021-11-18T13:07:09.4562320Z [INFO] 
> 2021-11-18T13:07:09.4562962Z [ERROR] Tests run: 314, Failures: 1, Errors: 0, Skipped: 116
> ```

@sbrossie I have made the changes in the PR and updated my comments.

1. When I test the new validation as part of uploadCatalog UI functionality it is working as expected and if  the validation fails as per 1465, the catalog is not uploaded in DB and we get the validation message. So using the UI it is working as expected.
2. When I test the new validation using the test case (`TestCatalogValidation#testUploadCatalogPlanValidation`), I have to use 2 catalogs. The first one will always get inserted in the DB (because the versionedCatalog as per line 102 in `https://github.com/tejinderghatore/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/api/user/DefaultCatalogUserApi.java` it returns null the first time for the first catalog and for the 2nd catalog, it gets the first catalog and runs the validation). I am not able to understand, why is this happening in the case of test case execution. If you can help me in understanding this, it will be heplful.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java
```diff
@@ -67,6 +65,26 @@ public void testUploadCatalog() throws Exception {
     	}
     }
     
+    @Test(groups = "slow", description = "https://github.com/killbill/killbill/issues/1465")
+    public void testUploadCatalogPlanValidation() throws Exception {
+	/*
+	* This is the test case for #1465. CatalogValidation-v3.xml is added to the DB even if the validation fails for 1465. So to test this #1465 validation, 2nd catalog CatalogValidation-v4.xml needs to be updated and it validates those validations.
```
sbrossie, line 71: I don't understand the comment, and how it relates to the test, but the fact that the comment is misleading is not even the issue here. The issue is that your test does not do what it should be doing, and it does not check that the validation code is correct (and it's not correct). so, let's unpack everything:


In the first validation for `CatalogValidation-v3.xml`, you have a `standard-monthly` with an `EVERGREEN` final phase whose duration is *not* unlimited, so this is indeed a good test case for #1465. However, you don't have any assertion to verify that the upload failed, and as a result you are not hitting the assertion in the `catch` statement line 77. As a result, your catalog version gets uploaded.

The reason why the validation does not work is because of at least these 2 reasons:
* [this](https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/api/user/DefaultCatalogUserApi.java#L109): Since this is the first version, `versionedCatalog` ends up being null, and you never enter the validation path.
* But there is more, even if you were to fix this, there is another issue [here](https://github.com/tejinderghatore/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/DefaultVersionedCatalog.java#L167): Your code only calls the `validatePlanShape` if there are multiple versions, which is true if you compare things across versions (#1481) but not true if you check things within one version (#1465)

Then, in the second validation for `CatalogValidation-v4.xml`, because of the previous issue**s**, there is now a version of the catalog uploaded with a plan `standard-monthly`, and so although  there is now a version of that same `Plan` with an `EVERGREEN` final phase whose duration is unlimited, it now fails - and it should not.

sbrossie (commented):
This is going in the right direction, but there are still some issues:

1. In the test, since you try to upload the catalog twice, and expect the first one to fail and the second one to succeed, the counter `catalogUploaded` [here](https://github.com/tejinderghatore/killbill/blob/master/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogValidation.java#L77) is insufficient to ensure that first one failed and second one succeeded -- it could very well be the reverse and the test would not catch it.
2.  You have now fixed the validation when we upload the catalog to ensure that it detects the issue for #1465 even on the first version. However, [the validation you do](https://github.com/tejinderghatore/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/api/user/DefaultCatalogUserApi.java#L121) in this case isn't incomplete, it only checks for the issue on #1465 - instead we should run the full validation.

3. I would have hoped to also see more tests for the other use cases, not just (`EVERGREEN `), and also try to organize your catalog xml in a way that is easier to manage -- instead of introducing new versions, perhaps use different plans for different use cases so this is more maintainable as we add more validation use cases in the future.

4. Although this is not the most important, there are also still quite a few style issues, some places you are using tabs instead of space (e.g in `validatePlanDuration`), ...

sbrossie:
Closed, replaced with https://github.com/killbill/killbill/pull/1534

Issue #1528, opened 2021-11-12
Pending subscription shows a `PENDING` state after being canceled


sbrossie:
Scenario:
1. Create a subscription with an future start (billing) date -> `PENDING` state
2. Cancel the subscription immediately -> This will default to the future start date
3. Fetch the subscription and check the state -> `CANCELLED` state

It seems that in step 3. `state` is incorrectly showing `PENDING`

reshmabidikar:
When a `PENDING` subscription is cancelled on its exact effective start date, the state returned is a question of semantics.  Simply changing the behavior to return a `CANCELLED` state triggers an issue with `uncancel` subscription and causes the [TestUserApiCancel#testCancelUncancelFutureSubscription](https://github.com/killbill/killbill/blob/081be96c2250f2ad7d36c9383897199a8fde8838/subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCancel.java#L391) test to fail. So, the following approach was decided:

When a subscription is created with a future date:

1. Return state `PENDING` until the subscription effective start date
2. Once effective start date is reached, check if the subscription is future cancelled and if so return state as `CANCELLED`

PR #1530, opened 2021-11-17
Fix for 1528

reshmabidikar:


subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -178,12 +179,17 @@ public EntitlementState getState() {
         }
 
         final SubscriptionBaseTransition pendingTransition = getPendingTransition();
-        if (pendingTransition != null) {
-            return EntitlementState.PENDING;
+        if (pendingTransition != null && pendingTransition.getEffectiveTransitionTime().isAfter(clock.getUTCNow())) {
```
sbrossie, line 182: I would expect that if `pendingTransition != null` then, it is always in the future, because of how it was computed, i.e iterator looking through `transitions` with [FUTURE_ONLY](https://github.com/killbill/killbill/blob/master/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java#L317)

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -178,12 +179,17 @@ public EntitlementState getState() {
...
+        		return EntitlementState.PENDING;
         }
+        else if(pendingTransition.getEffectiveTransitionTime().isBefore(clock.getUTCNow()) && isFutureCancelled()){
```
sbrossie, line 185: ⚠️  One way to reach this statement if if `pendingTransition = null` and then we would NPE

However, based on previous point, I don't think we could have non null `pendingTransition` whose date is in the past.


Stepping back: Do you need to even change the code for your test to pass -- and knowing that we kept the semantics we already have?

sbrossie (changes requested):

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
reshmabidikar: @sbrossie, yes you are right, the code does not need to be changed for the test to pass. So, does that mean that [1528 ](https://github.com/killbill/killbill/issues/1528) is not really a bug and the system behaves as expected?

sbrossie: Yep, according to our semantics, this is not a bug: The main reason for choosing this semantics (this is a choice we have), is that we want the ability to uncancel (if needed) and if the state is `canceled` prior we reach the effective date, i.e the create/cancel date, we can't do it (hence your other test that was failing).

reshmabidikar: okay, that makes sense. I will revert my code changes. So, should we close #1528?

sbrossie: Let's complete this PR, and then as we merge the PR, you can close the ticket.

reshmabidikar: Reverted changes to `DefaultSubscriptionBase`

pierre (approved):
👍 

We should document the behavior (maybe in the Slate API docs?).

reshmabidikar:
Yes agreed, I will make a note and add this to the slate documentation the next time I update it.

sbrossie (approved):

PR #1535, opened 2021-12-03
Added test for 1533

reshmabidikar:


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
```diff
@@ -0,0 +1,103 @@
...
+        clock.setTime(new DateTime("2021-12-01T12:56:02"));
+        testCallContext = setupTenant();
+
+        // Setup account in right tenant
+        account = setupAccount(testCallContext);
+    }
+
+    @Test(groups = "slow", description = "https://github.com/killbill/killbill/issues/1533")
+    public void testFixedTermPhaseAndRecurringBillingPeriod() throws Exception {
+        uploadCatalog("fixedterm-and-biennial-billing.xml");
```
sbrossie, line 72: I would suggest to use a different name for your catalog file, so that if/when we add more `Plans` (in addition to the existing `p1-biennial`), the name still makes sense.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
```diff
@@ -0,0 +1,103 @@
...
+        assertListenerStatus();
```
sbrossie, line 73: Nit: I don't think this is needed - but I know we have existing tests that do that. The purpose of `assertListenerStatus()` is to `assert` that all expected events, as configured by `busHandler.pushExpectedEvents()`  have been received and nothing else. It's also fine to leave it, does not hurt...

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
```diff
@@ -0,0 +1,103 @@
...
+                                      NextEvent.PAYMENT);
+
+        final PlanPhaseSpecifier spec = new PlanPhaseSpecifier("p1-biennial", null);
+        final UUID entitlementId = entitlementApi.createBaseEntitlement(account.getId(), new DefaultEntitlementSpecifier(spec, null, UUID.randomUUID().toString(), null), "something", null, null, false, true, ImmutableList.<PluginProperty>of(), testCallContext);
+        assertNotNull(entitlementId);
+        assertListenerStatus();
+
+        invoiceChecker.checkInvoice(account.getId(), 1, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2021, 12, 01), new LocalDate(2023, 12, 01), InvoiceItemType.RECURRING, new BigDecimal("40.00")));
+
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.INVOICE_PAYMENT, NextEvent.PAYMENT); //TODO: Is this correct?
```
sbrossie, line 85: > //TODO: Is this correct?
Yes it is (and therefore you can remove the TODO): We expect one new invoice and along with it a (successful) payment so those 3 are expected.

Note: If you set breakpoints as indicated in the doc I recently shared, you will there is one additional bus event emitted by the system, namely a `org.killbill.billing.tenant.api.user.DefaultTenantConfigChangeInternalEvent` but this is one of the few event being ignored by our test framework.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
```diff
@@ -0,0 +1,103 @@
...
+        clock.addYears(2);
+        assertListenerStatus();
+
+        invoiceChecker.checkInvoice(account.getId(), 2, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2023, 12, 01), new LocalDate(2024, 12, 01), InvoiceItemType.RECURRING, new BigDecimal("20.03")));
```
sbrossie, line 89: The pro-ration computation is surprising, I would have expected an amount of `20.00` and not `20.03` - I suspect once the fix is in place this may trigger more questions... 😉

beatrix/src/test/resources/catalogs/testCatalogFixedTerm/fixedterm-and-biennial-billing.xml
```diff
@@ -0,0 +1,72 @@
...
+            </changePolicyCase>
+        </changePolicy>
+        <cancelPolicy>
+            <cancelPolicyCase>
+                <policy>END_OF_TERM</policy>
+            </cancelPolicyCase>
+        </cancelPolicy>
+    </rules>
+    <plans>
+        <plan name="p1-biennial">
```
sbrossie, line 44: We typically name our plans `<productNameInLowerCase>-<billingPeriod>-<optionalOtherStuff>` as shown [here](https://github.com/killbill/killbill/blob/master/beatrix/src/test/resources/catalogs/default/catalogTest.xml#L509) to make it easier to understand what they are.

sbrossie (approved):
Left a few minor comments you may want to address but looks great overall, happy to merge.
(Indentation looks good too).

Next steps: come up with more test cases

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
reshmabidikar: Fixed

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
reshmabidikar: Fixed

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
reshmabidikar: @sbrossie, would you prefer if additional `Plans` are added to the same catalog, or would you prefer creating different catalog files for each test as done for other tests within the [/src/test/resources/catalog](https://github.com/killbill/killbill/tree/master/beatrix/src/test/resources/catalogs) folder? If you would prefer adding multiple plans to the same file, I will go ahead and rename the catalog file.

beatrix/src/test/resources/catalogs/testCatalogFixedTerm/fixedterm-and-biennial-billing.xml
reshmabidikar: Understood, in this case, the product is named `p1` and hence the plan name is `p1-biennial`. Do you want the product to be renamed to something more explanatory?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
reshmabidikar: Are you saying that the "band-aid" fix that you put in place does not calculate the prorated amount correctly?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
sbrossie: I think it depends on the use cases - I would say, if you don't need multiple versions, stay out of it, and have multiple Product/Plans in one catalog version.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
sbrossie: Nope, unrelated to the current impl - it has to be do with the pro-ration logic. We have quite a few nice unit tests for this (example here](https://github.com/killbill/killbill/blob/master/invoice/src/test/java/org/killbill/billing/invoice/proRations/inAdvance/annual/TestTrailingProRation.java) but none for `biennial`, so based on time, perhaps something you could add. Let's skip for now, and have this merged and we can revisit - it would be nice to at least be able to explain 'why' we get this, not necessarily change it.

beatrix/src/test/resources/catalogs/testCatalogFixedTerm/fixedterm-and-biennial-billing.xml
sbrossie: My bad, what you have is great.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
reshmabidikar: Got it, I've now renamed the catalog and added more plans to it.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
reshmabidikar: okay, I will make a note of this and do some preliminary investigation.

beatrix/src/test/resources/catalogs/testCatalogFixedTerm/fixedterm-and-biennial-billing.xml
reshmabidikar: I've now renamed the product to "pistol" to make it more explanatory and also renamed the plans accordingly

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
```diff
@@ -0,0 +1,155 @@
...
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.INVOICE_PAYMENT, NextEvent.PAYMENT); 
+        clock.addYears(2);
+        assertListenerStatus();
+        invoiceChecker.checkInvoice(account.getId(), 1, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2021, 12, 01), new LocalDate(2023, 12, 01), InvoiceItemType.RECURRING, new BigDecimal("40.00")));
+        
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.INVOICE_PAYMENT, NextEvent.PAYMENT); 
+        clock.addYears(2);
+        assertListenerStatus();
+        invoiceChecker.checkInvoice(account.getId(), 2, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2023, 12, 01), new LocalDate(2024, 12, 01), InvoiceItemType.RECURRING, new BigDecimal("20.03")));
+
```
sbrossie, line 116: Could you check like you did pervious test that moving the clock again leads to no invoice - just for safety?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
```diff
@@ -0,0 +1,155 @@
...
+        assertNotNull(entitlementId);
+        assertListenerStatus();
+        invoiceChecker.checkInvoice(account.getId(), 1, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2021, 12, 01), new LocalDate(2023, 12, 01), InvoiceItemType.RECURRING, new BigDecimal("40.00")));
+
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.INVOICE_PAYMENT, NextEvent.PAYMENT); 
+        clock.addYears(2);
+        assertListenerStatus();
+        invoiceChecker.checkInvoice(account.getId(), 2, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2023, 12, 01), new LocalDate(2024, 12, 01), InvoiceItemType.RECURRING, new BigDecimal("20.03")));
+
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.INVOICE_PAYMENT, NextEvent.PAYMENT,NextEvent.PHASE,NextEvent.NULL_INVOICE); //TODO: not sure why NULL_INVOICE is required here but found it in the bus_events_history table
```
sbrossie, line 136: > //TODO: not sure why NULL_INVOICE is required here but found it in the bus_events_history table

👍 for checking and adding it.

This is because there are 2 'events' triggering an invoice run:
* There is a 'notification' (`notifications` table) with an `effective_date=2024-12-01 12:56:02` that gets triggered as you move the clock and that triggers an invoice run
* There is a bus event (`bus_events` table) for the `PHASE` that also gets triggered as you move the clock and that also triggers an invoice run - to be precise here there is also a notification that first gets processed and as a result triggers this bus event.

If you look at the `notifications` table prior moving the clock you would see: ```select * from notifications;
+-----------+---------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------+---------------------+----------------+------------------+---------------------------+------------------+-------------+-------------+-------------+------------------------------------------+---------------------+--------------------------------------+
| record_id | class_name                                                                | event_json                                                                                                                                                                | user_token                           | created_date        | creating_owner | processing_owner | processing_available_date | processing_state | error_count | search_key1 | search_key2 | queue_name                               | effective_date      | future_user_token                    |
+-----------+---------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------+---------------------+----------------+------------------+---------------------------+------------------+-------------+-------------+-------------+------------------------------------------+---------------------+--------------------------------------+
|         1 | org.killbill.billing.subscription.engine.core.SubscriptionNotificationKey | {"eventId":"f2e2be15-985a-48e0-9986-12683df2d2ee","seqId":0}                                                                                                              | 409632a6-0485-497c-9acf-b785536abdf1 | 2021-12-01 12:56:03 | Cocodrilo      | NULL             | NULL                      | AVAILABLE        |           0 |           1 |           1 | subscription-service:subscription-events | 2024-12-01 12:56:02 | 64ac62c2-4a25-4de9-85a7-9af8c086f68a |
|         3 | org.killbill.billing.invoice.notification.NextBillingDateNotificationKey  | {"uuidKey":null,"uuidKeys":["4d886e08-6db5-4cf3-a946-5c864370824e"],"targetDate":"2024-12-01T12:56:02.000Z","isDryRunForInvoiceNotification":false,"isRescheduled":false} | bf295f60-0b45-4a6c-b8a9-8bb6573ca2e1 | 2023-12-01 12:56:04 | Cocodrilo      | NULL             | NULL                      | AVAILABLE        |           0 |           1 |           1 | invoice-service:next-billing-date-queue  | 2024-12-01 12:56:02 | 7f366c37-5fa5-4600-8931-97cdfea6139a |
+-----------+---------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------+---------------------+----------------+------------------+---------------------------+------------------+-------------+-------------+-------------+------------------------------------------+---------------------+--------------------------------------+
```

So, one of these events will trigger the invoice and the next one will lead to a null invoice.


Note: you can remove your `TODO` if my explanation made sense

sbrossie, line 136: Nit: Typically we have an implicit ordering for expected events :subscription events first (e.g `PHASE`) then invoice, then payments. Just make it easier to read and check everything is there

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
reshmabidikar: Did you mean adding a call to  `checkNoMoreInvoiceToGenerate` ? I've added this now here as well as in the `testFixedTermPhaseInAdvanceBilling` test

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTerm.java
reshmabidikar: Understood, thanks for the detailed explanation. I've removed the `TODO`

reshmabidikar: Fixed

PR #1538, opened 2021-12-08
beatrix: fix DataSource leak in tests

pierre:
CI seems to be _pretty_ green again!

### DIFF
--- a/beatrix/pom.xml
+++ b/beatrix/pom.xml
@@ -64,6 +64,17 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <exclusions>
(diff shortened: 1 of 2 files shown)


beatrix/pom.xml
```diff
@@ -63,6 +63,17 @@
             <artifactId>guice</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <exclusions>
```
sbrossie, line 69: Should we have a scope `test` ?

sbrossie (approved):

beatrix/pom.xml
pierre: Technically, yes, but Maven complains about it (as it is a transitive dependency for another non-test library).

PR #1540, opened 2021-12-09
Fix 1407 recurring negative plan price

xsalefter:

PR #1544, opened 2021-12-10
Tests for versioned catalogs

reshmabidikar:


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTermWithVersionedCatalogs.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import org.joda.time.DateTime;
+import org.joda.time.LocalDate;
+import org.killbill.billing.account.api.Account;
+import org.killbill.billing.api.TestApiListener.NextEvent;
+import org.killbill.billing.beatrix.util.InvoiceChecker.ExpectedInvoiceItemCheck;
+import org.killbill.billing.catalog.api.BillingPeriod;
```
sbrossie, line 15: This import (`BillingPeriod`) and a few others are not used, could you clean them -- same on your other test `TestCatalogFixedTerm`

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTermWithVersionedCatalogs.java
```diff
@@ -0,0 +1,132 @@
...
+        
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.PAYMENT, NextEvent.INVOICE_PAYMENT);
+        clock.addDays(15); //2021-03-01
+        assertListenerStatus();
+        invoiceChecker.checkInvoice(account.getId(), 4, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2021, 03, 01), new LocalDate(2021, 04, 01), InvoiceItemType.RECURRING, new BigDecimal("60.00")));
+        
+        busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.PAYMENT, NextEvent.INVOICE_PAYMENT);
+        clock.addMonths(1); //2021-04-01
+        assertListenerStatus();
+        invoiceChecker.checkInvoice(account.getId(), 5, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2021, 04, 01), new LocalDate(2021, 04, 11), InvoiceItemType.RECURRING, new BigDecimal("20.00")));
```
sbrossie, line 92: Perhaps add a comment that 100 days (length of the fixed term) from Jan 1st leads to April 11 (year 2021)

sbrossie (commented):
Good job on writing the test cases - this indeed checks that catalog transition events are not lost and the expiration date is also correctly reflected.

I think currently your build fails because you are missing copyright info for your new file `TestCatalogFixedTermWithVersionedCatalogs` -- therefore the annoying `rat` plugin makes the build fail. I also noticed that coms of your copyrights on new files show old dates ` Copyright 2014-2020 The Billing Project, LLC` -> but we are already in 2021 and soon moving in 2022. In Intellij we have a template file `.idea/copyright/apache.xml` that we can activate one new file and that computes the right date automatically. Perhaps there is something similar in Eclipse -- if so, please update the wiki with this info the rest of the team.

Also included a few inline comments.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTermWithVersionedCatalogs.java
reshmabidikar: Fixed

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTermWithVersionedCatalogs.java
reshmabidikar: Fixed

reshmabidikar:
Added copyright information, also found a way to automatically add this via Eclipse, I will update the wiki with this information.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTermWithVersionedCatalogs.java
```diff
@@ -123,6 +134,7 @@ public void testVersionedCatalogsEffectiveDateAfterEndOfFixedTerm() throws Excep
         busHandler.pushExpectedEvents(NextEvent.INVOICE, NextEvent.PAYMENT, NextEvent.INVOICE_PAYMENT);
         clock.addMonths(1); //2021-04-01
         assertListenerStatus();
+        //Fixed term duration of 100 days ends on 2021-04-11
         invoiceChecker.checkInvoice(account.getId(), 4, testCallContext, new ExpectedInvoiceItemCheck(new LocalDate(2021, 04, 01), new LocalDate(2021, 04, 11), InvoiceItemType.RECURRING, new BigDecimal("13.33")));
         
         checkNoMoreInvoiceToGenerate(account.getId(), testCallContext);
```
reshmabidikar, line 140: Just a general query, the `testVersionedCatalogsEffectiveDateAfterEndOfFixedTerm`  tests the `plumber-insurance-monthly2` plan. In this plan, the `effectiveDateForExistingSubscriptions` is `2021-04-12` which is at the end of the fixed-term duration. Wondering how such a catalog does not make sense. I mean, what use would a plan whose `effectiveDate`  is after the end of the fixed-term duration be?

sbrossie: 2 observations:
1. You wrote the test `testVersionedCatalogsEffectiveDateAfterEndOfFixedTerm`, so you tell me :-) ?
2.  `In this plan, the effectiveDateForExistingSubscriptions is 2021-04-12 which is at the end of the fixed-term duration.` -> It depends when you start your subscription - in your test, this is indeed useless, but I thought that's what you wanted to verify. But if you had started the subscription later, things could have been different, e.g your previous scenario `testVersionedCatalogsEffectiveDateBetweenFixedTerm `

beatrix/src/test/resources/catalogs/testCatalogFixedTerm/versionedCatalogs/WaterUtility-v1.xml
```diff
@@ -0,0 +1,106 @@
...
+            </cancelPolicyCase>
+        </cancelPolicy>
+    </rules>
+
+    <plans>
+     <plan name="plumber-insurance-monthly">
+            <product>PlumberInsurance</product>
+            <initialPhases>
+            </initialPhases>
+            <finalPhase type="FIXEDTERM">
```
sbrossie, line 61: The xml are all misaligned - perhaps the one you copied from where already like that? In any case, if this is easy, can you reformat them - there are probably online formatter you could use or perhaps a plugin in eclipse?

sbrossie (approved):
I just merged PR `https://github.com/killbill/killbill/pull/1540`, so could you pull latest master, and then if tests are green, we are good to merge.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTermWithVersionedCatalogs.java
reshmabidikar: Agreed, makes sense. I wrote the test because I found that in order for the `https://github.com/killbill/killbill/blob/fix-attempt-fix-term/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java#L674` to be executed, the `effectiveDateForExistingSubscriptions` should be after the fixed term period. I guess what I was really asking is what real-world situation would lead to the code at `https://github.com/killbill/killbill/blob/fix-attempt-fix-term/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java#L674` to be executed since such a plan (which we agree is useless) would never be used in the real world. Anyway never mind, it's not really relevant here, so don't want to take your time :)

reshmabidikar:
So, do you want me to merge the changes from `master` into the `fix-attempt-fix-term` branch?

beatrix/src/test/resources/catalogs/testCatalogFixedTerm/versionedCatalogs/WaterUtility-v1.xml
reshmabidikar: Fixed, sorry about that.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestCatalogFixedTermWithVersionedCatalogs.java
sbrossie: Yes, I understand your point, you are correct: If the FIXEDTERM subscription expires prior the catalog change happens, returning such catalog change billing event is a no-op, it does not hurt, but also does not serve any purpose as it is never used.

PR #1545, opened 2021-12-10
usage support proration

wishma2013:
usage support proration in billingPeriod:MONTHLY & usageType:CONSUMABLE & tierBlockPolicy:ALL_TIERS

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
@@ -70,9 +70,9 @@ import com.google.common.collect.Multimap;
 import com.google.common.collect.Range;
 import com.google.inject.Inject;
 
+import static org.killbill.billing.invoice.generator.InvoiceDateUtils.adjForCalendarMonth;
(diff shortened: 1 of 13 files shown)


sbrossie:
@wishma2013 We typically don't accept feature PR without first discussing the use case. Could you please provide a detailed description of the use case you are trying to solve?

wishma2013:
> @wishma2013 We typically don't accept feature PR without first discussing the use case. Could you please provide a detailed description of the use case you are trying to solve?

Ok.When occur some usage of a subscripted BASE plan, sometimes the BASE plan billing is RECURRING in MONTH, and so is the usage.That is if BASE plan was not subscripted at the billing day, the attendant usage charging will only count the period from subscribe day to next billing day, because the usage we supply to customer is for the whole billing month.That's what we are facing in our business.

sbrossie:
@wishma2013 Unfortunately I am not following the use case: If you could provide something more detailed, e.g with specific dates, how you record your usage points, etc... and highlight the issue based on the scenario that would be useful.

Also prior any PR merge 
* There are lots of conflicting files that need to be resolved
* You are touching some sensitive pieces of code, and yet there is no additional test cases for it -- this needs to be added
* All our regression tests (at least `mvn test -Pmysql` would need to run without error)

sbrossie:
Closing as I don't have enough context on what this PR is supposed to address - see https://github.com/killbill/killbill/pull/1545#issuecomment-998341981

Issue #1497, opened 2021-09-02
Fix issue in loading overdue.xml that uses totalUnpaidInvoiceBalanceEqualsOrExceeds condition


reshmabidikar:
There is an issue in loading an overdue.xml file that has a `totalUnpaidInvoiceBalanceEqualsOrExceeds ` condition. The problematic `overdue.xml` is attached.
[bad_overdue.zip](https://github.com/killbill/killbill/files/7097226/bad_overdue.zip)

Steps to reproduce:

1. Create a new tenant
2. Upload attached overdue.xml
3. Click the **+** sign in the **Overdue Show** tab  to upload overdue configuration again
4. This results in an `Error while communicating with the Kill Bill server:` error in Kaui and a `NullPointerException` in Kill Bill logs.

xsalefter:
Ok so the `NullPointerException` thrown **not** because of `<totalUnpaidInvoiceBalanceEqualsOrExceeds />`, but because an XML in `BadOverdue.xml` sample above has no `<timeSinceEarliestUnpaidInvoiceEqualsOrExceeds />` attribute in its `<condition />`.

Actually, when you doing point 2 (Uploading `BadOverdue.xml`), the file uploaded but throw NPE immediatelly when marshalling (Bad) `OverdueConfig` into JSON. 

So, `+` sign clicked or not, NPE still thrown when loading the page, because of a problem I explained above.

There's several test case for this:
- Test [that simulate](https://github.com/killbill/killbill/blob/68a011f266e49f9a0f8630556d7f759bedb1fcb1/jaxrs/src/test/java/org/killbill/billing/jaxrs/resources/TestOverdueResource.java#L82) `BadOverdue.xml` .
- There's [another test](https://github.com/killbill/killbill/blob/68a011f266e49f9a0f8630556d7f759bedb1fcb1/jaxrs/src/test/java/org/killbill/billing/jaxrs/resources/TestOverdueResource.java#L116) that if I adding `<timeSinceEarliestUnpaidInvoiceEqualsOrExceeds />` element, the test passed.
[...]

PR #1548, opened 2021-12-16
Fix 1497 fix issue in loading overdue xml

xsalefter:
This [commit](https://github.com/killbill/killbill/pull/1548/commits/a6d864d3dc03f9e4b351ce20ec15d6d6efc34dce) add test case for #1497 , and [this commit](https://github.com/killbill/killbill/pull/1548/commits/85f38a5021a642772c5067ada65b3f995a276777) is my suggestion how to fix it.

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/OverdueConditionJson.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/OverdueConditionJson.java
@@ -19,6 +19,7 @@ package org.killbill.billing.jaxrs.json;
 
 import java.math.BigDecimal;
 
+import org.killbill.billing.catalog.api.TimeUnit;
 import org.killbill.billing.jaxrs.json.CatalogJson.DurationJson;
 import org.killbill.billing.overdue.api.OverdueCondition;
 import org.killbill.billing.overdue.config.DefaultDuration;
@@ -57,7 +58,9 @@ public class OverdueConditionJson {
     }
 
     public OverdueConditionJson(final OverdueCondition overdueCondition) {
-        this.timeSinceEarliestUnpaidInvoiceEqualsOrExceeds = new DurationJson(overdueCondition.getTimeSinceEarliestUnpaidInvoiceEqualsOrExceeds());
+        this.timeSinceEarliestUnpaidInvoiceEqualsOrExceeds = overdueCondition.getTimeSinceEarliestUnpaidInvoiceEqualsOrExceeds() == null ?
+                                                             new DurationJson(TimeUnit.DAYS, 1) :
+                                                             new DurationJson(overdueCondition.getTimeSinceEarliestUnpaidInvoiceEqualsOrExceeds());
         this.controlTagInclusion = overdueCondition.getInclusionControlTagType();
         this.controlTagExclusion = overdueCondition.getExclusionControlTagType();
         this.numberOfUnpaidInvoicesEqualsOrExceeds = overdueCondition.getNumberOfUnpaidInvoicesEqualsOrExceeds();
--- a/overdue/src/test/java/org/killbill/billing/overdue/config/TestOverdueConfig.java
+++ b/overdue/src/test/java/org/killbill/billing/overdue/config/TestOverdueConfig.java
@@ -19,7 +19,12 @@ package org.killbill.billing.overdue.config;
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 
-import org.killbill.billing.overdue.api.EmailNotification;
+import org.joda.time.Period;
+import org.killbill.billing.catalog.api.Duration;
+import org.killbill.billing.overdue.api.OverdueCondition;
+import org.killbill.billing.overdue.api.OverdueConfig;
+import org.killbill.billing.overdue.api.OverdueState;
+import org.killbill.billing.overdue.api.OverdueStatesAccount;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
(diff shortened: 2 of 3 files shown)


overdue/src/test/java/org/killbill/billing/overdue/config/TestOverdueConfig.java
```diff
@@ -75,4 +80,33 @@ public void testParseConfig() throws Exception {
 
         Assert.assertEquals(c.getOverdueStatesAccount().getFirstState().getName(), "OD1");
     }
+
+    @Test(groups = "fast", description = "https://github.com/killbill/killbill/issues/1497")
+    public void testBadOverdueXml() throws Exception {
+        final OverdueConfig overdueConfig = getOverdueConfig("BadOverdue.xml");
```
reshmabidikar, line 86: I would suggest renaming the test to `testTotalUnpaidInvoiceBalanceEqualsOrExceedsCondition`. Also, I think it would be better to create a different class for this test, but I will let @sbrossie confirm

overdue/src/test/resources/org/killbill/billing/overdue/BadOverdue.xml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
reshmabidikar, line 1: Again, I would suggest renaming this to something like `OverdueWithTotalUnpaidInvoiceBalanceCondition`

overdue/src/test/java/org/killbill/billing/overdue/config/TestOverdueConfig.java
```diff
@@ -75,4 +80,33 @@ public void testParseConfig() throws Exception {
...
+
+            Assert.assertNotNull(condition.getTotalUnpaidInvoiceBalanceEqualsOrExceeds());
+
+            final Duration duration = condition.getTimeSinceEarliestUnpaidInvoiceEqualsOrExceeds();
+            Assert.assertNull(duration);
+
+            // Thus, calling method in duration will throw NullPointerException
+        }
+    }
 }
```
reshmabidikar, line 112: I'm thinking whether it would be good to add some more tests for other overdue configuration files that have some of the other conditions like `timeSinceEarliestUnpaidInvoiceEqualsOrExceeds`, just to confirm that this fix has not caused any regression issues? @sbrossie, what do you think?

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/OverdueConditionJson.java
```diff
@@ -57,7 +58,9 @@ public OverdueConditionJson(@JsonProperty("timeSinceEarliestUnpaidInvoiceEqualsO
     }
 
     public OverdueConditionJson(final OverdueCondition overdueCondition) {
-        this.timeSinceEarliestUnpaidInvoiceEqualsOrExceeds = new DurationJson(overdueCondition.getTimeSinceEarliestUnpaidInvoiceEqualsOrExceeds());
+        this.timeSinceEarliestUnpaidInvoiceEqualsOrExceeds = overdueCondition.getTimeSinceEarliestUnpaidInvoiceEqualsOrExceeds() == null ?
```
sbrossie, line 61: ⚠️ I think checking the null value of `overdueCondition#getTimeSinceEarliestUnpaidInvoiceEqualsOrExceeds()` makes a lot of sense otherwise the code will NPE inside the `DurationJson ` CTOR; however, I don't think providing a default of one day is good, you are essentially changing the condition. Instead, I think we should default to `this.timeSinceEarliestUnpaidInvoiceEqualsOrExceeds = null` (but to be tested).

overdue/src/test/java/org/killbill/billing/overdue/config/TestOverdueConfig.java
sbrossie: > I would suggest renaming the test to testTotalUnpaidInvoiceBalanceEqualsOrExceedsCondition
Agreed.

And also change the name of the test, because after your fix this is **not** a 'bad' overdue XML config ...

> Also, I think it would be better to create a different class for this test, but I will let @sbrossie confirm

We could go either way, since we already have a Test class for checking xml config, adding one test inside is fine with me.

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/OverdueConditionJson.java
sbrossie: Also, I am wondering what triggered this change - if you run your test below, the code never goes though this code as the `jaxrs` layer is our api/endpoint layer, and any test inside the overdue module will not exercise this code.

overdue/src/test/java/org/killbill/billing/overdue/config/TestOverdueConfig.java
```diff
@@ -75,4 +80,33 @@ public void testParseConfig() throws Exception {
...
+
+            final OverdueCondition condition = state.getOverdueCondition();
+            Assert.assertNotNull(condition);
+
+            Assert.assertNotNull(condition.getTotalUnpaidInvoiceBalanceEqualsOrExceeds());
+
+            final Duration duration = condition.getTimeSinceEarliestUnpaidInvoiceEqualsOrExceeds();
+            Assert.assertNull(duration);
+
+            // Thus, calling method in duration will throw NullPointerException
```
sbrossie, line 109: If I run this test, it passes for me - is that not expected?

sbrossie (changes requested):

xsalefter:
I'll going to close this and put more context in issue comment and more accurate test in #1549 .

PR #1560, opened 2022-01-12
beatrix: Add invoice optimization tests (bill run scenario)

sbrossie:


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoiceOptimization.java
```diff
@@ -789,6 +790,181 @@ public void testBillRunInArrearLeadingProration() throws Exception {
     }
 
 
+    @Test(groups = "slow")
+    public void testBillRunInArrearWithUsageAndRecurring() throws Exception {
+
+        // Usage records are pulled using the most aggressive cuttoffDt
```
pierre, line 796: 👏 for these plain English comments...

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoiceOptimization.java
```diff
@@ -789,6 +790,181 @@ public void testBillRunInArrearLeadingProration() throws Exception {
...
+        recordUsageData(sub2.getId(), "tracking-old-1", "hours", new LocalDate(2020, 9, 19), 1L, callContext);
+        recordUsageData(sub2.getId(), "tracking-old-2", "hours", new LocalDate(2020, 10, 19), 1L, callContext);
+        recordUsageData(sub2.getId(), "tracking-old-3", "hours", new LocalDate(2020, 11, 19), 1L, callContext);
+        recordUsageData(sub2.getId(), "tracking-old-4", "hours", new LocalDate(2020, 12, 19), 1L, callContext);
+
+        // Generate usage data for this month
+        recordUsageData(sub2.getId(), "tracking-1", "hours", new LocalDate(2021, 1, 19), 1L, callContext);
+
+        // 2021-02-01
+        // (Hum... Interestingly, there is no future notification set on the 1st although sub2#BCD is the 1st...)
```
pierre, line 849: Were you expecting a notification for an invoice run?

When we process the initial subscription events, we return early because of `AUTO_INVOICING_OFF` https://github.com/killbill/killbill/blob/3a172831303caccdb0e16d0091d660a97089d8fa/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L367

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoiceOptimization.java
```diff
@@ -789,6 +790,181 @@ public void testBillRunInArrearLeadingProration() throws Exception {
...
+        clock.addDays(17);
+        Thread.sleep(1000);
```
pierre, line 851: That's just to ensure there are no rogue events? Couldn't you just call `assertListenerStatus()` instead?

pierre (approved):

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoiceOptimization.java
sbrossie: I would expect to see an entry in the notificationQ for `2021-02-01`, because there is something to invoice for that date (the usage subscription is aligned on the 1st). However the notification does not exist (-> I think this is a bug). However, because of `AUTO_INVOICING_OFF` I do not expect a `NULL_INVOICE` (as you said)

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoiceOptimization.java
sbrossie: > That's just to ensure there are no rogue events? 
yes

> Couldn't you just call assertListenerStatus() instead?
I don't think so, since i did not push any events to expect `busHandler.pushExpectedEvents` there is nothing to wait for, it would go straight. 

I think I could remove these sleeps (it just makes the test more sturdy, and makes it more clear nothing bis happening (but takes longer to run)

Issue #1469, opened 2021-08-05
Usage amount should be a floating point not a long 


sbrossie:
This is a GH consolidation from https://github.com/killbill/killbill-api/issues/76

Our usage models define passing the 'amount' of units as a long, as shown in [RolledUpUnit#amount](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/usage/api/RolledUpUnit.java#L25) and [UsageRecord#amount](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/usage/api/UsageRecord.java#L27). However, there are many cases where we want to define a high granularity unit (e.g GBytes instead of MBytes or even KBytes, ...) for invoice presentation, and as a result the usage quantity should allow for 'decimal' values. 

However, we don't want to use decimal types such as `float` or `double` because of the loss of precision. Instead, we suggest to use `BigDecimal`.

sbrossie:
Also see https://github.com/killbill/killbill/issues/1325

sbrossie:
The work will require changing our public api defined in `https://github.com/killbill/killbill-api/` and `https://github.com/killbill/killbill-plugin-api` (plugin spi), and therefore will need to be done in our next feature branch `work-for-release-0.23.x`.

It will also require updating our schema to store this new type. Instead of using a `bigint` as shown [here](`https://github.com/killbill/killbill/blob/master/usage/src/main/resources/org/killbill/billing/usage/ddl.sql#L10`), we want to use a numeric type that is big enough, e.g `numeric(18,9)`.

sbrossie:
As a first step, I would suggest to just change the [database schema](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/usage/src/main/resources/org/killbill/billing/usage/ddl.sql#L10) in Kill Bill and make sure everything works.

xsalefter:
Just for note: As [discussed in this PR](https://github.com/killbill/killbill/pull/1579#pullrequestreview-871196271), there are other APIs in `Catalog` that need to be updated as well:

```
    TieredBlock#getMax
    TieredBlockPriceOverride#getMax
    Block#getSize
    Block#getMinTopUpCredit
    BlockPriceOverride#getSize
    Limit#getMax
    Limit#getMin
```

shaun-forgie:
Any chance we could also include #1756 into the enhancements for usage API?

PR #1561, opened 2022-01-13
Fix 1469 killbill usage amount should be a floating point

xsalefter:
All commit only fix for compile-error because of changes in `RolledUpUnit`, `UsageRecord` and `RawUsageRecord`. There are  `//FIXME-1469 ....` across source code on this PR that indicate the line will be changed when other API (like invoice API) that affected by original issue in #1469 fixed.

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionUsageRecordJson.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionUsageRecordJson.java
@@ -101,11 +101,11 @@ public class SubscriptionUsageRecordJson {
     public static class UsageRecordJson {
 
         private final LocalDate recordDate;
-        private final Long amount; // FIXME-1469 change to correct BigDecimal implementation
+        private final BigDecimal amount;
 
         @JsonCreator
         public UsageRecordJson(@JsonProperty("recordDate") final LocalDate recordDate,
-                               @JsonProperty("amount") final Long amount) {
+                               @JsonProperty("amount") final BigDecimal amount) {
             this.recordDate = recordDate;
             this.amount = amount;
         }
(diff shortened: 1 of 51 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionUsageRecordJson.java
```diff
@@ -101,7 +101,7 @@ public UsageRecord apply(final UsageRecordJson input) {
     public static class UsageRecordJson {
 
         private final LocalDate recordDate;
-        private final Long amount;
+        private final Long amount; // FIXME-1469 change to correct BigDecimal implementation
```
sbrossie, line 104: Is there anything that prevents making this change now?

sbrossie (commented):
Good job overall 👏 


I re-parented your PR against `work-for-1533-0.23.x`. `work-for-1533-0.23.x` is the same as `work-for-release-0.23.x`, but it will provide an integration branch where we can keep merging your intermediate PRs (like this one) until we have everything, and then we can create a PR from `work-for-1533-0.23.x` -> `work-for-release-0.23.x` to give us a chance to review the whole thing (*)

(*) In particular it will allow us to verify that when everything is in place, there is no more conversion such as  `amount.longValue()` or `BigDecimal.valueOf(amount)` as all the apis have also been updated.


In terms of next steps:
* Before we can merge this PR:
  * I left one comment where I think we can make the change but if you think differently, let me know. 
  * You should rebase your branch with  `work-for-1533-0.23.x` as there were quite a few updates that you should get - in particular, we integrated a lot of Vijay's work and updated the parent `pom.xml`, ...
* Once this has been merged, you should setup your env to pull all the repos you need (`work-for-release-0.23.x` branch) and 1/ build locally and 2/ start submitting PRs in these repos to include the api changes we need. This step will basically be a lot of _undo_ operation to remove all the type conversions we currently have in this PR.

PR #1565, opened 2022-01-19
junction, subscription: Propagate invoice optimization cutoffDt to ju…

sbrossie:
…nction

The goal is to filter active subscriptions (not cancelled) to avoid returning too much state.

Note that, this is not this end of the story: A big piece missing is the filtering of the blocking states - anybody with good 💡, please let me know...


subscription/src/main/resources/org/killbill/billing/subscription/engine/dao/SubscriptionSqlDao.sql.stg
```diff
@@ -51,6 +51,20 @@ where bundle_id = :bundleId
 ;
 >>
 
+getActiveByAccountRecordId() ::= <<
+select
+<allTableFields("s.")>
+from subscriptions s
+left join subscription_events se on se.subscription_id = s.id and se.user_type = 'CANCEL' and se.is_active
+where
+(se.id IS NULL or se.effective_date \> :cutoffDt)
```
pierre, line 60: `>=`?

pierre (approved):
Easy enough! 😅

subscription/src/main/resources/org/killbill/billing/subscription/engine/dao/SubscriptionSqlDao.sql.stg
sbrossie: Note that:
* `cutoffDt` is a localdate, e.g `2022-01-18`
* `effective_date` is a full time date, e.g `2022-01-18 00:00:00`

However, the query code seems to be smart and treat `2022-01-18 00:00:00` as `2022-01-18`, meaning with these 2 dates, the subscription would not be returned. So the question is whether we want to return a subscription cancelled exactly on the `0` second of the bill run - knowing no invoice item would be generated for it?

pierre: > knowing no invoice item would be generated for it?
What about the case where we bill in arrear? Don't we need that subscription?

pierre:
👍

PR #1573, opened 2022-01-26
server: support for encoded audit headers

pierre:
Support utf-8 characters in audit headers by allowing them to be sent encoded.

The header `X-Killbill-Encoding` must be specified (only base64 is supported at the moment).

### DIFF
--- a/profiles/killbill/src/main/java/org/killbill/billing/server/filters/RequestDataFilter.java
+++ b/profiles/killbill/src/main/java/org/killbill/billing/server/filters/RequestDataFilter.java
@@ -18,8 +18,10 @@
 package org.killbill.billing.server.filters;
 
 import java.io.OutputStream;
+import java.util.Base64;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiFunction;
 
 import javax.inject.Singleton;
 import javax.ws.rs.container.ContainerRequestContext;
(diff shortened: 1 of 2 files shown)


profiles/killbill/src/main/java/org/killbill/billing/server/filters/RequestDataFilter.java
```diff
@@ -31,22 +33,61 @@
...
+
     @Override
     public void filter(final ContainerRequestContext requestContext) {
         final List<String> requestIdHeaderRequests = getRequestId(requestContext);
         final String requestId = (requestIdHeaderRequests == null || requestIdHeaderRequests.isEmpty()) ? UUIDs.randomUUID().toString() : requestIdHeaderRequests.get(0);
         Request.setPerThreadRequestData(new RequestData(requestId));
+
+        final List<String> kbEncodingHeaders = requestContext.getHeaders().get(KILL_BILL_ENCODING_HEADER);
+        final String kbEncodingHeader = (kbEncodingHeaders == null || kbEncodingHeaders.isEmpty()) ? null : kbEncodingHeaders.get(0);
+        if (kbEncodingHeader != null) {
```
sbrossie, line 61: Should not we check the value here, i.e 'Base64' ?

sbrossie (approved):

profiles/killbill/src/main/java/org/killbill/billing/server/filters/RequestDataFilter.java
pierre: It's done line 73: https://github.com/killbill/killbill/pull/1573/files/404496a97513d3f004998eeabc2e6221a077a097#diff-0f99ec375e3ac23eab1c6990a24a66790c9ebeaa27f242d4317883edced53eedR73

sbrossie: Yep, my point is that you probably did not need to do it each, just one time here would have been enough.

pierre: Ah, right -- I was anticipating the case where we support multiple schemes down the line (so need to find the right decoder each time) 😄

PR #1583, opened 2022-02-03
ChargeThroughDate Optimizations

sbrossie:
Ability to make one call to set all the CTDs.

The code also simplify the CTD logic by removing unused code in d4d12979a1e

### DIFF
--- a/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestAccountTimeline.java
+++ b/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestAccountTimeline.java
@@ -275,10 +275,8 @@ public class TestAccountTimeline extends TestJaxrsBase {
                 verifyAuditLog(subscriptionEvent2AuditLogs.get(0), ChangeType.INSERT, reason, comment, createdBy, startTime, endTime);
             } else {
                 // Audits for bundles
-                Assert.assertEquals(bundleAuditLogs.size(), 3);
+                Assert.assertEquals(bundleAuditLogs.size(), 1);
                 verifyAuditLog(bundleAuditLogs.get(0), ChangeType.INSERT, reason, comment, createdBy, startTime, endTime);
-                verifyAuditLog(bundleAuditLogs.get(1), ChangeType.UPDATE, null, null, TRANSITION, startTime, endTime);
-                verifyAuditLog(bundleAuditLogs.get(2), ChangeType.UPDATE, null, null, TRANSITION, startTime, endTime);
 
                 // Audits for subscriptions
                 Assert.assertEquals(subscriptionAuditLogs.size(), 3);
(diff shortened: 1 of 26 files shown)


profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestAccountTimeline.java
```diff
@@ -275,10 +275,8 @@ private void verifyBundles(final UUID accountId, final DateTime startTime, final
                 verifyAuditLog(subscriptionEvent2AuditLogs.get(0), ChangeType.INSERT, reason, comment, createdBy, startTime, endTime);
             } else {
                 // Audits for bundles
-                Assert.assertEquals(bundleAuditLogs.size(), 3);
+                Assert.assertEquals(bundleAuditLogs.size(), 1);
                 verifyAuditLog(bundleAuditLogs.get(0), ChangeType.INSERT, reason, comment, createdBy, startTime, endTime);
-                verifyAuditLog(bundleAuditLogs.get(1), ChangeType.UPDATE, null, null, TRANSITION, startTime, endTime);
```
pierre, line 280: Do you understand why *all* updates have disappeared? I would have expected the code change to merge some, but not to make all of them disappear (1 update instead of multiple ones).

pierre (approved):

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/SubscriptionSqlDao.java
```diff
@@ -55,4 +57,10 @@ public void updateChargedThroughDate(@Bind("id") String id,
                                          @Bind("chargedThroughDate") Date chargedThroughDate,
                                          @SmartBindBean final InternalCallContext context);
 
+    @SqlUpdate
+    @Audited(ChangeType.UPDATE)
```
pierre, line 61: FYI I've checked that we already seem to support audited batch calls: https://github.com/killbill/killbill/blob/585d68891b06f3329379548583a33fe94b9250ea/util/src/main/java/org/killbill/billing/util/entity/dao/EntitySqlDaoWrapperInvocationHandler.java#L435

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestAccountTimeline.java
sbrossie: Yep, I understand, this is because we don't update the [bundle#lastSysUpdateDate](https://github.com/killbill/killbill/pull/1583/files#diff-fa76b2b89459501b1c3442c56007ecc4a5c6f0aee2085469347119267b73e6e4L523)

sbrossie: As mentioned in my [comment](https://github.com/killbill/killbill/pull/1583/commits/d4d12979a1e204aa1a73872e905d7efb213debcb) we should even remove the column in 0.23.x

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/SubscriptionSqlDao.java
sbrossie: Am i missing a `@BatchChunkSize` annotation?

PR #1587, opened 2022-02-03
server: introduce org.killbill.jaxrs.get.ro property

pierre:
When `org.killbill.jaxrs.get.ro=false` (non default), GET calls will not attempt to use the RO db connection pool, even if
configured: only plugins forcing a RO connection (via the use of `ROTenantContext`) would use it.


sbrossie (approved):

util/src/main/java/org/killbill/billing/util/config/definition/JaxrsConfig.java
```diff
@@ -49,4 +49,9 @@
     @DefaultNull
     @Description("Base host address to use for redirect URLs")
     String getJaxrsLocationHost();
+
+    @Config("org.killbill.jaxrs.get.ro")
```
pierre, line 53: @reshmabidikar Juste a note to remember to update our config docs with this new property.

PR #1588, opened 2022-02-04
invoice: Add new config InvoiceConfig#isUsageMissingLenient

sbrossie:
With this new config, we prevent failing invoice generation when past usage records seem missing


util/src/main/java/org/killbill/billing/util/config/definition/InvoiceConfig.java
```diff
@@ -72,6 +72,16 @@
     @Description("Whether we disable writing $0 usage amounts")
     boolean isUsageZeroAmountDisabled(@Param("dummy") final InternalTenantContext tenantContext);
 
+    @Config("org.killbill.invoice.usage.missing.lenient")
```
pierre, line 75: @reshmabidikar Juste a note to remember to update our config docs with this new property.

pierre (approved):

PR #1592, opened 2022-02-08
Changes for inserting notification/subscription events

reshmabidikar:


subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
```diff
@@ -512,6 +515,14 @@ private void doChangePlan(final DefaultSubscriptionBase subscription,
         if (nextPhaseEvent != null) {
             events.add(nextPhaseEvent);
         }
+        if(nextTimedPhase == null && currentPhase.getPhase().getPhaseType() == PhaseType.FIXEDTERM) {
```
sbrossie, line 518: Suggestion: Instead of `if(nextTimedPhase == null && currentPhase.getPhase().getPhaseType() == PhaseType.FIXEDTERM) {` you could just leverage the existing `if` and have a `else if ( currentPhase.getPhase().getPhaseType() == PhaseType.FIXEDTERM)`.

Nit: Also formatting-wise we typically have a space between `if` and `(`

subscription/src/main/java/org/killbill/billing/subscription/engine/core/DefaultSubscriptionBaseService.java
```diff
@@ -209,13 +213,20 @@ private boolean onPhaseEvent(final DefaultSubscriptionBase subscription, final S
                                               null;
             if (nextPhaseEvent != null) {
                 dao.createNextPhaseEvent(subscription, readyPhaseEvent, nextPhaseEvent, context);
-                return true;
+                phaseEventCreated = true;
             }
-        } catch (final SubscriptionBaseError e) {
+            
+            if(subscription.getCurrentPhase().getPhaseType() == PhaseType.FIXEDTERM) {
```
sbrossie, line 219: Shouldn't the logic be similar to what you had in `DefaultSubscriptionBaseApiService#getEventsOnCreation`, i.e:
1. If there is a next phase then create next phase event
2. Else if (this is the last phase) and this is a `FIXEDTERM` then insert the EXPIRED event

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
```diff
@@ -538,6 +538,29 @@ public Void inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFa
...
+                cancelNextPhaseEventFromTransaction(subscriptionId, entitySqlDaoWrapperFactory, context);
+                createAndRefresh(transactional, new SubscriptionEventModelDao(nextPhaseEvent), context);
+                recordFutureNotificationFromTransaction(entitySqlDaoWrapperFactory,
+                                                        nextPhaseEvent.getEffectiveDate(),
+                                                        new SubscriptionNotificationKey(nextPhaseEvent.getId()), context);
+
+                // Notify the Bus
+                notifyBusOfRequestedChange(entitySqlDaoWrapperFactory, subscription, nextPhaseEvent, SubscriptionBaseTransitionType.EXPIRED, 0, context);
+                if(!phaseEventCreated)//TODO_1533: I'm not completely sure passing the phaseEventCreated flag as a parameter to this method and using it like this is required. The only condition I can think of where this would be required is when there is a a phase following the FIXEDTERM case (to prevent the bus event from being triggered twice). So, I'm thinking it is better to keep this code just to be on the safe side. If this is a common scenario, I will write a test for this.
```
sbrossie, line 557: > I'm not completely sure passing the phaseEventCreated flag as a parameter to this method and using it like this is required. The only condition I can think of where this would be required is when there is a a phase following the FIXEDTERM case (to prevent the bus event from being triggered twice). So, I'm thinking it is better to keep this code just to be on the safe side. If this is a common scenario, I will write a test for this.


If there is a phase following the `FIXEDTERM`, we don't need the new Expired event (only as a last phase), so in this case we don't need to enter this method.

I **think** we don't need the flag `phaseEventCreated` and we may even be able to reuse/share the `createNextPhaseEvent` code.

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestSubscriptionHelper.java
```diff
@@ -150,15 +150,20 @@ public DefaultSubscriptionBase createSubscription(final SubscriptionBaseBundle b
...
+        return createSubscription(false, bundle, productName, term, planSet, phaseType, requestedDate,null);
     }
+    
+    public DefaultSubscriptionBase createSubscription(final SubscriptionBaseBundle bundle, final String planName)
+            throws SubscriptionBaseApiException {
+        return createSubscription(false, bundle, null, null, null, null, null,planName);
+    }        
 
-    private DefaultSubscriptionBase createSubscription(final boolean noEvents, @Nullable final SubscriptionBaseBundle bundle, final String productName, final BillingPeriod term, final String planSet, final PhaseType phaseType, final LocalDate requestedDate)
+    private DefaultSubscriptionBase createSubscription(final boolean noEvents, @Nullable final SubscriptionBaseBundle bundle, final String productName, final BillingPeriod term, final String planSet, final PhaseType phaseType, final LocalDate requestedDate, final String planName)
```
sbrossie, line 166: Could we group all the catalog info in the signature, e.g `productName`, `term`, `planSet`, `planName` and then the `requestedDate` ?

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCreate.java
```diff
@@ -216,6 +217,146 @@ public void testSimpleCreateSubscription() throws SubscriptionBaseApiException {
...
+        final DateTime expiryDate = ((ExpiredEvent) events.get(0)).getEffectiveDate();
+        final DateTime expectedExpiryDate = TestSubscriptionHelper.addDuration(subscription.getStartDate(), currentPhase.getDuration());
+        assertEquals(expiryDate, expectedExpiryDate);
+
+        //MOVE PAST FIXEDTERM PHASE
+        testListener.pushExpectedEvent(NextEvent.EXPIRED);
+        final Interval it = new Interval(clock.getUTCNow(), clock.getUTCNow().plusMonths(36));
+        clock.addDeltaFromReality(it.toDurationMillis());
+        assertListenerStatus();
+    }
```
sbrossie, line 252: 👍

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCreate.java
```diff
@@ -216,6 +217,146 @@ public void testSimpleCreateSubscription() throws SubscriptionBaseApiException {
...
+        final DateTime expectedExpiryDate = TestSubscriptionHelper.addDuration(subscription.getStartDate().plusDays(30), currentPhase.getDuration()); 
+        assertEquals(expiryDate, expectedExpiryDate);
+        
+        
+        //MOVE PAST FIXEDTERM PHASE
+        testListener.pushExpectedEvents(NextEvent.EXPIRED);
+        it = new Interval(clock.getUTCNow(), clock.getUTCNow().plusMonths(12));
+        clock.addDeltaFromReality(it.toDurationMillis());
+        assertListenerStatus();
+    }
```
sbrossie, line 297: 👍

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiCreate.java
```diff
@@ -216,6 +217,146 @@ public void testSimpleCreateSubscription() throws SubscriptionBaseApiException {
...
+        final DateTime expectedExpiryDate = TestSubscriptionHelper.addDuration(subscription.getStartDate().plusDays(30).plusMonths(6), currentPhase.getDuration()); 
+        assertEquals(expiryDate, expectedExpiryDate);
+        
+        
+        //MOVE PAST FIXEDTERM PHASE
+        testListener.pushExpectedEvents(NextEvent.EXPIRED);
+        it = new Interval(clock.getUTCNow(), clock.getUTCNow().plusMonths(12));
+        clock.addDeltaFromReality(it.toDurationMillis());
+        assertListenerStatus();        
+    }
```
sbrossie, line 358: 👍

sbrossie (commented):
Looking good, left a few comments to look at. Also it looks like  CI test are failing, not sure if this expected at this stage -- if so, could be an interesting thing to add in the PR comment.

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
reshmabidikar: Fixed

subscription/src/main/java/org/killbill/billing/subscription/engine/core/DefaultSubscriptionBaseService.java
reshmabidikar: Fixed

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
reshmabidikar: Modified  `createExpiredEvent` to remove `phaseEventCreated` parameter. You are right, it might be possible to combine the `createNextPhaseEvent` and `createExpiredEvent` methods, I've not done this change, just added a `TODO` statement as of now.

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestSubscriptionHelper.java
reshmabidikar: Fixed

reshmabidikar:
Commit https://github.com/killbill/killbill/pull/1592/commits/5eefb42e4e0ffae4fddf72aef725eb4aa813454d includes the following changes:

- Changes as per review comments
- Additional test for the condition where a `FIXEDTERM` phase is followed by an `EVERGREEN` phase (to ensure that the `Expired` event is not created in this case)
- Formatting changes

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestSubscriptionHelper.java
```diff
@@ -99,6 +99,39 @@ public TestSubscriptionHelper(final SubscriptionBaseInternalApi subscriptionApi,
         this.internalCallContextFactory = internalCallContextFactory;
     }
 
+    public static DateTime addOrRemoveDuration(final DateTime input, final List<Duration> durations, final boolean add) {
```
sbrossie, line 102: Just to make sure I understand, this code just got moved from bottom of the file to the top, but there was no change, right?

sbrossie (approved):

subscription/src/test/java/org/killbill/billing/subscription/api/user/TestSubscriptionHelper.java
reshmabidikar: Yes, after running the code formatter in Intellij, this method was moved from the bottom of the file to the top. There was no other change.

sbrossie (approved):

PR #1595, opened 2022-02-14
invoice: print catalog version in error message

pierre:


sbrossie (approved):
Should we include this info in more log output - for e.g there are multiple occurence of `ILLEGAL INVOICING STATE` in the code, should this be present each time?

pierre:
See https://github.com/killbill/killbill/pull/1601 instead.

PR #1597, opened 2022-02-15
payment: fix race condition in on-the-fly Janitor

pierre:
When running the on-the-fly Janitor while the onSuccessCalls of the initial payment control state machine haven't finished running yet, the plugin properties could have been wiped.

Draft PR as this is very much a conversation starter (we cannot just store the initial plugin properties, as it would introduce a regression for payment deployments).


sbrossie:
@pierre Should we close or merge - don't have a strong feeling either way, your choice.

pierre:
Should we address this [TODO](https://github.com/killbill/killbill/pull/1597/files#diff-d841db8edf5d8f41574f4ed0353e2b124833bf69eae6fbc8811105fadd6db883R129), i.e. introducing a config for this, and commit the fix?

PR #1601, opened 2022-02-17
include extra info on logs when exception is raised by ContiguousIntervalUsageInArrear during invoicing

noorarahili:
When killbill start invoicing the usage subscriptions, ContiguousIntervalUsageInArrear will look into the units if they are included on the catalog or not.
This PR is to include more useful information on logs which will help to understand the exact catalog doesn't contain the mentioned unit.


pierre:
First test suites are passing, merging due to urgency of fix and low risk.

sbrossie (commented):
👍 Thanks @noorarahili

Issue #1325, opened 2020-07-17
Large usage amount could trigger overflow


sbrossie:
See [email thread from the forum](https://groups.google.com/g/killbilling-users/c/IW31ZpYwFzM)

msqr:
If you would like to iterate on the [work I invested thus far](https://github.com/SolarNetwork/killbill/commit/53805e93f27e0797b096d59deef9f3e34bb9c393), I can submit that commit as PR.

msqr:
I think there might also be an overflow issue when `UsageDetailMode.DETAIL` results are requested, because the usage amounts are copied to the `InvoiceItem.quantity` property, which is defined as an Integer as well. I didn't noticed this before because I have not been using the `UsageDetailMode.DETAIL` mode.

sbrossie:
Also see https://github.com/killbill/killbill-api/issues/76

sbrossie:
@xsalefter I believe the change you made in https://github.com/killbill/killbill/issues/1469 will also address this issue, as we won't have the Integer limitation: The new type is `BigDecimal`, which is arbitrary precision so it can scale beyond what `Integer` offers.

PR #1603, opened 2022-02-17
Revert work for 1469 0.23.x

xsalefter:
This PR revert f5b7921f.


invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalConsumableUsageInArrear.java
```diff
@@ -199,75 +201,78 @@ protected UsageInArrearAggregate getToBeBilledUsageDetails(final LocalDate start
...
             tierNum++;
-            final long blockTierSize = tieredBlock.getSize().longValue();
-            final long tmp = remainingUnits / blockTierSize + (remainingUnits % blockTierSize == 0 ? 0 : 1);
-            long nbUsedTierBlocks;
-            if (tieredBlock.getMax() != (double) -1 && tmp > tieredBlock.getMax() ) {
-                nbUsedTierBlocks = tieredBlock.getMax().longValue();
-                remainingUnits -= tieredBlock.getMax() * blockTierSize;
+            final BigDecimal blockTierSize = tieredBlock.getSize();
+            final BigDecimal blockTierMax = tieredBlock.getMax();
+            final long tmp = remainingUnits.longValue() / blockTierSize.longValue() + (remainingUnits.longValue() % blockTierSize.longValue() == 0 ? 0 : 1);
```
sbrossie, line 216: I think we should also use the `BigDecimal#divideAndRemainder` instead of converting to long here - the way it's done a few lines below ?

sbrossie (commented):
It looks good, it seems like you were able to fix the merge issue.

I left you one comment (I think we missed that), and I think we should **test** the ability to specify non integer usage record and properly invoice for them.  You could leverage our existing usage api test [here](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestUsage.java#L50) to verify that.

invoice/src/main/java/org/killbill/billing/invoice/usage/ContiguousIntervalConsumableUsageInArrear.java
```diff
@@ -199,75 +201,78 @@ protected UsageInArrearAggregate getToBeBilledUsageDetails(final LocalDate start
...
             } else {
                 targetBlock = tieredBlock;
                 targetTierNum = tierNum;
                 break;
             }
         }
         final long lastBlockTierSize = targetBlock.getSize().longValue();
-        final long nbBlocks = units / lastBlockTierSize + (units % lastBlockTierSize == 0 ? 0 : 1);
-        return new UsageConsumableInArrearTierUnitAggregate(targetTierNum, targetBlock.getUnit().getName(), targetBlock.getPrice().getPrice(getCurrency()), targetBlock.getSize().longValue(), nbBlocks);
+        final long nbBlocks = units.longValue() / lastBlockTierSize + (units.longValue() % lastBlockTierSize == 0 ? 0 : 1);
```
xsalefter, line 269: @sbrossie do you think calculation here should be converted to BigDecimal as well? 

If so, what happened if this line
```java
units.longValue() / lastBlockTierSize // which will be converted to: units.divide(lastBlockTierSize).add( ..... );
```

Having fractional number left? For example, [this test line](https://github.com/killbill/killbill/blob/d1c910ba7828e0baa8fc79dfbad1b46ae9591dec/invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java#L277) will return `51.001` instead of `51.0`.

sbrossie: @xsalefter Yes, thank you for pointing that out. This should also be changed - all the computation should be done using `BigDecimal` (no more `longValue()` or `doubleValue()` conversion)


So, I think this is as simple as replacing the 2 lines:

```
        final long lastBlockTierSize = targetBlock.getSize().longValue();
        final long nbBlocks = units.longValue() / lastBlockTierSize + (units.longValue() % lastBlockTierSize == 0 ? 0 : 1);

```

With
```
        final BigDecimal lastBlockTierSize = targetBlock.getSize();
        final BigDecimal[] divRemaining =  units.divideAndRemainder(lastBlockTierSize);
        final BigDecimal nbBlocks =  (divRemaining[1].compareTo(BigDecimal.ZERO) == 0) ? divRemaining[0] : divRemaining[0].add(BigDecimal.ONE);
```
(And removing the now unecessary `BigDecimal.valueOf` conversion below)

sbrossie:
> I think we should test the ability to specify non integer usage record and properly invoice for them. You could leverage our existing usage api test [here](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestUsage.java#L50) to verify that.

@xsalefter I was also looking at this other related GH ticket, #1325, and [commented](https://github.com/killbill/killbill/issues/1325#issuecomment-1045406936). If you are writing a test to verify we can now use 'floating' point value, perhaps another test involving very large amount (bigger than an `int`) would be good as well.

xsalefter:
@sbrossie I marked all conversation here as "resolved". Also, marked this ready to review because most of CI passed. The one that's failed because failed test in `TestEntitlement`. But when I tested it locally, it's fine.

PR #1605, opened 2022-02-18
.github: add e2e job

pierre:
Ruby (core) tests are running again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,3 +8,161 @@ on:
 jobs:
   ci:
     uses: killbill/gh-actions-shared/.github/workflows/ci.yml@main
+  e2e:
+    runs-on: ubuntu-latest
+    services:
+      mysql:
+        image: mysql:8.0
+        env:
+          MYSQL_ROOT_PASSWORD: root
+        ports:
+        - 3306:3306
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=10
+    strategy:
+      matrix:
+        include:
+          - ruby-version: '2.4.2'
+            ref: refs/heads/master
+            database-adapter: 'mysql2'
+            database-user: 'root'
+            database-password: 'root'
+            database-port: '3306'
+            database: 'mysql'
+            java-version: '11'
+            java-distribution: 'temurin'
+            maven_version: '3.8.4'
+    steps:
+      - name: Checkout killbill
+        uses: actions/checkout@v2.4.0
+        with:
+          persist-credentials: false
+      - name: Set up JDK
+        uses: actions/setup-java@v2.5.0
+        with:
+          java-version: ${{ matrix.java-version }}
+          distribution: ${{ matrix.java-distribution }}
+          cache: 'maven'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -8,3 +8,161 @@ on:
...
+        if: ${{ matrix.database == 'postgresql' }}
+        run: |
+          curl https://raw.githubusercontent.com/killbill/killbill-avatax-plugin/master/src/main/resources/ddl.sql | psql -h 127.0.0.1 -U postgres -p 5432 -d killbill
+          curl https://raw.githubusercontent.com/killbill/killbill-payment-test-plugin/master/src/main/resources/ddl.sql | psql -h 127.0.0.1 -U postgres -p 5432 -d killbill
+          curl https://raw.githubusercontent.com/killbill/killbill-email-notifications-plugin/master/src/main/resources/ddl.sql | psql -h 127.0.0.1 -U postgres -p 5432 -d killbill
+      - name: Start Kill Bill
+        timeout-minutes: 4
+        run: |
+          mkdir -p /tmp/test-results
+          nohup ./bin/start-server -s > /tmp/test-results/killbill.log 2>&1 &
```
sbrossie, line 109: Should we either modify our default `profiles/killbill/src/main/resources/killbill-server.properties` to have the `org.killbill.security.shiroNbHashIterations=1` or create a special one for the test?

sbrossie: Also, I am confused how it works for the `postgresql' `matrix.database` - wouldn't we need to modify the `org.killbill.dao.url` property?

.github/workflows/ci.yml
```diff
@@ -8,3 +8,161 @@ on:
...
+    strategy:
+      matrix:
+        include:
+          - ruby-version: '2.4.2'
+            ref: refs/heads/master
+            database-adapter: 'mysql2'
+            database-user: 'root'
+            database-password: 'root'
+            database-port: '3306'
+            database: 'mysql'
```
sbrossie, line 30: Seems like the job supports different values for `database` but all properties here are targeted for `mysql` - Am i missing something?

sbrossie (approved):
👏

.github/workflows/ci.yml
pierre: > Should we either modify our default `profiles/killbill/src/main/resources/killbill-server.properties` to have the `org.killbill.security.shiroNbHashIterations=1` or create a special one for the test?

👍 for changing the default. Maybe in `work-for-release-0.23.x`?

> Also, I am confused how it works for the `postgresql` `matrix.database` - wouldn't we need to modify the `org.killbill.dao.url` property?

Yup, the PostgreSQL setup is only 80% done. I wanted to see how stable the MySQL one is for a few weeks first, before also enabling PostgreSQL.

.github/workflows/ci.yml
pierre: PostgreSQL support is for future work 😄 

We have it enabled on the other repo (https://github.com/killbill/killbill-integration-tests/actions/runs/1863918667), but I just wanted to see how stable the job is here, as the setup is a tad different.

PR #1619, opened 2022-03-07
Changes for cancel and changePlan processing

reshmabidikar:
Changes for cancelPlan and changePlan processing. A few notes:

1. I've assumed that it should not be possible to cancel an expired subscription or change plan once a subscription is expired since this is in line with how cancel behaves (that is it is not possible to cancel/change a plan for a canceled subscription).
2. `cancel` require a number of tests for various scenarios (`cancel`, `cancelWithDate`, `cancelWithPolicy`, `cancelBeforeFixedTermExpiry`, `cancelAfterFixedTermExpiry`, `cancelSubscriptionWithAddOn` to name just a few). The same goes for change plan. For now, I've written only a subset of these tests in `TestUserApiExpire`. Let me know if you think we need more tests to cover some of the above scenarios and/or other scenarios that I may have missed.
1. For plan change, I've only written tests for `START_OF_SUBSCRIPTION` and` START_OF_PLAN` alignments since these seem like the most common change alignment cases. Let me know if more tests are needed for the other types of alignments.

### DIFF
--- a/api/src/main/java/org/killbill/billing/subscription/api/SubscriptionBase.java
+++ b/api/src/main/java/org/killbill/billing/subscription/api/SubscriptionBase.java
@@ -84,6 +84,8 @@ public interface SubscriptionBase extends Entity, Blockable {
     public DateTime getEndDate();
 
     public DateTime getFutureEndDate();
+    
+    public DateTime getFutureExpiryDate();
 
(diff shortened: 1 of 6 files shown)


sbrossie (approved):
Great job!

PR #1621, opened 2022-03-11
Update KB ecosystem to use Java fluent api - killbill-util

xsalefter:
refactor collection manupulation from Guava to Java Stream and Collection API. This commit only affected (half of) killbill-util module.


util/src/main/java/org/killbill/billing/util/PluginProperties.java
```diff
@@ -50,13 +49,13 @@
...
+        final List<PluginProperty> list = new ArrayList<>();
         if (data != null) {
             for (final String key : data.keySet()) {
                 final PluginProperty property = new PluginProperty(key, data.get(key), false);
-                propertiesBuilder.add(property);
+                list.add(property);
             }
         }
-        return propertiesBuilder.build();
+        return List.copyOf(list);
```
pierre, line 59: For these `ImmutableList`, should we instead return a `Collections.unmodifiableList` view? That way, we keep the _Immutable_ aspect?

util/src/main/java/org/killbill/billing/util/cache/KillBillCacheController.java
```diff
@@ -127,6 +124,7 @@ public void removeAll() {
 
     @Override
     public int size() {
+        // FIXME-1615 : Dont know what to do to replace Iterables.size().
```
pierre, line 127: Can you use `Stream`? Not sure how performance it is though:

```
StreamSupport.stream(cache.spliterator(), false).count();
```

util/src/main/java/org/killbill/billing/util/customfield/api/DefaultCustomFieldUserApi.java
```diff
@@ -38,21 +39,18 @@
...
 public class DefaultCustomFieldUserApi implements CustomFieldUserApi {
 
-    private static final Function<CustomFieldModelDao, CustomField> CUSTOM_FIELD_MODEL_DAO_CUSTOM_FIELD_FUNCTION = new Function<CustomFieldModelDao, CustomField>() {
-        @Override
-        public CustomField apply(final CustomFieldModelDao input) {
-            return new StringCustomField(input);
-        }
-    };
+    /**
+     * FIXME-1615 : Cannot replaced by java.util.function.Function because
```
pierre, line 49: For these, we could replace them step by step, e.g. do a single PR to address `DefaultPaginationHelper` across all modules.

util/src/main/java/org/killbill/billing/util/customfield/api/DefaultCustomFieldUserApi.java
```diff
@@ -117,18 +115,15 @@ public DefaultCustomFieldUserApi(final InternalCallContextFactory internalCallCo
     public void addCustomFields(final List<CustomField> customFields, final CallContext context) throws CustomFieldApiException {
         if (!customFields.isEmpty()) {
             final InternalCallContext internalCallContext = internalCallContextFactory.createInternalCallContext(customFields.get(0).getObjectId(), customFields.get(0).getObjectType(), context);
-            final Iterable<CustomFieldModelDao> transformed = Iterables.transform(customFields, new Function<CustomField, CustomFieldModelDao>() {
-                @Override
-                public CustomFieldModelDao apply(final CustomField input) {
-                    // Respect user-specified ID
-                    // TODO See https://github.com/killbill/killbill/issues/35
```
pierre, line 124: Can you keep these comments?

util/src/main/java/org/killbill/billing/util/glue/KillBillShiroModule.java
```diff
@@ -145,6 +144,8 @@ protected void bindSecurityManager(final AnnotatedBindingBuilder<? super Securit
         //super.bindSecurityManager(bind);
         bind.toInstance(defaultSecurityManager);
 
+        // FIXME-1615 : Intellij report there' duplicate line like this, which is in KillBillShiroWebModule line 86.
```
pierre, line 147: `KillBillShiroModule` and `KillBillShiroWebModule` are very similar, but serve quite a different purpose. Not sure it's worth abstracting it (e.g. through a base class) though.

util/src/main/java/org/killbill/billing/util/tag/dao/TagDefinitionModelDao.java
```diff
@@ -31,6 +31,13 @@
 
 public class TagDefinitionModelDao extends EntityModelDaoBase implements EntityModelDao<TagDefinition> {
 
+    /**
+     * FIXME-1615 : We can use String Joiner here, but:
+     * 1. it only accept string, so we need to do things like
+     *    tag.getApplicableObjectTypes().stream().map( ... ).collect() (see line 58)
+     * 2. as this is in several places, maybe we need a utility classes dedicated for this
+     * 3. And lead to, "Can't we just use Google's Joiner?")
```
pierre, line 39: Indeed, good question.

The primary reason for this work is to follow the migration suggestions from https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Iterables.html:

> Java 8 users: several common uses for this class are now more comprehensively addressed by the new Stream library. Read the method documentation below for comparisons. This class is not being deprecated, but we gently encourage you to migrate to streams.

A secondary goal is to investigate if we can get rid of the library completely, mostly because of transitive dependency issues (check the comments from https://github.com/google/guava/issues/3320). I don't know if it's really feasable or realistic though. I would suggest you keep track of these questions and we make a decision down the line (if it's only a couple of utility functions, maybe then it's worth copying them).

pierre (commented):
Great start 👍 

I've added a few thoughts and comments.

util/src/main/java/org/killbill/billing/util/PluginProperties.java
xsalefter: Yes, I had a doubt whether List.copyOf() is immutable, but according to javadocs, it is. So I think `List.copyOf()` is enough.

util/src/main/java/org/killbill/billing/util/tag/dao/TagDefinitionModelDao.java
xsalefter: Ok so lets keep the comment for a while.

util/src/main/java/org/killbill/billing/util/cache/CacheController.java
```diff
@@ -32,6 +32,7 @@
 
     boolean remove(K key);
 
+    // FIXME-1615 : We can change this, but will affect at least CacheController and AdminResource (in jax-rs module)
```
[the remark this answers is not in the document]
xsalefter: > Can we rely on java java.util.function.Function

As I said in the FIXME, we can, but it affect other module. So perhaps keep that FIXME for a while?

Or, am I not answering your question?

util/src/main/java/org/killbill/billing/util/entity/DefaultPagination.java
```diff
@@ -157,6 +157,8 @@ public boolean equals(final Object o) {
         if (currentOffset != null ? !currentOffset.equals(that.currentOffset) : that.currentOffset != null) {
             return false;
         }
+        // FIXME-1615 : I think If iterator is really needed here, then Guava ImmutableList.copyOf() does a good job
+        // to handle iterator.
```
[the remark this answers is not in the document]
xsalefter: I keep the FIXME there, with up-to-date description.

sbrossie (approved):

Issue #1627, opened 2022-03-16
Transfer api should allow specifying Subscription#externalKey


sbrossie:
Our bundle transfer api was designed prior we introduced `Subscription#ExternalKey` - at the time we only had `SubscriptionBundle#ExternalKey`. We should enhance the api to allow passing such keys.

sbrossie:
We should also allow to set the per-subscription BCD

sbrossie:
Proposal for what's next:
1. For the `quantity` value, we implement the [missing piece](https://github.com/killbill/killbill/pull/1805#discussion_r1036188588) to always transfer `quantity` subscription events - when they exist.
2.  For the BCD:

We introduce a new parameter `BcdTransfer` for the [transfer api](https://github.com/killbill/killbill-api/blob/work-for-release-0.23.x/src/main/java/org/killbill/billing/entitlement/api/EntitlementApi.java#L201), whose values are:
```
    public enum BcdTransfer {
        NONE, /* Does not transfer per-subscription BCD */
        USE_EXISTING, /* Transfer existing per-subscription BCD if exists */ 
        ALIGN_WITH_EFF_DT, /* Set per-subscription BCD to align with the day of the transfer effective date */
    }
```

Note that, as part of this [commit](https://github.com/killbill/killbill/commit/0a579dde347b39e4d498983c338c2ea51441da32), the default behavior is to basically implement `USE_EXISTING`, and the goal would be to make it more flexible.

PR #1629, opened 2022-03-17
subscription, jaxrs: Provide a (hacky) way to specify Subscription keys during transfer. See #1627

sbrossie:
Because of apis restrictions, we rely on plugin properties to pass this information.

The format of the (plugin properties) keys is undocumented and is subject to change in future releases; it is namespaced with `KB_` to minimize collisions with possible use cases of entitlement plugin using their own keys.

Proper fix will be introduced in `0.24` release.

### DIFF
--- a/entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApiFast.java
+++ b/entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApiFast.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2020-2022 Equinix, Inc
+ * Copyright 2014-2022 The Billing Project, LLC
+ *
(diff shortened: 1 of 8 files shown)


entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApiFast.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2010-2013 Ning, Inc.
```
sbrossie, line 2: I 'll fix the copyright, don't want to start another CI run just yet.

pierre: FYI the feature finally shipped last week:
image

pierre (approved):

PR #1637, opened 2022-03-23
junction: NPE fix when blocking events after CANCEL

pierre:


junction/src/main/java/org/killbill/billing/junction/plumbing/billing/BlockingCalculator.java
```diff
@@ -202,12 +202,13 @@ public boolean insertBlockingEvents(final SortedSet<BillingEvent> billingEvents,
             // The last one during of before the duration
             final BillingEvent precedingFinalEvent = precedingBillingEventForSubscription(duration.getEnd(), subscriptionBillingEvents);
 
-            if (precedingInitialEvent != null) { // there is a preceding billing event
+            // We ignore anything beyond the CANCEL event
+            if (precedingInitialEvent != null && precedingInitialEvent.getTransitionType() != SubscriptionBaseTransitionType.CANCEL) { // there is a preceding billing event
```
sbrossie, line 206: I think it would be better to handle the `CANCEL` case inside the `precedingBillingEventForSubscription` -> `precedingActiveBillingEventForSubscription`.

Also, when  this gets merged into `work-for-release-0.23.x`, we will need to add the `EXPIRED` use case (hence the reason to embed all this in the `precedingActiveBillingEventForSubscription` method.

pierre: ack, sounds good. See f0c410d4faa5a5259ca8202f54715705d301e00c.

sbrossie (approved):

PR #1641, opened 2022-03-25
subscription: Fix issue in bundle transfer

sbrossie:
In cases where we have a BCD_UPADTE event, the code would throw an `java.lang.IndexOutOfBoundsException` because the event was not correctly handled.
This fix makes sure that when we have both a CREATE and BCD_UPADTE prior the transfer date, the code does not throw exception and also transfers the BCD alignment.


subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
```diff
@@ -114,66 +116,76 @@ private SubscriptionBaseEvent createEvent(final boolean firstEvent, final Existi
...
                 }
 
                 // Add previous event the first time if needed
-                if (prevEvent != null) {
-                    event = createEvent(firstEvent, prevEvent, subscription, transferDate, context);
-                    if (event != null) {
-                        result.add(event);
-                        firstEvent = false;
-                    }
+                if (insertEventToResult(firstEvent, prevEvent, subscription, transferDate, result)) {
```
pierre, line 150: I think I mostly managed to convince myself it works but I find the way it's written a tad off / hard to follow 😅 

We're essentially in a `for` loop and the bulk of the work is the last `insertEventToResult` (with the `cur`). The statements before are only there to insert the first `CREATE`/`BCD_CHANGE` events. Since the `if` statements are disjoints, it looks almost as if they could be inserted at any time, even though I believe they need to be insert together as soon as `cur` is after the transfer date.

subscription/src/test/java/org/killbill/billing/subscription/api/transfer/TestTransfer.java
```diff
@@ -515,4 +517,36 @@ public void testTransferWithUncancel() throws Exception {
...
+                                                                                                         ObjectType.ACCOUNT,
+                                                                                                         this.internalCallContext.getUpdatedBy(),
+                                                                                                         this.internalCallContext.getCallOrigin(),
+                                                                                                         this.internalCallContext.getContextUserType(),
+                                                                                                         this.internalCallContext.getUserToken(),
+                                                                                                         this.internalCallContext.getTenantRecordId());
+
+        subscriptionInternalApi.updateBCD(baseSubscription.getId(), 8, clock.getUTCToday(), originalcontext);
+
+        // 2012-5-8
```
pierre, line 540: Could you add a second `BCD_CHANGE` event to make sure we pick the right one?

pierre (approved):

sbrossie:
Merging as-is, will do a follow up PR with comments.

Issue #1615, opened 2022-03-03
Update KB ecosystem to use Java fluent api


sbrossie:
Our code base was started prior Java fluent api, and so we embraced Guava (e.g `Iterables.xxx`) to perform filtering, sorting, merging operations on data structures - e.g [here](https://github.com/killbill/killbill/blob/killbill-0.22.28/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestUsage.java#L85)

We would like to replace all those with new Java syntax and ideally remove guava as a dependency:

```
object.stream()
                .filter(xxx)
                .***
```

sbrossie:
In order to do this, we could work, repo by repo, and then for big repos such as `killbill/killbill` module by module (and make one PR per module).

PR #1650, opened 2022-04-01
1615 killbill catalog

xsalefter:
Work for #1615 in `catalog` module. 20+ classes, but one thing that probably worth to mention is about:

`<!-- Required scope for the maven shade plugin (tools) -->`

comment in Guava dependency [pom.xml](https://github.com/killbill/killbill/compare/work-for-release-0.23.x...xsalefter:1615-killbill_catalog?expand=1#diff-19265f7f2096d2f336503b67d72a52073836d053b0e4d995875919c0f0271364L41). We able to remove guava dependency and `fast`, `slow` build is success, but I'm not sure with shade plugin. Perhaps it is just old comment?

### DIFF
--- a/catalog/pom.xml
+++ b/catalog/pom.xml
@@ -39,12 +39,6 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <!-- Required scope for the maven shade plugin (tools) -->
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
             <groupId>com.google.inject</groupId>
             <artifactId>guice</artifactId>
             <scope>provided</scope>
--- a/catalog/src/main/java/org/killbill/billing/catalog/CatalogEntityCollection.java
+++ b/catalog/src/main/java/org/killbill/billing/catalog/CatalogEntityCollection.java
@@ -22,31 +22,30 @@ import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.util.Collection;
+import java.util.Comparator;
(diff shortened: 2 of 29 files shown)


pierre:
> one thing that probably worth to mention is about:
> 
> `<!-- Required scope for the maven shade plugin (tools) -->`
> 
> comment in Guava dependency [pom.xml](https://github.com/killbill/killbill/compare/work-for-release-0.23.x...xsalefter:1615-killbill_catalog?expand=1#diff-19265f7f2096d2f336503b67d72a52073836d053b0e4d995875919c0f0271364L41). We able to remove guava dependency and `fast`, `slow` build is success, but I'm not sure with shade plugin. Perhaps it is just old comment?

Good catch 👍 

This comment refers to the `*-load-tool.jar` and `*-xsd-tool.jar` (shaded) artifacts under `target/`. To verify they still work:

* `java -jar killbill-catalog-*-SNAPSHOT-load-tool.jar ../src/test/resources/org/killbill/billing/catalog/SpyCarBasic.xml`. You should see `Success: Catalog loads!` (ignore the slf4j warnings, if any).
* `java -jar killbill-catalog-*-SNAPSHOT-xsd-tool.jar foo.xsd`. You should see a valid `foo.xsd` file (~36K bytes).

xsalefter:
@pierre Thanks! I run them and see that `<XXX>-load-tool.jar` and `<XXX>-xsd-tool.jar` still working as you explained above.

sbrossie (approved):
Unclear about the comment, this is [indeed quite old](https://github.com/killbill/killbill/commit/fb0e21535ec31f472ce9e49d48cdd94ccb8603ab), perhaps @pierre has a clue?

PR #1657, opened 2022-04-09
subscription: Fix issue when transferring pending subscription

sbrossie:
The code needs to realign the requested date with the pending start date.
This needs to happen for the cancelation of the subscription being transferred, but also transferrable events like BCD_UPDATE.

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
@@ -241,7 +241,13 @@ public class DefaultSubscriptionBaseTransferApi extends SubscriptionApiBase impl
                                                          effectiveTransferDate.isBefore(oldSubscription.getChargedThroughDate()) ?
                                                          oldSubscription.getChargedThroughDate() : effectiveTransferDate;
 
-                    // Reuse logic from validateEffectiveDate where we check the latest transition and readjust the candidateCancelDate to match such transition if needed
+                    //
+                    // We are checking that if the subscription is PENDING (start date in the future) and if the requestedDate
+                    // for the transfer is prior to the startDate, then it gets realigned with the startDate.
+                    // The code goes further (reuse logic from validateEffectiveDate) and also checks that we if we have a subscription with multiple
+                    // change plans (that we want to transfer), we don't end up cancelling prior a previous transition as this would create some
+                    // weird REPAIR scenarios.
+                    //
                     final SubscriptionBaseTransition previousTransition = oldSubscription.getPreviousTransition();
                     final DateTime earliestValidDate = previousTransition != null ? previousTransition.getEffectiveTransitionTime() : oldSubscription.getStartDate();
                     final DateTime effectiveCancelDate = (candidateCancelDate.isBefore(earliestValidDate)) ? earliestValidDate : candidateCancelDate;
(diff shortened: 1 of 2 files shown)


subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
```diff
@@ -226,17 +228,24 @@ public SubscriptionBaseBundle transferBundle(final UUID sourceAccountId, final U
                 if (oldSubscription.getState() == EntitlementState.CANCELLED) {
                     continue;
                 }
+
                 final List<ExistingEvent> existingEvents = cur.getExistingEvents();
                 final ProductCategory productCategory = existingEvents.get(0).getProductCategory();
 
                 // For future add-on cancellations, don't add a cancellation on disk right away (mirror the behavior
                 // on base plan cancellations, even though we don't support un-transfer today)
```
pierre, line 236: The comment

subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
```diff
@@ -226,17 +228,24 @@ public SubscriptionBaseBundle transferBundle(final UUID sourceAccountId, final U
...
                 if (productCategory != ProductCategory.ADD_ON || cancelImmediately) {
                     // Create the cancelWithRequestedDate event on effectiveCancelDate
-                    final DateTime effectiveCancelDate = !cancelImmediately && oldSubscription.getChargedThroughDate() != null &&
+                    final DateTime candidateCancelDate = !cancelImmediately &&
+                                                         oldSubscription.getChargedThroughDate() != null &&
                                                          effectiveTransferDate.isBefore(oldSubscription.getChargedThroughDate()) ?
                                                          oldSubscription.getChargedThroughDate() : effectiveTransferDate;
 
+                    // Reuse logic from validateEffectiveDate where we check the latest transition and readjust the candidateCancelDate to match such transition if needed
+                    final SubscriptionBaseTransition previousTransition = oldSubscription.getPreviousTransition();
```
pierre, line 245: I'm a bit far from this code, but doesn't `getPreviousTransition()` depend on `now()`? Just wondering if that could impact the logic as it depends when the transfer call is made (regardless of the transfer date). Happy to chat through it IRL.

sbrossie: So, mostly what we are interested in doing here, is checking that if the subscription is `PENDING` (start date in the future) and if the requestedDate for the transfer is prior to the startDate, then it gets realigned with the startDate. 

The code as written goes a bit further and also checks that we if we have a subscription with multiple change plans (that we want to transfer), we don't end up cancelling prior a previous transition - that would create some weird REPAIR scenarios. So checking the previous transition as compared to `now` seems reasonable.

If there is a pending transition, we would allow the transfer to occur prior with no realignment and then have to decide what to do as shown [here](https://github.com/killbill/killbill/blob/killbill-0.22.28/subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java#L106)

sbrossie: I also checked that we should not have multiple future transitions as we have code to deactivate those [e.g here](https://github.com/killbill/killbill/blob/killbill-0.22.28/subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java#L767).

pierre: The comments above are quite helpful, I would integrate them in the code 😄

pierre (approved):

subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
sbrossie: Fixed in https://github.com/killbill/killbill/pull/1657/commits/1c70d812f6159a13f7c06f13f63508b9d2fb9666

PR #1659, opened 2022-04-13
Changes for 1375

reshmabidikar:
@sbrossie, I've added a few comments where I need clarifications from you. Could you please take a look when you get a chance?

### DIFF
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
@@ -254,13 +254,13 @@ public class DefaultEntitlement extends EntityBase implements Entitlement {
     }
 
     @Override
-    public LocalDate getEffectiveStartDate() {
-        return eventsStream.getEntitlementEffectiveStartDate();
+    public DateTime getEffectiveStartDate() {
+        return eventsStream.getEntitlementEffectiveStartDateTime();
     }
 
     @Override
-    public LocalDate getEffectiveEndDate() {
-        return eventsStream.getEntitlementEffectiveEndDate();
+    public DateTime getEffectiveEndDate() {
+        return eventsStream.getEntitlementEffectiveEndDateTime();
     }
 
     @Override
@@ -316,7 +316,7 @@ public class DefaultEntitlement extends EntityBase implements Entitlement {
         // Get the latest state from disk
         refresh(callContext);
 
-        if (entitlementEffectiveDate != null && entitlementEffectiveDate.compareTo(getEffectiveStartDate()) < 0) {
+        if (entitlementEffectiveDate != null && entitlementEffectiveDate.compareTo(internalTenantContext.toLocalDate(getEffectiveStartDate())) < 0) { //TODO_1375 - Change DateTime to LocalDate for now. Revisit once we change API method to receive DateTime
             throw new EntitlementApiException(ErrorCode.SUB_INVALID_REQUESTED_DATE, entitlementEffectiveDate, getEffectiveStartDate());
         }
 
@@ -538,7 +538,7 @@ public class DefaultEntitlement extends EntityBase implements Entitlement {
             default:
                 throw new RuntimeException("Unsupported policy " + entitlementPolicy);
         }
-        return (cancellationDate.compareTo(getEffectiveStartDate()) < 0) ? getEffectiveStartDate() : cancellationDate;
+        return (cancellationDate.compareTo(internalTenantContext.toLocalDate(getEffectiveStartDate())) < 0) ? internalTenantContext.toLocalDate(getEffectiveStartDate()) : cancellationDate; //TODO_1375 - Return LocalDate for now. Revisit later to check if this method should return DateTime
     }
 
 
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionJson.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionJson.java
@@ -135,7 +135,7 @@ public class SubscriptionJson extends JsonBase {
             final PlanPhase phase = subscriptionEvent.getNextPhase() != null ? subscriptionEvent.getNextPhase() : subscriptionEvent.getPrevPhase();
             this.eventId = subscriptionEvent.getId();
             this.billingPeriod = billingPeriod;
-            this.effectiveDate = subscriptionEvent.getEffectiveDate().toLocalDate(); //TODO_1375 - Added this to keep the jaxrs layer the same, will be updated in step 2. Not sure how to access internalCallContext here so I've used toLocalDate
+            this.effectiveDate = subscriptionEvent.getEffectiveDate() != null ? subscriptionEvent.getEffectiveDate().toLocalDate() : null; //TODO_1375 - Added this to keep the jaxrs layer the same, will be updated in step 2. Not sure how to access internalCallContext here so I've used toLocalDate
             this.plan = plan != null ? plan.getName() : null;
             this.product = product != null ? product.getName() : null;
             this.priceList = priceList != null ? priceList.getName() : null;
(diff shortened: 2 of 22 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/CatalogResource.java
```diff
@@ -457,13 +457,13 @@ private SubscriptionEvent getLastEventBeforeDate(final UUID subscriptionId, fina
         SubscriptionEvent lastEventBeforeRequestedDate = null;
         for (final SubscriptionEvent subscriptionEvent : subscription.getSubscriptionEvents()) {
             if (lastEventBeforeRequestedDate == null) {
-                if (subscriptionEvent.getEffectiveDate().compareTo(requestedDate) > 0) {
+                if (subscriptionEvent.getEffectiveDate().compareTo(requestedDateTime) > 0) { //TODO_1375 - Is this correct or should we compare LocalDate?
```
sbrossie, line 460: I think we should keep it the way we have it (using `DateTime`): The code now manipulates `DateTime` - if this causes an issue, then we can revisit.

sbrossie (approved):

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -632,7 +632,7 @@ public Response doOperation(final CallContext ctx)
                 final Subscription subscription = subscriptionApi.getSubscriptionForEntitlementId(newEntitlement.getId(), ctx);
                 if (subscription.getBillingEndDate() != null) {
                     final Account account = accountUserApi.getAccountById(subscription.getAccountId(), callContext);
-                    final boolean inTheFuture = isInTheFuture(subscription.getBillingEndDate(), account);
+                    final boolean inTheFuture = isInTheFuture(subscription.getBillingEndDate().toLocalDate(), account); //TODO_1375 - Converting to LocalDate for comparison, should we instead change isInTheFuture to accept DateTime? Also, not sure how to access internalCallContext here so I've used toLocalDate
                     if (inTheFuture) {
```
reshmabidikar, line 636: @sbrossie, could you please comment on the above too? Do, you think we should use `DateTime` here? If not, if we should use `LocalDate` how should we do the conversion? I was not able to find a way to access `internalCallContext` in the `SubscriptionResoource` class.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java
```diff
@@ -133,7 +133,7 @@ public Response recordUsage(final SubscriptionUsageRecordJson json,
         final Entitlement entitlement = entitlementApi.getEntitlementForId(json.getSubscriptionId(), callContext);
         if (entitlement.getEffectiveEndDate() != null) {
             final LocalDate highestRecordDate = getHighestRecordDate(json.getUnitUsageRecords());
-            if (entitlement.getEffectiveEndDate().compareTo(highestRecordDate) < 0) {
+            if (entitlement.getEffectiveEndDate().toLocalDate().compareTo(highestRecordDate) < 0) { //TODO_1375 - Converting to LocalDate for comparison, should we instead convert highestRecordDate to DateTime and use it? Also, not sure how to access internalCallContext here so I've used toLocalDate
                 return Response.status(Status.BAD_REQUEST).build();
```
reshmabidikar, line 137: Same comment as above, should we use `DateTime` or `LocalDate` ?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
```diff
@@ -538,7 +538,7 @@ private LocalDate getLocalDateFromEntitlementPolicy(final EntitlementActionPolic
             default:
                 throw new RuntimeException("Unsupported policy " + entitlementPolicy);
         }
-        return (cancellationDate.compareTo(getEffectiveStartDate()) < 0) ? getEffectiveStartDate() : cancellationDate;
+        return (cancellationDate.compareTo(internalTenantContext.toLocalDate(getEffectiveStartDate())) < 0) ? internalTenantContext.toLocalDate(getEffectiveStartDate()) : cancellationDate; //TODO_1375 - Return LocalDate for now. Revisit later to check if this method should return DateTime
```
reshmabidikar, line 541: Same comment as above, should we change this method to return a `DateTime` instead of `LocalDate` ?

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionJson.java
```diff
@@ -135,7 +135,7 @@ public EventSubscriptionJson(final SubscriptionEvent subscriptionEvent, @Nullabl
             final PlanPhase phase = subscriptionEvent.getNextPhase() != null ? subscriptionEvent.getNextPhase() : subscriptionEvent.getPrevPhase();
             this.eventId = subscriptionEvent.getId();
             this.billingPeriod = billingPeriod;
-            this.effectiveDate = subscriptionEvent.getEffectiveDate();
+            this.effectiveDate = subscriptionEvent.getEffectiveDate() != null ? subscriptionEvent.getEffectiveDate().toLocalDate() : null; //TODO_1375 - Added this to keep the jaxrs layer the same, will be updated in step 2. Not sure how to access internalCallContext here so I've used toLocalDate
             this.plan = plan != null ? plan.getName() : null;
```
reshmabidikar, line 139: Again, need your input here. I'm converting `DateTime` to `LocalDate` to keep the jaxrs layer the same like we discussed. However, I was not sure how to access `internalCallContext` in the `SubscriptionJson` class, so I've used the  `toLocalDate()` method.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
```diff
@@ -316,7 +316,7 @@ public Entitlement cancelEntitlementWithDate(@Nullable final LocalDate entitleme
         // Get the latest state from disk
         refresh(callContext);
 
-        if (entitlementEffectiveDate != null && entitlementEffectiveDate.compareTo(getEffectiveStartDate()) < 0) {
+        if (entitlementEffectiveDate != null && entitlementEffectiveDate.compareTo(internalTenantContext.toLocalDate(getEffectiveStartDate())) < 0) { //TODO_1375 - Change DateTime to LocalDate for now. Revisit once we change API method to receive DateTime
```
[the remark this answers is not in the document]
sbrossie: I think what you have now makes sense - we receive as an input a `LocalDate` and so we want to compare that date with the startDate (which is now a `DateTime`) so it is safer to make the comparison the way you have it.. We 'll revisit if necessary.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
sbrossie: Keep it the way you have it (but keep the `TODO_1375` for these choices of converting one way or the other. I would say for now, we play safe (the way you have it). keep in mind, stage 2 is to introduce new apis that will take `DateTime` as an input, so this may force us to revisit things

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java
sbrossie: Same answer

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionJson.java
sbrossie: Ok, keep the TODO

pierre:
@reshmabidikar Regarding the `verify` job failing (`Found a problem with the direct dependency org.kill-bill.billing:killbill-api of the current project`), I've added a workaround in https://github.com/killbill/killbill-oss-parent/pull/371. Updating the parent `pom.xml` to `0.145.3-23d82a6-SNAPSHOT` should hopefully fix it.

pierre:
Regarding the Ruby integration tests, I can confirm the following one is constantly failing:

```
`test_create_subscription_with_tz_minus_11_no_requested_date'
     18: 
     19:     def test_create_subscription_with_tz_minus_11_no_requested_date
     20:       # No requested date -> will take today's date (and all conversion in account timezone will lead to '2013-08-01' and NOT '2013-08-02')
  => 21:       test_scenario_fixed_price('2013-08-02T06:00:00.000Z', 'Pacific/Samoa', nil, '2013-08-01', '2013-08-01')
     22:     end
     23: 
     24:     def test_create_subscription_with_tz_minus_11_with_requested_date_today
<"2013-08-01"> expected but was
<"2013-08-02">
```

Worth spending time debugging it before merging...

PR #1660, opened 2022-04-14
replace Guava usage with stream API and JSR-330 in kb-entitlement

xsalefter:
Nothing special. Just `DefaultEventsStream` have little bit method refactor.

Also, Guava dependency cannot removed yet because this module (BlockingStateDao) still using Guava's Optional and used by another module. [Example 1](https://github.com/killbill/killbill/compare/work-for-release-0.23.x...xsalefter:1615-entitlement?expand=1#diff-272cb459311988b9b136cfe3271df0d9c3b32fd07b79195f3e82e6efe89d6302R37), [Example 2](https://github.com/killbill/killbill/compare/work-for-release-0.23.x...xsalefter:1615-entitlement?expand=1#diff-a21e7c2e49e4f64ff24df4cdc6b8e769404b985847dbd91269f1c3258d27ddc7R69).

### DIFF
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementContext.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementContext.java
@@ -17,8 +17,11 @@
 
 package org.killbill.billing.entitlement.api;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import javax.annotation.Nullable;
 
@@ -32,9 +35,7 @@ import org.killbill.billing.util.callcontext.CallContext;
 import org.killbill.billing.util.callcontext.CallOrigin;
 import org.killbill.billing.util.callcontext.UserType;
 
-import com.google.common.base.Function;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
+import static org.killbill.billing.util.collect.CollectionTransformer.iterableToList;
 
 public class DefaultEntitlementContext implements EntitlementContext {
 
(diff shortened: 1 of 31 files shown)


sbrossie (approved):

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementContext.java
```diff
@@ -122,7 +120,9 @@ public DefaultBaseEntitlementWithAddOnsSpecifier apply(final BaseEntitlementWith
     private static <T> Iterable<T> lastUnlessEmptyOrFirst(final Iterable<T> prevValues, final Iterable<T> newValues) {
         // Be lenient if a plugin returns an empty list (default behavior for Ruby plugins): at this point,
         // we know the isAborted flag hasn't been set, so let's assume the user actually wants to use the previous list
-        if (newValues != null && !Iterables.isEmpty(newValues)) {
+        // FIXME-1615 : Instead of "iterableToList(newValues).isEmpty()", should we copy guava version to new class
+        //  "CollectionUtils.isIterableEmpty(Iterable)" ?
```
pierre, line 124: 👍

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementContext.java
```diff
@@ -32,9 +35,7 @@
 import org.killbill.billing.util.callcontext.CallOrigin;
 import org.killbill.billing.util.callcontext.UserType;
 
-import com.google.common.base.Function;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
+import static org.killbill.billing.util.collect.CollectionTransformer.iterableToList;
```
pierre, line 38: Are we always consistent with static vs no-static import? Wondering also if there is a rule we could add in our IntelliJ `codeStyle` to help enforce this?

entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
```diff
@@ -353,73 +336,70 @@ private Collection<BlockingState> computeAddonsBlockingStatesForNextSubscription
...
+                                                       KILLBILL_SERVICES.ENTITLEMENT_SERVICE.getServiceName(),
+                                                       true,
+                                                       true,
+                                                       false,
+                                                       blockingStateEffectiveDate))
+                .collect(Collectors.toUnmodifiableList());
+    }
+
+    /**
+     * FIXME-1615: Complex logic. Need to test properly.
```
pierre, line 365: Following-up on the Slack discussion, I think there is room for additional unit tests in a lot of our modules.

We have lots of integration tests, either in Kill Bill (`beatrix`) or outside (https://github.com/killbill/killbill-integration-tests), which verify various features, but we could enhance our coverage on a per method / class basis. This is sometimes difficult though as the setup required to write such tests is complex.

Let's continue to think about it, maybe something we can slowly improve over time, especially as we near the 0.24 release towards the end of the year.

On a related note, addressing Spotbugs warnings and errors would be nice too (/cc @vnandwana)...

pierre (approved):
😅

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementContext.java
xsalefter: Ok, instead of `CollectionUtils`, maybe will changed to our future `Iterables.isEmpty()` (as discussed by me and @sbrossie in slack)

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementContext.java
xsalefter: Hmm.. in general AFAIK we're not.. :sweat_smile: . For example, I've seen some assertion do use static import, and some don't. Should we open discussion about this in slack?

entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
xsalefter: Ok then.. Btw, lets keep that FIXME and I will back to this once `killbill-util` `collect` package refactored.

Issue #1661, opened 2022-04-15
Change in Plan#effectiveDateForExistingSubscriptions behavior


reshmabidikar:
Currently, the `Plan#effectiveDateForExistingSubscriptions` works such that the new plan is effective on the specified date and this results in prorations. There is a request from some customers for this feature to work so that the price change happens in the next period and avoids proration. So, we need to make the  `Plan#effectiveDateForExistingSubscriptions`  configurable so that it works in both situations. 

The desired behavior can be reproduced as follows:

1. Set date to `2022-03-29`
2. Upload `catalog-v1.xml`
3. Upload `catalog-v2.xml`
4. Set date to `2022-03-30`
5. Create a new account/subscription to the `standard-monthly` plan
6. This results in an invoice being generated for  `$0.16` for the period `2022-03-30` to `2022-03-31` (prorated on the basis of the old price of `$5`). 
7. Move the clock to  `2022-03-31` - This results in an invoice being generated for `$5.81` for the period `2022-03-31` to `2022-04-30` (prorated on the basis of the new price of `$6`). 
8. However, customers expect the user to be charged the old price (`$5`) for the current billing period ( `2022-03-30` to  `2022-04-30`) and  and the updated price ($6) for the next billing period ( `2022-04-30` to  `2022-05-30`) 

[catalog-v2.zip](https://github.com/killbill/killbill/files/8493354/catalog-v2.zip)
[catalog-v1.zip](https://github.com/killbill/killbill/files/8493355/catalog-v1.zip)

shaun-forgie:
@sbrossie so the isEffectiveDateForExistingSubscriptionsAlignedToBCD setting as used in subscription/src/main/java/org/killbill/billing/subscription/config/MultiTenantSubscriptionConfig.java can be found by retrieving tenant specific configuration settings as documented here https://docs.killbill.io/latest/plugin_installation.html#per-tenant-configuration ?

reshmabidikar:
Does not seem to work after testing via Kaui. This is what I did:
1. Set the `org.killbill.subscription.align.effectiveDateForExistingSubscriptions=true` property
2. Set date to `2022-03-29`
3. Uploaded catalogs `catalog-v1.xml` and `catalog-v2.xml`
4. Set date to `2022-03-30`
5. Create a new account with `BCD=30` and created a subscription to the `standard-monthly` plan
[...]

PR #1663, opened 2022-04-18
Prototype code for #1661

sbrossie:
* In https://github.com/killbill/killbill/commit/85a045f5a14b7112dc8a615414bd5cb8f52ecef8, there is a new test case that shows the current behavior.
* In https://github.com/killbill/killbill/commit/415c0c0a050ccf372989563688fe2c04d8fd8593, there is a prototype fix that removes the pro-ration when transition to a new Plan version.

`ci` tests fail, because the new behavior is hardcoded and in particular make the existing test `TestCatalogWithEffectiveDateForExistingSubscriptions.testRecurringPlan` fail (See TestCatalogWithEffectiveDateForExistingSubscriptions.java:120)

Notes:

- This new [method](https://github.com/killbill/killbill/pull/1663/files#diff-6c9485b2174859a0d155b69e387e4b868e364c2608ba08fd0a2db9c49101fc46R694) is very similar to [this one](https://github.com/killbill/killbill/blob/killbill-0.22.28/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java#L732), which we know need some attention (e.g billing alignment policies).
- The wiring is difficult and leads to that terrible piece [here](https://github.com/killbill/killbill/pull/1663/files#diff-6c9485b2174859a0d155b69e387e4b868e364c2608ba08fd0a2db9c49101fc46R761) // Would need tio be fixed
- The config inside `Subscription` does not exist, so if we go this direction this is more work. On `0.24.x`, we could make the behavior configurable inside the catalog.

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
@@ -38,7 +38,6 @@ import org.killbill.billing.subscription.catalog.SubscriptionCatalog;
 import org.killbill.billing.subscription.events.SubscriptionBaseEvent;
 import org.killbill.billing.util.callcontext.CallContext;
 import org.killbill.billing.util.callcontext.TenantContext;
-import org.skife.config.Param;
 
 public interface SubscriptionBaseApiService {
 
@@ -111,6 +110,6 @@ public interface SubscriptionBaseApiService {
 
     int getAccountBCD(InternalTenantContext context) throws SubscriptionBaseApiException;
 
-    // If we expose more than one config, we should not instead return the SubscriptionConfig object
-    boolean isEffectiveDateForExistingSubscriptionsAlignedToBCD(@Param("dummy") final InternalTenantContext tenantContext);
+    // If we expose more than one config, we should instead return the SubscriptionConfig object
+    boolean isEffectiveDateForExistingSubscriptionsAlignedToBCD(final InternalTenantContext tenantContext);
 }
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
@@ -108,6 +108,6 @@ public interface SubscriptionBaseApiService {
 
     boolean undoChangePlan(DefaultSubscriptionBase defaultSubscriptionBase, CallContext context) throws SubscriptionBaseApiException;
 
-    int getBCD(InternalTenantContext context) throws SubscriptionBaseApiException;
+    int getAccountBCD(InternalTenantContext context) throws SubscriptionBaseApiException;
 
 }
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
@@ -38,6 +38,7 @@ import org.killbill.billing.subscription.catalog.SubscriptionCatalog;
 import org.killbill.billing.subscription.events.SubscriptionBaseEvent;
 import org.killbill.billing.util.callcontext.CallContext;
 import org.killbill.billing.util.callcontext.TenantContext;
+import org.skife.config.Param;
 
 public interface SubscriptionBaseApiService {
 
@@ -110,4 +111,6 @@ public interface SubscriptionBaseApiService {
 
     int getAccountBCD(InternalTenantContext context) throws SubscriptionBaseApiException;
 
+    // If we expose more than one config, we should not instead return the SubscriptionConfig object
+    boolean isEffectiveDateForExistingSubscriptionsAlignedToBCD(@Param("dummy") final InternalTenantContext tenantContext);
 }
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
@@ -68,12 +68,12 @@ public interface SubscriptionBaseApiService {
 
     // Return the effective date of the change
     public DateTime changePlanWithRequestedDate(DefaultSubscriptionBase subscription, EntitlementSpecifier spec,
-                                                 DateTime requestedDate, CallContext context)
+                                                DateTime requestedDate, CallContext context)
             throws SubscriptionBaseApiException;
 
     // Return the effective date of the change
     public DateTime changePlanWithPolicy(DefaultSubscriptionBase subscription, EntitlementSpecifier spec,
-                                          BillingActionPolicy policy, CallContext context)
+                                         BillingActionPolicy policy, CallContext context)
             throws SubscriptionBaseApiException;
 
     public int handleBasePlanEvent(final DefaultSubscriptionBase subscription, final SubscriptionBaseEvent event, SubscriptionCatalog catalog, final CallContext context) throws CatalogApiException;
@@ -107,4 +107,7 @@ public interface SubscriptionBaseApiService {
                                                              final InternalTenantContext internalTenantContext) throws CatalogApiException;
 
     boolean undoChangePlan(DefaultSubscriptionBase defaultSubscriptionBase, CallContext context) throws SubscriptionBaseApiException;
+
+    int getBCD(InternalTenantContext context) throws SubscriptionBaseApiException;
+
 }
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
@@ -694,14 +694,15 @@ public class DefaultSubscriptionBase extends EntityBase implements SubscriptionB
     }
 
 
-    private DateTime alignToNextBCDIfRequired(final Plan curPlan, final PlanPhase curPlanPhase, final DateTime originalDate, final SubscriptionCatalog catalog, final Integer bcdLocal, final InternalTenantContext context) throws SubscriptionBaseApiException, CatalogApiException {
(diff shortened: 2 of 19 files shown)


subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
```diff
@@ -107,4 +107,7 @@ public List<SubscriptionBaseEvent> getEventsOnCancelPlan(final DefaultSubscripti
                                                              final InternalTenantContext internalTenantContext) throws CatalogApiException;
 
     boolean undoChangePlan(DefaultSubscriptionBase defaultSubscriptionBase, CallContext context) throws SubscriptionBaseApiException;
+
+    int getBCD(InternalTenantContext context) throws SubscriptionBaseApiException;
```
pierre, line 111: Rename it to `getAccountBCD`?

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -687,6 +690,30 @@ public List<SubscriptionBillingEvent> getSubscriptionBillingEvents(final Version
         }
     }
 
+
+    private DateTime alignToNextBCD(final Plan curPlan, final PlanPhase curPlanPhase, final DateTime originalDate, final SubscriptionCatalog catalog, final Integer bcdLocal, final InternalTenantContext context) throws SubscriptionBaseApiException, CatalogApiException {
```
pierre, line 694: `bcdLocal` -> `catalogChangeEffectiveDate`

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -687,6 +690,30 @@ public List<SubscriptionBillingEvent> getSubscriptionBillingEvents(final Version
...
+
+        final BillingAlignment billingAlignment = catalog.billingAlignment(new PlanPhaseSpecifier(curPlan.getName(), curPlanPhase.getPhaseType()),
```
pierre, line 696: This is not new (I think it came up when we were doing https://github.com/killbill/killbill/issues/784), but it's not clear whether we want the `prev` or `next` plan here (e.g. after change plan) to find the billing alignment.

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -687,6 +690,30 @@ public List<SubscriptionBillingEvent> getSubscriptionBillingEvents(final Version
...
+                                                                           originalDate, originalDate);
+
+        final int accountBillCycleDayLocal = apiService.getBCD(context);
+        Integer bcd = bcdLocal;
+        if (bcd == null) {
```
pierre, line 701: Cannot be `null`

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
sbrossie: I think in this case, the problem does not exist as we are not going from a `prev` -> to a `next` (this would happen if we were to look for 'change of plan' alignment rules).  In this case, given the `cur` Plan/Phase return the billing alignment.

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
sbrossie: What about [this](https://github.com/killbill/killbill/pull/1663/files#diff-6c9485b2174859a0d155b69e387e4b868e364c2608ba08fd0a2db9c49101fc46R645) 🤔 ?

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
sbrossie: What about [this](https://github.com/killbill/killbill/pull/1663/files#diff-6c9485b2174859a0d155b69e387e4b868e364c2608ba08fd0a2db9c49101fc46R645) 🤔 ?

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
sbrossie: Are you suggesting I change `originalDate` -> `catalogChangeEffectiveDate` or `bcdLocal` -> `catalogChangeEffectiveDate` ?

subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
sbrossie: Fixed in https://github.com/killbill/killbill/pull/1663/commits/d5fb66fb4deb852b0f4e712d8829f0824e81d57a

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
pierre: `bcdLocal` -> `catalogChangeEffectiveDate` -- it's not a _bcd_? 🤔

pierre: humm - scratch that. I'm wondering if I read an earlier version of the code (see next comment).

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
pierre: Weird, I remember reading a different piece of code 🙃  In this case, it's always `null`?

sbrossie: I think the caller needs to maintain the current per-subscription BCD as it goes through billing event, which is why I have this parameter, but right not implemented - let's keep this open.

subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
```diff
@@ -107,4 +108,9 @@ public List<SubscriptionBaseEvent> getEventsOnCancelPlan(final DefaultSubscripti
                                                              final InternalTenantContext internalTenantContext) throws CatalogApiException;
 
     boolean undoChangePlan(DefaultSubscriptionBase defaultSubscriptionBase, CallContext context) throws SubscriptionBaseApiException;
+
+    int getAccountBCD(InternalTenantContext context) throws SubscriptionBaseApiException;
+
+    // If we expose more than one config, we should not instead return the SubscriptionConfig object
```
pierre, line 114: Is this a question? 

BTW I don't think you need the `@Param("dummy")` here.

subscription/src/main/java/org/killbill/billing/subscription/config/MultiTenantSubscriptionConfig.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public boolean isEffectiveDateForExistingSubscriptionsAlignedToBCD() {
+        return staticConfig.isEffectiveDateForExistingSubscriptionsAlignedToBCD();
+    }
+
+    @Override
+    public boolean isEffectiveDateForExistingSubscriptionsAlignedToBCD(final InternalTenantContext tenantContext) {
+        final String result = getStringTenantConfig("isEffectiveDateForExistingSubscriptionsAlignedToBCD", tenantContext);
```
pierre, line 47: We'll need to update the config docs eventually (/cc @reshmabidikar).

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -687,6 +689,34 @@ public List<SubscriptionBillingEvent> getSubscriptionBillingEvents(final Version
...
+                                                                           originalDate, originalDate);
+
+        final int accountBillCycleDayLocal = apiService.getAccountBCD(context);
+        Integer bcd = bcdLocal;
+        if (bcd == null) {
+            bcd = BillCycleDayCalculator.calculateBcdForAlignment(null, this, this, billingAlignment, context, accountBillCycleDayLocal);
+        }
+
+        // TODO : no Recurring PlanPhase, Usage sections, ...
+        final BillingPeriod billingPeriod = curPlanPhase.getRecurring() != null ? curPlanPhase.getRecurring().getBillingPeriod() : BillingPeriod.NO_BILLING_PERIOD;
```
pierre, line 709: Interesting we don't have a utility method to encapsulate "END_OF_TERM" already. Worth considering maybe?

pierre (approved):

subscription/src/main/java/org/killbill/billing/subscription/config/MultiTenantSubscriptionConfig.java
reshmabidikar: Noted, I will include this change when I update the docs next.

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
sbrossie: I implemented it in https://github.com/killbill/killbill/pull/1663/commits/4479e18ee70a81cf8aca0310ec48884d0daf3768

subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionBaseApiService.java
sbrossie: Fixed the comment and removed the unused dummy param

pierre (approved):

PR #1664, opened 2022-04-19
util: support for mariadb-java-client 3.0.4

pierre:
See https://github.com/killbill/killbill-oss-parent/pull/366 and https://github.com/killbill/killbill-commons/pull/116.

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
+++ b/util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
@@ -22,6 +22,7 @@ import java.sql.Clob;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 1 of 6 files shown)


pierre:
Note: e2e test `KillBillIntegrationTests::TestWithDatesAndTimezones` is failing as expected.

util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
```diff
@@ -94,21 +95,32 @@ public void exportDataForAccount(final DatabaseExportOutputStream out, final Int
         }
 
         final List<ColumnInfo> columnsForTable = new ArrayList<ColumnInfo>();
+        // Separate lookup table, to keep the ordering of the columns
+        final Map<String, Integer> columnsLookup = new TreeMap<>(String.CASE_INSENSITIVE_ORDER); // Ignore casing (for H2)
```
sbrossie, line 99: Why do we need to sort the keys -> isn't the point of the value (the index) to specify which column comes first?

util/src/test/java/org/killbill/billing/util/export/dao/TestDatabaseExportDao.java
```diff
@@ -86,7 +86,7 @@ public Void withHandle(final Handle handle) throws Exception {
         final String newDump = getDump();
 
         Assert.assertEquals(newDump, "-- accounts record_id|id|external_key|email|name|first_name_length|currency|billing_cycle_day_local|parent_account_id|is_payment_delegated_to_parent|payment_method_id|reference_time|time_zone|locale|address1|address2|company_name|city|state_or_province|country|postal_code|phone|notes|migrated|created_date|created_by|updated_date|updated_by|tenant_record_id\n" +
-                                     String.format("%s|%s|%s|%s|%s|%s||||false||%s|%s|||||||||||false|%s|%s|%s|%s|%s", internalCallContext.getAccountRecordId(), accountId, accountId, accountEmail, accountName, firstNameLength, "1970-05-24T18:33:02.000+00:00", timeZone,
+                                     String.format("%s|%s|%s|%s|%s|%s||||true||%s|%s|||||||||||false|%s|%s|%s|%s|%s", internalCallContext.getAccountRecordId(), accountId, accountId, accountEmail, accountName, firstNameLength, "1970-05-24T18:33:02.000+00:00", timeZone,
```
sbrossie, line 89: 🙄

util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
pierre: We actually don't need the ordering, we just need a map with case-insensitive keys.

sbrossie: Right, so technically you don't need a `TreeMap` (but not that it matters)

PR #1673, opened 2022-04-28
subscription: Rework event ordering to ensure we have always a 'start…

sbrossie:
…' event first

The previous ordering logic to broke ties between events having the exact same effective date
based on the underlying `record_id`. The issue with this approach is that our [current changePlan logic](https://github.com/killbill/killbill/blob/killbill-0.22.28/subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java#L752) rewrites events and a result a new 'start' event may end up being inserted with a greater `record_id`.

Example of what happened with events all having exact same effective date:

Before change operation
* `CREATE`/`TRANSFER`, `BCD_UPDATE`

After change operation
* `CREATE`/`TRANSFER` (inactive), `BCD_UPDATE`, `CREATE`/`TRANSFER` = `BCD_UPDATE`, `CREATE`/`TRANSFER`

So the first event is now a  `BCD_UPDATE`

This would lead to the following trace:

```
Caused by: java.lang.NullPointerException: null
	at org.killbill.billing.subscription.api.user.DefaultSubscriptionBase.getSourceType(DefaultSubscriptionBase.java:197)
	at org.killbill.billing.entitlement.api.DefaultEntitlement.getSourceType(DefaultEntitlement.java:253)
	at org.killbill.billing.jaxrs.json.SubscriptionJson.<init>(SubscriptionJson.java:389)
	at org.killbill.billing.jaxrs.resources.SubscriptionResource.getSubscription(SubscriptionResource.java:173)
	at org.killbill.billing.jaxrs.resources.SubscriptionResource_$$_jvst5f5_2._d46getSubscription(SubscriptionResource_$$_jvst5f5_2.java)
	at jdk.internal.reflect.GeneratedMethodAccessor539.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.glassfish.hk2.utilities.reflection.ReflectionHelper.invoke(ReflectionHelper.java:1268)
	at org.jvnet.hk2.internal.MethodInterceptorHandler$MethodInvocationImpl.proceed(MethodInterceptorHandler.java:164)
	at org.killbill.commons.skeleton.metrics.TimedResourceInterceptor.invoke(TimedResourceInterceptor.java:70)
	at org.jvnet.hk2.internal.MethodInterceptorHandler.invoke(MethodInterceptorHandler.java:97)
	at org.killbill.billing.jaxrs.resources.SubscriptionResource_$$_jvst5f5_2.getSubscription(SubscriptionResource_$$_jvst5f5_2.java)
	at jdk.internal.reflect.GeneratedMethodAccessor538.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory.lambda$static$0(ResourceMethodInvocationHandlerFactory.java:52)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:124)
	... 89 common frames omitted
```

The code now takes into account the type of events to ensure we also have a 'start' event first.

Also, the ordering logic is now embedded in the `SubscriptionBaseEvent` so we can imply call `Collections.sort()` from various places.

Additionally, we also sort events inside the SubscriptionDao to ensure that looking at events in this layer provides the right view (that was another bug).

TODO:
* Should the sorting logic be stricter at ordering ties for variuous types such as `BCD_UPDATE`, `PHASE`, `CHANGE_PLAN`, i.e is any combination of those leading to the same result?
* Should we have a  proxy `SubscriptionEventSqlDao` layer that ensures the ordering instead of forcing caller to always call sort?
* Am i missing sort operations?
* Could we extract the sort for the entitlement module as well?

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
@@ -532,10 +532,10 @@ public class DefaultSubscriptionDao extends EntityDaoBase<SubscriptionBundleMode
         transactionalSqlDao.execute(false, new EntitySqlDaoTransactionWrapper<Void>() {
             @Override
             public Void inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFactory) throws Exception {
-                final SubscriptionEventSqlDao transactional = entitySqlDaoWrapperFactory.become(SubscriptionEventSqlDao.class);
+                final SubscriptionEventSqlDao eventSqlDao = entitySqlDaoWrapperFactory.become(SubscriptionEventSqlDao.class);
                 final UUID subscriptionId = subscription.getId();
                 cancelNextPhaseEventFromTransaction(subscriptionId, entitySqlDaoWrapperFactory, context);
-                createAndRefresh(transactional, new SubscriptionEventModelDao(nextPhaseEvent), context);
+                createAndRefresh(eventSqlDao, new SubscriptionEventModelDao(nextPhaseEvent), context);
                 recordFutureNotificationFromTransaction(entitySqlDaoWrapperFactory,
                                                         nextPhaseEvent.getEffectiveDate(),
                                                         new SubscriptionNotificationKey(nextPhaseEvent.getId()), context);
@@ -554,7 +554,8 @@ public class DefaultSubscriptionDao extends EntityDaoBase<SubscriptionBundleMode
         return transactionalSqlDao.execute(true, new EntitySqlDaoTransactionWrapper<SubscriptionBaseEvent>() {
             @Override
             public SubscriptionBaseEvent inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFactory) throws Exception {
-                final SubscriptionEventModelDao model = entitySqlDaoWrapperFactory.become(SubscriptionEventSqlDao.class).getById(eventId.toString(), context);
+                final SubscriptionEventSqlDao eventSqlDao = entitySqlDaoWrapperFactory.become(SubscriptionEventSqlDao.class);
+                final SubscriptionEventModelDao model = eventSqlDao.getById(eventId.toString(), context);
                 return SubscriptionEventModelDao.toSubscriptionEvent(model);
             }
         });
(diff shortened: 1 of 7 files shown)


subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
```diff
@@ -577,7 +578,9 @@ public List<SubscriptionBaseEvent> getPendingEventsForSubscription(final UUID su
         return transactionalSqlDao.execute(true, new EntitySqlDaoTransactionWrapper<List<SubscriptionBaseEvent>>() {
             @Override
             public List<SubscriptionBaseEvent> inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFactory) throws Exception {
-                final List<SubscriptionEventModelDao> eventModels = entitySqlDaoWrapperFactory.become(SubscriptionEventSqlDao.class).getFutureActiveEventForSubscription(subscriptionId.toString(), now, context);
+                final SubscriptionEventSqlDao eventSqlDao = entitySqlDaoWrapperFactory.become(SubscriptionEventSqlDao.class);
+                final List<SubscriptionEventModelDao> eventModels = eventSqlDao.getFutureActiveEventForSubscription(subscriptionId.toString(), now, context);
```
pierre, line 582: Nit: I suspect changing the DAO methods to return a `SortedSet` instead would return an ordered collection automagically. That way, no need to call manually `Collections.sort` each time.

pierre (approved):
Might be worth adding a (fast?) regression test with your example above if it's not too hard...

> * Should we have a  proxy `SubscriptionEventSqlDao` layer that ensures the ordering instead of forcing caller to always call sort?

See my comment. See also https://github.com/killbill/killbill-commons/blob/bc5077bf4f69f3600c56528917813bd5091d028d/jdbi/src/main/java/org/skife/jdbi/v2/ContainerFactoryRegistry.java#L99

> * Should the sorting logic be stricter at ordering ties for variuous types such as `BCD_UPDATE`, `PHASE`, `CHANGE_PLAN`, i.e is any combination of those leading to the same result?
> * Am i missing sort operations?
> * Could we extract the sort for the entitlement module as well?

I would create a separate task to track this. Consolidating the sort operations in both modules and adding new fast unit tests on the comparable methods would provide some validation.

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
sbrossie: Great point. Fixed in https://github.com/killbill/killbill/pull/1673/commits/2e16492b0b219d98ec28f4139a8768d9ea43e0fc

sbrossie: Also created [GH issue] to track my questions.

PR #1679, opened 2022-05-03
Work for release 0.23 jax rs change

reshmabidikar:
- Changes to return `DateTime` instead of `LocalDate` 
- Fixed tests that fail due to the above change


pierre:
FYI I've merged this into https://github.com/killbill/killbill/pull/1680.

PR #1680, opened 2022-05-03
ci: point e2e tests to work-for-release-0.23.x

pierre:
Verifying https://github.com/killbill/killbill-integration-tests/pull/50.

### DIFF
--- a/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestUsage.java
+++ b/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestUsage.java
@@ -42,8 +42,6 @@ import org.killbill.billing.util.api.AuditLevel;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
-import com.google.common.collect.ImmutableMap;
-
 public class TestUsage extends TestJaxrsBase {
 
     private  static Subscriptions createSubscriptions(final Account accountJson) {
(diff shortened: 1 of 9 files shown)


profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestUsage.java
```diff
@@ -42,6 +42,8 @@
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+import com.google.common.collect.ImmutableMap;
```
pierre, line 45: @reshmabidikar We're trying to move away from Guava. @xsalefter can help with the transition if needed, although I don't think the import is needed.

pierre (commented):
@reshmabidikar Can this be merged? Let me know if the other PRs are still needed.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestUsage.java
xsalefter: Yes, on this file, `ImmutableMap` is not needed, so perhaps just remove it? 

On other files, we're migrating guava's collection to Java's stream API, as described in issue #1615. But if it's too much for you, just leave it as is. Currently, I'm working on `killbill-invoice` module, so naturally you would see a lot of `import com.google.common.<XXX>` statement in `killbill-jaxrs`.

PR #1681, opened 2022-05-03
Add info logging for account lock

sbrossie:


invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -215,14 +215,19 @@ public void processSubscriptionStartRequestedDate(final RequestedSubscriptionInt
 
         GlobalLock lock = null;
         try {
+            log.info("Account Lock requested for account {}", accountId.toString());
```
pierre, line 218: Since the message is the same each time, I'm not sure this will be enough for debugging. We might need the name of the method (e.g. `[InvoiceDispatcher#processSubscriptionStartRequestedDate] Account Lock requested for account {}`).

Also, if we want to eventually merge this, could we use `debug` level instead? It's easy for the user to toggle at runtime through JMX the level of individual loggers (ignore if we won't be merging this).

pierre (changes requested):

sbrossie:
Closing PR - it was used for debugging mostly.

PR #1687, opened 2022-05-09
CR Integration

sbrossie:
Addresses remaining comments from https://github.com/killbill/killbill/pull/1676

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java
@@ -280,7 +280,7 @@ public class DefaultInvoiceUserApi implements InvoiceUserApi {
         if (result.isEmpty()) {
             throw new InvoiceApiException(ErrorCode.INVOICE_NOTHING_TO_DO, accountId, targetDate != null ? targetDate : "null");
         } else {
-            Preconditions.checkState(result.size() == 1);
+            Preconditions.checkState(result.size() == 1, String.format("Expected one invoice but got %d", result.size()));
             return result.get(0);
         }
     }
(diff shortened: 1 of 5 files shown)


invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java
```diff
@@ -280,7 +280,7 @@ public Invoice triggerDryRunInvoiceGeneration(final UUID accountId, final LocalD
         if (result.isEmpty()) {
             throw new InvoiceApiException(ErrorCode.INVOICE_NOTHING_TO_DO, accountId, targetDate != null ? targetDate : "null");
         } else {
-            Preconditions.checkState(result.size() == 1);
+            Preconditions.checkState(result.size() == 1, String.format("Expected one invoice but got %d", result.size()));
```
pierre, line 283: You don't need the `String.format`, check https://github.com/killbill/killbill/blob/812ddcc7aaed8b1e0b0b75016b590fb553da972a/util/src/main/java/org/killbill/billing/util/Preconditions.java#L62

I would actually lean towards removing this method https://github.com/killbill/killbill/blob/812ddcc7aaed8b1e0b0b75016b590fb553da972a/util/src/main/java/org/killbill/billing/util/Preconditions.java#L206-L210 so we never forget to put a message. @xsalefter thoughts?

pierre (approved):

invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java
xsalefter: > .... so we never forget to put a message.

I agree. Let me check that `checkState(boolean)` usage across the code base. Maybe for this time we just `@Deprecated` it?

pierre: Sure -- my idea is that we should always add a message with enough context to debug why the precondition was hit.

sbrossie: Fixed in https://github.com/killbill/killbill/pull/1687/commits/d454e21b8445b116f7520126981093f4af2be90d

PR #1691, opened 2022-05-13
1375 Step 3 - Added new method for cancel with DateTime

reshmabidikar:


entitlement/src/main/java/org/killbill/billing/entitlement/logging/EntitlementLoggingHelper.java
```diff
@@ -158,7 +159,7 @@ public static void logTransferEntitlement(final Logger log,
         }
     }
 
-    public static void logCancelEntitlement(final Logger log, final Entitlement entitlement, final LocalDate entitlementEffectiveDate, final Boolean overrideBillingEffectiveDate, final EntitlementActionPolicy entitlementPolicy, final BillingActionPolicy billingPolicy) {
+    public static void logCancelEntitlement(final Logger log, final Entitlement entitlement, final LocalDate entitlementEffectiveDate, final Boolean overrideBillingEffectiveDate, final EntitlementActionPolicy entitlementPolicy, final BillingActionPolicy billingPolicy, final DateTime entitlementEffectiveDateTime, final DateTime billingEffectiveDateTime) {
```
pierre, line 162: Instead of introducing new arguments, maybe use `Serializable` for the dates:

```
[...], final Object entitlementEffectiveDate, final Object billingEffectiveDateTime, final Boolean overrideBillingEffectiveDate, [...]
```

I couldn't find a common ancestor between `LocalDate` and `DateTime` except `Serializable` unfortunately.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
```diff
@@ -375,6 +375,55 @@ public Entitlement doCall(final EntitlementApi entitlementApi, final DefaultEnti
...
+        checkForPermissions(Permission.ENTITLEMENT_CAN_CANCEL, callContext);
+        // Get the latest state from disk
+        refresh(callContext);
+        if (entitlementEffectiveDate != null && entitlementEffectiveDate.compareTo(getEffectiveStartDate()) < 0) {
+            throw new EntitlementApiException(ErrorCode.SUB_INVALID_REQUESTED_DATE, entitlementEffectiveDate, getEffectiveStartDate());
+        }
+        if (billingEffectiveDate != null && billingEffectiveDate.compareTo(getSubscriptionBase().getStartDate()) < 0) { //TODO_1375 - Added this check for billingEffectiveDate similar to the check above for entitlementEffectiveDate. Is the comparison to getSubscriptionBase().getStartDate() correct?
+            throw new EntitlementApiException(ErrorCode.SUB_INVALID_REQUESTED_DATE, entitlementEffectiveDate, getEffectiveStartDate());
+        }
+        //TODO_1375 - Is a null check required here for entitlementEffectiveDate and billingEffectiveDate?
```
pierre, line 393: I would add one, in the `if` above, e.g.

```
if (entitlementEffectiveDate == null || (entitlementEffectiveDate != null && entitlementEffectiveDate.compareTo(getEffectiveStartDate()) < 0))
```

reshmabidikar: Fixed

entitlement/src/main/java/org/killbill/billing/entitlement/logging/EntitlementLoggingHelper.java
reshmabidikar: Fixed

entitlement/src/main/java/org/killbill/billing/entitlement/logging/EntitlementLoggingHelper.java
```diff
@@ -158,7 +159,7 @@ public static void logTransferEntitlement(final Logger log,
         }
     }
 
-    public static void logCancelEntitlement(final Logger log, final Entitlement entitlement, final LocalDate entitlementEffectiveDate, final Boolean overrideBillingEffectiveDate, final EntitlementActionPolicy entitlementPolicy, final BillingActionPolicy billingPolicy) {
+    public static void logCancelEntitlement(final Logger log, final Entitlement entitlement, final Serializable entitlementEffectiveDate, final Serializable billingEffectiveDate, final Boolean overrideBillingEffectiveDate, final EntitlementActionPolicy entitlementPolicy, final BillingActionPolicy billingPolicy) {
```
sbrossie, line 162: I see.. the only common interface between `LocalDate` and `DateTime` is `Serializable` (which does not mean much) but works here as we just call `toString()`. So in this case we might as well to pass a java `Object` here which is the one defining the `toString()` ?

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlement.java
```diff
@@ -71,6 +71,98 @@ public void testCancelWithEntitlementDate() throws AccountApiException, Entitlem
...
+
+        final DateTime cancelDateTime = new DateTime(clock.getUTCNow().plusDays(5));
+        entitlement.cancelEntitlementWithDate(cancelDateTime, cancelDateTime, Collections.emptyList(), callContext);
+        assertListenerStatus();
+
+        clock.addDays(5);
+        testListener.pushExpectedEvents(NextEvent.CANCEL, NextEvent.BLOCK);
+        final Entitlement entitlement2 = entitlementApi.getEntitlementForId(entitlement.getId(), callContext);
+        assertEquals(entitlement2.getState(), EntitlementState.CANCELLED);
+        assertEquals(internalCallContext.toLocalDate(entitlement2.getEffectiveEndDate()), clock.getUTCToday());
```
sbrossie, line 130: Can we do better and check for the exact timestamp (DateTime) to verify the provided value was indeed honored?

sbrossie: Also, can we check both billing and entitlement canceled date(time) ?

reshmabidikar: > Can we do better and check for the exact timestamp (DateTime) to verify the provided value was indeed honored?

 Earlier we had decided to compare dates (and not DateTime) in the tests to prevent test flakiness.  Since we are cancelling with an exact `DateTime` in this case, I _think_ the tests would not be flaxy even if we compare `DateTime`. However, this would make the tests look different from all the other tests (that compare `LocalDate`).

Let me know what you think

sbrossie: > Earlier we had decided to compare dates (and not DateTime) in the tests to prevent test flakiness. 

Yes, because we had existing tests that were *only* comparing (local) dates as our models/apis did not support DateTime

However, now we are introducing new apis to support DateTime, and so the new tests should verify the equality of the DateTime (not just the local dates).

reshmabidikar: Agreed, yes, makes sense, I will change these tests to compare DateTime 👍

entitlement/src/main/java/org/killbill/billing/entitlement/logging/EntitlementLoggingHelper.java
reshmabidikar: Fixed

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlement.java
reshmabidikar: Fixed, added check for billing and entitlement cancelled date time and changed test to compare DateTime

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlement.java
```diff
@@ -71,6 +71,106 @@ public void testCancelWithEntitlementDate() throws AccountApiException, Entitlem
...
+        clock.addDays(5);
+
+        testListener.pushExpectedEvents(NextEvent.CANCEL, NextEvent.BLOCK);
+        final DateTime cancelDateTime = new DateTime(clock.getUTCNow());
+        entitlement.cancelEntitlementWithDate(cancelDateTime, cancelDateTime, Collections.emptyList(), callContext);
+        assertListenerStatus();
+
+        final Entitlement entitlement2 = entitlementApi.getEntitlementForId(entitlement.getId(), callContext);
+        assertEquals(entitlement2.getState(), EntitlementState.CANCELLED);
+        assertEquals(entitlement2.getEffectiveEndDate(), cancelDateTime);
```
pierre, line 101: Better to use `assertEquals(entitlement2.getEffectiveEndDate().compareTo(cancelDateTime), 0);` for these, to avoid tests flakiness.

pierre (approved):
Just one comment, otherwise :shipit:

sbrossie:
@reshmabidikar Merging your PR, you can open a new one to address [this comment](https://github.com/killbill/killbill/pull/1691#discussion_r873834974) or factor in in your next PR

sbrossie (approved):

reshmabidikar:
Thanks, @sbrossie for merging, I will note the review comment and address it in my next PR.

PR #1693, opened 2022-05-17
1615 invoice part 5

xsalefter:
Last PR for `killbill-invoice`. Most of work is Guava to stream, except adding `BiMap` and its `SimpleHashBiMap` implementation. Decided to implements simple implementation because forking google BiMap implementation require a lot of Guava's classes to added as well.

I think what I've done cover all basic requirement of BiMap, there's [test case](https://github.com/killbill/killbill/pull/1693/files#diff-d3f24acffa59da2471175404c40c16556f355410f74c026d5c31db4a2cb99829) for it, and also I have [gist](https://gist.github.com/xsalefter/37eb938888a4bc9514ad0fa7e3dd59dd) for comparing with Guava's version.


util/src/main/java/org/killbill/billing/util/collect/SimpleHashBiMap.java
```diff
@@ -0,0 +1,158 @@
...
+            if (!isRemoved) {
+                throw new IllegalStateException("Illegal BiMap#focePut() behaviour. Remove operation failed");
+            }
+        }
+        return super.put(key, value);
+    }
+
+    @Override
+    public BiMap<V, K> inverse() {
+        return this.keySet()
```
pierre, line 68: It looks like we're only using BiMap in tests, so this is probably fine, but I suspect the Guava version is O(1) while this is O(n). Might be worth adding a comment in case we decide to use this for production code at some point.

invoice/src/test/java/org/killbill/billing/invoice/usage/TestContiguousIntervalConsumableInArrear.java
```diff
@@ -51,28 +54,14 @@
...
-import com.google.common.collect.Lists;
-import com.google.common.collect.Ordering;
 
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertTrue;
 
 public class TestContiguousIntervalConsumableInArrear extends TestUsageInArrearBase {
 
-    // Only works if the RolledUpUsage have at least one
```
pierre, line 68: I think the comment is still valid.

pierre (approved):

pierre:
I've restarted the job. Could you create a GitHub issue with the details, so that we don't forget to investigate the source of the flakiness?

sbrossie (commented):
It looks like `BiMap` and its new implementation `SimpleHashBiMap` is only referenced once from the `MockInvoiceDao` - is this accurate? The `MockInvoiceDao` probably manages very few entries for a couple of tests and so do we really need this additional complexity, i/e couldn't we we just have the `MockInvoiceDao` reverse the map when needed ?

xsalefter:
@sbrossie yeah I think I agree with you. I just delete `BiMap` and its implementation, and use internal private method to find map key by its value.

xsalefter:
@sbrossie @pierre Same problem as described in #1694 . I added [new comment](https://github.com/killbill/killbill/issues/1694#issuecomment-1129725114) about my observation.

Issue #1471, opened 2021-08-05
Ability to pass plugin properties to invoice plugin


sbrossie:
When triggering an invoice run -- dry-run or not -- allow the caller to specify plugin properties that would be passed to the plugin.

PR #1695, opened 2022-05-18
Add ability to pass plugin properties on invoice runs. See #1471

sbrossie:
Depends on https://github.com/killbill/killbill-api/pull/92

We also need regenerate the java client to complete the work, but the jaxrs apis are not broken as the plugin properties are passed as query parameters. The java client signatures would change though.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
@@ -94,7 +94,7 @@ public class InvoicePluginDispatcher {
                               final boolean isRescheduled,
                               final CallContext callContext,
                               // The pluginProperties list passed to plugins is mutable by the plugins
-                              @SuppressWarnings("TypeMayBeWeakened") final LinkedList<PluginProperty> properties,
+                              @SuppressWarnings("TypeMayBeWeakened") final Iterable<PluginProperty> properties,
                               final InternalTenantContext internalTenantContext) throws InvoiceApiException {
         log.debug("Invoking invoice plugins priorCall: targetDate='{}', isDryRun='{}', isRescheduled='{}'", targetDate, isDryRun, isRescheduled);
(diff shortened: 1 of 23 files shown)


invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
```diff
@@ -94,7 +94,7 @@ public DateTime priorCall(final LocalDate targetDate,
                               final boolean isRescheduled,
                               final CallContext callContext,
                               // The pluginProperties list passed to plugins is mutable by the plugins
-                              @SuppressWarnings("TypeMayBeWeakened") final LinkedList<PluginProperty> properties,
+                              @SuppressWarnings("TypeMayBeWeakened") final Iterable<PluginProperty> properties,
```
pierre, line 97: NACK. There's a reason why we have a `@SuppressWarnings("TypeMayBeWeakened")` here 😄 

See https://github.com/killbill/killbill/pull/1333/files

pierre (changes requested):

invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java
sbrossie: See https://github.com/killbill/killbill/pull/1696

sbrossie: PR has been merged and should address the comment.

pom.xml
```diff
@@ -59,8 +59,10 @@
         <url>http://github.com/killbill/killbill</url>
     </issueManagement>
     <properties>
+        <check.fail-enforcer>false</check.fail-enforcer>
```
pierre, line 62: You can use parent pom `0.145.3-e8112a5-SNAPSHOT`.

pierre (approved):

pom.xml
sbrossie: Fixed in https://github.com/killbill/killbill/pull/1695/commits/bb4bf4dd1168409ec025644a41307fb43aab2f85

Issue #1312, opened 2020-06-15
Inconsistent behavior in our InvoicePluginApi#priorCall


sbrossie:
For the payment control plugin api, we allow plugins to modify plugin properties passed in the `priorCall`, in such a way that next plugins in the list would see the properties introduced by the prev. plugin.

However, for invoice plugin, we have not implemented this and this seems a bit inconsistent, so we should probably fix it.

PR #1696, opened 2022-05-18
invoice: Ability to have invoice plugins modify plugin properties. 

sbrossie:
See #1312

This has been implemented for priorCall and completion apis (similar to payment control). See https://github.com/killbill/killbill-plugin-api/pull/23

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
@@ -37,6 +37,7 @@ import org.killbill.billing.ErrorCode;
 import org.killbill.billing.callcontext.InternalCallContext;
 import org.killbill.billing.catalog.api.Currency;
 import org.killbill.billing.invoice.InvoicePluginDispatcher;
+import org.killbill.billing.invoice.InvoicePluginDispatcher.PriorCallResult;
 import org.killbill.billing.invoice.dao.InvoiceDao;
 import org.killbill.billing.invoice.dao.InvoiceItemModelDao;
 import org.killbill.billing.invoice.dao.InvoiceModelDao;
@@ -86,8 +87,8 @@ public class InvoiceApiHelper {
         final InternalCallContext internalCallContext = internalCallContextFactory.createInternalCallContext(accountId, contextMaybeWithoutAccountId);
         final CallContext context = internalCallContextFactory.createCallContext(internalCallContext);
 
-        final DateTime rescheduleDate = invoicePluginDispatcher.priorCall(targetDate, existingInvoices, isDryRun, isRescheduled, context, properties, internalCallContext);
-        if (rescheduleDate != null) {
+        final PriorCallResult priorCallResult = invoicePluginDispatcher.priorCall(targetDate, existingInvoices, isDryRun, isRescheduled, context, properties, internalCallContext);
+        if (priorCallResult.getRescheduleDate() != null) {
             throw new InvoiceApiException(ErrorCode.INVOICE_PLUGIN_API_ABORTED, "delayed scheduling is unsupported for API calls");
         }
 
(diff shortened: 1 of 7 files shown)


pierre (approved):

sbrossie:
See https://github.com/killbill/killbill/pull/1696/commits/b3a37ee4cea86fc88a18dbc720b953f174102860 to match plugin api change requested in https://github.com/killbill/killbill-plugin-api/pull/23#issuecomment-1132009980

Also see https://github.com/killbill/killbill/pull/1696/commits/61c4cbf7715fd96e0750f487557e65e3694befc4 got have the build go through - probably we don't want to merge this piece. I suppose I have to recreate a plugin api snapshot using api snapshot?

sbrossie:
See https://github.com/killbill/killbill/pull/1696/commits/b3a37ee4cea86fc88a18dbc720b953f174102860 to match plugin api change requested in https://github.com/killbill/killbill-plugin-api/pull/23#issuecomment-1132009980

Also see https://github.com/killbill/killbill/pull/1696/commits/61c4cbf7715fd96e0750f487557e65e3694befc4 got have the build go through - probably we don't want to merge this piece. I suppose I have to recreate a plugin api snapshot using api snapshot?

invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
```diff
@@ -129,10 +137,10 @@ public List<InvoiceItem> dispatchToInvoicePluginsAndInsertItems(final UUID accou
             if (success) {
                 for (final Invoice invoiceForPlugin : invoicesForPlugins) {
                     final DefaultInvoice refreshedInvoice = new DefaultInvoice(dao.getById(invoiceForPlugin.getId(), internalCallContext));
-                    invoicePluginDispatcher.onSuccessCall(targetDate, refreshedInvoice, existingInvoices, isDryRun, isRescheduled, context, priorCallResult.getPluginProperties(), internalCallContext);
```
pierre: Don't we need to keep track of the adjusted properties and pass them to the next plugin?

invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
```diff
@@ -103,7 +109,9 @@ public List<InvoiceItem> dispatchToInvoicePluginsAndInsertItems(final UUID accou
             final List<InvoiceModelDao> invoiceModelDaos = new LinkedList<InvoiceModelDao>();
             for (final DefaultInvoice invoiceForPlugin : invoicesForPlugins) {
                 // Call plugin(s)
-                invoicePluginDispatcher.updateOriginalInvoiceWithPluginInvoiceItems(invoiceForPlugin, isDryRun, context, priorCallResult.getPluginProperties(), internalCallContext);
+                final AdditionalInvoiceItemsResult itemsResult = invoicePluginDispatcher.updateOriginalInvoiceWithPluginInvoiceItems(invoiceForPlugin, isDryRun, context, pluginProperties, internalCallContext);
+                // Could be a bit weird for a plugin to keep updating properties for each invoice
```
pierre, line 113: FWIW that only works because the plugins are called in sequence for each invoice. In other words, for each invoice iteration, the first plugin would add the right plugin properties for taxation (based on each invoice item id) for the next invoice plugin, before moving on to the next invoice.

With this in mind, wondering then if we need to _clear_ the adjusted properties between each invoice?

pierre:
> I suppose I have to recreate a plugin api snapshot using api snapshot?

See https://github.com/killbill/killbill-oss-parent/blob/work-for-release-0.23.x/pom.xml#L1889

We are missing the various entries for `org.kill-bill.billing.plugin` artifacts.

invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
sbrossie: The model is to rely the union of all the properties (input + adjusted). I don't think the Kill Bill core should clear anything, but the plugin could of course decide to clear things (if it knows the consequences)

invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
sbrossie: Not sure if this comment is still valid: For reference, we are keeping track of the adjusted properties in `pluginProperties` and this is what we pass for each plugin invocation.

PR #1705, opened 2022-05-31
1615 Beatrix part 1

xsalefter:
Part 1 of 3.

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestHardenCatalogPlugin.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestHardenCatalogPlugin.java
@@ -17,10 +17,11 @@
 
 package org.killbill.billing.beatrix.integration;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
-import javax.annotation.Nullable;
 import javax.inject.Inject;
 
 import org.joda.time.DateTime;
@@ -34,10 +35,7 @@ import org.killbill.billing.catalog.DefaultVersionedCatalog;
 import org.killbill.billing.catalog.StandaloneCatalog;
 import org.killbill.billing.catalog.StandaloneCatalogWithPriceOverride;
 import org.killbill.billing.catalog.api.CatalogUserApi;
-import org.killbill.billing.catalog.api.Plan;
 import org.killbill.billing.catalog.api.PlanPhaseSpecifier;
-import org.killbill.billing.catalog.api.PriceList;
-import org.killbill.billing.catalog.api.Product;
 import org.killbill.billing.catalog.api.StaticCatalog;
 import org.killbill.billing.catalog.override.PriceOverride;
 import org.killbill.billing.catalog.plugin.TestModelStandalonePluginCatalog;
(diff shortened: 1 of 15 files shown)


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestHardenCatalogPlugin.java
```diff
@@ -265,25 +259,19 @@ public void reset() {
...
-                private <I, C extends I> List<I> listOf(@Nullable C[] input) {
-                    return (input != null) ? ImmutableList.<I>copyOf(input) : ImmutableList.<I>of();
-                }
-            });
+            return input.stream()
+                    .map(staticCatalog -> {
+                        final StandaloneCatalog standaloneCatalog = (StandaloneCatalog) staticCatalog;
+                        return new TestModelStandalonePluginCatalog(new DateTime(staticCatalog.getEffectiveDate()),
+                                                                    List.of(standaloneCatalog.getSupportedCurrencies()),
+                                                                    List.copyOf(standaloneCatalog.getProducts()),
```
pierre, line 267: `of` vs `copyOf` on purpose?

Also, in the past, you've used `Iterables.toUnmodifiableList` (https://github.com/killbill/killbill/pull/1697#discussion_r876986821). Just wondering if we are consistently using the same patterns?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestTagApi.java
```diff
@@ -102,7 +102,7 @@ public void testApiTagOnInvoice() throws Exception {
         // Create a new tag definition
         //
         busHandler.pushExpectedEvents(NextEvent.TAG_DEFINITION);
-        final TagDefinition tagDefinition = tagUserApi.createTagDefinition("foo", "foo desc", ImmutableSet.<ObjectType>of(ObjectType.ACCOUNT, ObjectType.INVOICE), callContext);
+        final TagDefinition tagDefinition = tagUserApi.createTagDefinition("foo", "foo desc", new TreeSet<>(Set.of(ObjectType.ACCOUNT, ObjectType.INVOICE)), callContext);
```
pierre, line 105: You might want to add a comment that the `TreeSet` is needed for the ordering.

pierre (approved):

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestHardenCatalogPlugin.java
xsalefter: @pierre on that lines:
- standaloneCatalog.getSupportedCurrencies() return an array
- standaloneCatalog.getProducts() return a `Collection`

In `of` vs `copyOf` : because not like Guava's `ImmutableList.copyOf` that have wide range of target parameter (array, `Collection`, `Iterator`, etc), `List.copyOf` only applied to `Collection` (not even `Iterable`). If we want to apply to array, we should use `List.of`. None of List methods accept `Iterables`, so `Iterables.toUnmodifiableList()` (or `toList()) comes as helper.

That's why we have different `List` methods applied to parameter values.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestTagApi.java
xsalefter: Done

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestHardenCatalogPlugin.java
pierre: Got it, thanks!

sbrossie (approved):

PR #1706, opened 2022-05-31
overdue: Avoid computing billing state when account is tagged with OV…

sbrossie:
…ERDUE_ENFORCEMENT_OFF

The logic to check the OVERDUE_ENFORCEMENT_OFF is now moved earlier to avoid first computing the state (which can be costly for large accounts).

### DIFF
--- a/overdue/src/main/java/org/killbill/billing/overdue/wrapper/OverdueWrapper.java
+++ b/overdue/src/main/java/org/killbill/billing/overdue/wrapper/OverdueWrapper.java
@@ -100,6 +100,11 @@ public class OverdueWrapper {
     }
 
     private OverdueState refreshWithLock(final DateTime effectiveDate, final InternalCallContext context) throws OverdueException, OverdueApiException {
+        if (overdueStateApplicator.isAccountTaggedWith_OVERDUE_ENFORCEMENT_OFF(context)) {
+            log.debug("OverdueStateApplicator: apply returns because account (recordId={}) is set with OVERDUE_ENFORCEMENT_OFF", context.getAccountRecordId());
+            return null;
+        }
+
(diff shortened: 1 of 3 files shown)


overdue/src/main/java/org/killbill/billing/overdue/wrapper/OverdueWrapper.java
```diff
@@ -100,6 +100,11 @@ public OverdueState refresh(final DateTime effectiveDate, final InternalCallCont
     }
 
     private OverdueState refreshWithLock(final DateTime effectiveDate, final InternalCallContext context) throws OverdueException, OverdueApiException {
+        if (overdueStateApplicator.isAccountTaggedWith_OVERDUE_ENFORCEMENT_OFF(context)) {
+            log.debug("OverdueStateApplicator: apply returns because account (recordId={}) is set with OVERDUE_ENFORCEMENT_OFF", context.getAccountRecordId());
+            return null;
```
pierre, line 105: Shouldn't a clear state be returned instead?
That being said, it doesn't look like the return value is used. You could also just change this to return `void`.

pierre (approved):

overdue/src/main/java/org/killbill/billing/overdue/wrapper/OverdueWrapper.java
sbrossie: Fixed in https://github.com/killbill/killbill/pull/1706/commits/4b7051cdf469c63a5fefcc7eaf19e0a632850fc4

pierre (approved):

pierre:
FYI You can ignore the `snapshot` build failing (expected on `master`).

PR #1710, opened 2022-06-04
1375 - DateTime changes for create

reshmabidikar:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -352,10 +351,12 @@ public Response createSubscriptionsWithAddOnsInternal(final List<BulkSubscriptio
 
             final LocalDate resolvedEntitlementDate = toLocalDate(entitlementDate);
             final LocalDate resolvedBillingDate = toLocalDate(billingDate);
+            final DateTime entitlementDateTime = resolvedEntitlementDate != null ? resolvedEntitlementDate.toDateTimeAtCurrentTime() : null; //TODO_1375 Revisit once JAX_RS changes are done
```
pierre, line 354: As discussed - fix it using `TimeAwareContext`.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
```diff
@@ -309,67 +309,59 @@ public Entitlement cancelEntitlementWithPolicy(final EntitlementActionPolicy ent
...
-        baseEntitlementWithAddOnsSpecifierList.add(baseEntitlementWithAddOnsSpecifier);
-        final EntitlementContext pluginContext = new DefaultEntitlementContext(OperationType.CANCEL_SUBSCRIPTION,
-                                                                               getAccountId(),
-                                                                               null,
-                                                                               baseEntitlementWithAddOnsSpecifierList,
-                                                                               null,
-                                                                               properties,
-                                                                               callContext);
-        
+        //TODO_1375 - Using callContext.getCreatedDate() instead of creating DefaultEntitlementContext and using pluginContext.getCreatedDate()
```
sbrossie, line 322: Why do you need this `TODO` though? Prior, we were using the `pluginContext` (which itself was relying on `callContext.getCreatedDate()`), and now because code is shared using `cancelEntitlementWithDateInternal`, the `pluginContext` is created inside the shared method, so removing it makes perfect sense, and the use of `callContext.getCreatedDate()` is correct. Should we just remove the `TODO` or if not what did you have in mind?

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
```diff
@@ -332,7 +338,7 @@ public UUID doCall(final EntitlementApi entitlementApi, final DefaultEntitlement
 
                     final DefaultBaseEntitlementWithAddOnsSpecifier baseEntitlementWithAddOnsSpecifier = getFirstBaseEntitlementWithAddOnsSpecifier(updatedPluginContext.getBaseEntitlementWithAddOnsSpecifiers());
 
-                    final DateTime requestedDate = dateHelper.fromLocalDateAndReferenceTime(baseEntitlementWithAddOnsSpecifier.getBillingEffectiveDate(), updatedPluginContext.getCreatedDate(), contextWithSourceAccountRecordId);
+                    final DateTime requestedDate = baseEntitlementWithAddOnsSpecifier.getBillingEffectiveDate();
```
sbrossie, line 341: @pierre Interesting, should we also add an api with `DateTime` for transfer?

sbrossie: @reshmabidikar Also, I noticed that we are not always super consistent with where we get the date from: Sometimes we use the date from the `pluginContext` (-> `baseEntitlementWithAddOnsSpecifier` -> `getBillingEffectiveDate`) and sometimes we use the date built on top of the `doCall` method (as a closure). It should be the same date though, assuming we don't mess up.

(Perhaps, this was already the case prior your change).

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
```diff
@@ -352,10 +351,12 @@ public Response createSubscriptionsWithAddOnsInternal(final List<BulkSubscriptio
 
             final LocalDate resolvedEntitlementDate = toLocalDate(entitlementDate);
```
sbrossie, line 352: My understanding is that both `entitlementDate` and `billingDate` could be either a `LocalDate` or a `DateTime`: So we should not anymore convert using `toLocalDate`:
* If this is a `DateTime` (i.e we can deserialize the input string into a DateTime), we take it as-is
* If not, then it should be a valid `LocalDate` and we do the gymnastics to convert it into a `DateTime`

entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApi.java
```diff
@@ -745,6 +745,154 @@ public void testCreateBaseWithBillingInTheFuture() throws AccountApiException, E
...
+
+        //Verify that state is now ACTIVE
+        entitlement = entitlementApi.getEntitlementForId(entitlementIds.get(0), callContext);
+        assertEquals(entitlement.getState(), EntitlementState.ACTIVE);
+
+        //Move clock to BillingDateTime and verify that subscription starts
+        testListener.pushExpectedEvents(NextEvent.CREATE);
+        clock.setTime(billingDatetime);
+        assertListenerStatus();
+    }
```
sbrossie, line 895: 👏  for all these good tests!

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestMigrationSubscriptions.java
```diff
@@ -237,8 +237,8 @@ public void testSimpleMigrationBundle() throws Exception {
         accountChecker.checkAccount(account.getId(), accountData, callContext);
 
         // We set both entitlement and billing date with desired value
-        final LocalDate entitlementMigrationDate = new LocalDate(2015, 12, 20);
-        final LocalDate billingMigrationDate = new LocalDate(2016, 2, 1);
+        final DateTime entitlementMigrationDate = new DateTime(2015, 12, 20, 0, 0, 0);
```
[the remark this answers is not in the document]
sbrossie: I would argue, that it works, but it's more about what we want the test to do: Right now everything is aligned at `00:00:00`, (e.g the `account#reference_time` is and then we also force the subscriptions dates to be also aligned to `00:00:00`.  It seems fine with me, but as always we could try bother combinations if we wanted, but @reshmabidikar Already write some good tests to verify the dates being passed. @pierre Did you have something else in mind?

sbrossie (commented):
Looks great, there are few things to all agree on and review.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestMigrationSubscriptions.java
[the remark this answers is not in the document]
pierre: My point was that the code implicitly assumes the reference time is midnight. I was thinking constructing the `DateTime` time portion using the reference time, or at least adding a comment.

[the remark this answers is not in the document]
reshmabidikar: I've modified this now to use:

```
final DateTime entitlementMigrationDateBundle1 = internalCallContext.toUTCDateTime(new LocalDate(2015, 12, 20));
```

LMK if this makes sense. If so, there might be some other tests that need to be updated as above.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
reshmabidikar: Fixed

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
reshmabidikar: Agreed, I guess I got a bit confused, but your explanation makes sense, I've deleted the TODO

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
reshmabidikar: > @pierre Interesting, should we also add an api with `DateTime` for transfer?

LMK if this needs to be added now or later on (we had discussed this in the [original google doc](https://docs.google.com/document/d/1pE3AaQHRnu14nU2_SFnPb-RDzFxj3eTi5x4zdhHsydw/edit#heading=h.ftyog36esr8v) and decided to skip for now.

reshmabidikar: > @reshmabidikar Also, I noticed that we are not always super consistent with where we get the date from: Sometimes we use the date from the `pluginContext` (-> `baseEntitlementWithAddOnsSpecifier` -> `getBillingEffectiveDate`) and sometimes we use the date built on top of the `doCall` method (as a closure). It should be the same date though, assuming we don't mess up.
> 
> (Perhaps, this was already the case prior your change).

Yes, I guess wherever possible, the code uses the dates passed to it (after the conversion to datetime if applicable) within the `doCall` method and in other cases uses the date from the `pluginContext`.  Do you want me to make changes so that the behavior is consistent throughout? If so, I think using the date from  `pluginContext` makes more sense. What do you think?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResource.java
reshmabidikar: Yes agreed, I will be making these JAXRS layer changes for all methods ( `create`, `change`, `cancel` ..) in the next round.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestMigrationSubscriptions.java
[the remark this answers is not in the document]
pierre: Sounds good -- we can always fix the other tests next time we update them.

Just FYI, your fix works because of this: https://github.com/killbill/killbill/blob/793b841c0f5b7642df680d1d3bb50b9744942cb5/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestIntegrationBase.java#L513

pierre (approved):

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlementApi.java
sbrossie: I think using the dates from the pluginContext is indeed better - as we ensure that both KB core logic and plugin see the same dates. We can fix it in another PR - not super urgent, but it would help with maintainability moving forward.

sbrossie (approved):

PR #1712, opened 2022-06-07
1615 jaxrs part 2

xsalefter:
Some notes:

- [JaxRsResourceBase.deleteCustomFields()](https://github.com/killbill/killbill/blob/cef3351c4ec44347d478f61e420ae8d443fc62f8/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java#L300), add small logic `customFieldList == null` to match specs in javadoc, " .... or null if they should all be removed"

- Move `TestJaxRsResourceBase.testGetHighestRecordDate()` to `TestUsageResource.testGetHighestRecordDate()`

- extract `BundleResource.getBundles()` and `BundleResource.searchBundles()` methods to `buildBundleStreamPaginationResponse()` because they are contains (almost) the same method body.

### DIFF
--- a/jaxrs/src/test/java/org/killbill/billing/jaxrs/resources/TestJaxRsResourceBase.java
+++ b/jaxrs/src/test/java/org/killbill/billing/jaxrs/resources/TestJaxRsResourceBase.java
@@ -28,7 +28,6 @@ import org.killbill.billing.payment.api.PluginProperty;
 import org.killbill.billing.util.UUIDs;
 import org.killbill.billing.util.api.CustomFieldApiException;
 import org.killbill.billing.util.api.CustomFieldUserApi;
-import org.killbill.billing.util.callcontext.CallContext;
 import org.killbill.billing.util.collect.Iterables;
 import org.killbill.billing.util.customfield.CustomField;
 import org.mockito.Mockito;
@@ -40,7 +39,7 @@ public class TestJaxRsResourceBase extends JaxrsTestSuiteNoDB {
 
     private CustomFieldUserApi customFieldUserApi;
 
-    @BeforeMethod
+    @BeforeMethod(groups = "fast", alwaysRun = true)
     public void beforeMethod() {
         customFieldUserApi = Mockito.mock(CustomFieldUserApi.class);
     }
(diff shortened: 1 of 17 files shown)


xsalefter:
@sbrossie @pierre failed because problem described in image below. Should we ignore this? Or perhaps change regex in `DefaultTagUserApi.createTagDefinition()` to something like this:

```java
    public static boolean containsUpperCase(final String s) {
        if (isNullOrEmpty(s)) {
            return false; // same as regex ".*[A-Z].*"
        }
        return !s.equals(s.toLowerCase());
    }
```


Screenshot from 2022-06-08 07-53-51

sbrossie:
> @sbrossie @pierre failed because problem described in image below. Should we ignore this? Or perhaps change regex in `DefaultTagUserApi.createTagDefinition()` to something like this:
> 
> ```java
>     public static boolean containsUpperCase(final String s) {
>         if (isNullOrEmpty(s)) {
>             return false; // same as regex ".*[A-Z].*"
>         }
>         return !s.equals(s.toLowerCase());
>     }
> ```
> 
> Screenshot from 2022-06-08 07-53-51

@xsalefter If the concern is that the input string may be too long and run into `n^k`, we could first check that the string is not too long - it should not be longer than what our [schema](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/util/src/main/resources/org/killbill/billing/util/ddl.sql#L53) allows, and then keep the regex as-is, as there would be no more concern.

jaxrs/src/test/java/org/killbill/billing/jaxrs/resources/TestJaxRsResourceBase.java
```diff
@@ -17,31 +17,58 @@
...
 import org.testng.annotations.Test;
 
-import com.google.common.collect.ImmutableList;
-
 public class TestJaxRsResourceBase extends JaxrsTestSuiteNoDB {
 
-    private final JaxRsResourceBaseTest base = new JaxRsResourceBaseTest();
+    private CustomFieldUserApi customFieldUserApi;
+
+    @BeforeMethod(groups = "fast", alwaysRun = true)
```
pierre, line 42: No need for `alwaysRun = true`.

jaxrs/src/test/java/org/killbill/billing/jaxrs/resources/TestJaxRsResourceBase.java
```diff
@@ -17,31 +17,58 @@
...
+    public void beforeMethod() {
+        customFieldUserApi = Mockito.mock(CustomFieldUserApi.class);
```
pierre, line 44: Is there a `super` method to call?

Regardless, I think you also need:
```
        if (hasFailed()) {
            return;
        }
```
to work around issues with TestNG where it runs these methods even if the suite is failing.

pierre (approved):
Just a couple of comments, but otherwise :shipit:

pierre:
> @sbrossie @pierre failed because problem described in image below. Should we ignore this? Or perhaps change regex in `DefaultTagUserApi.createTagDefinition()` to something like this:
> 
> ```java
>     public static boolean containsUpperCase(final String s) {
>         if (isNullOrEmpty(s)) {
>             return false; // same as regex ".*[A-Z].*"
>         }
>         return !s.equals(s.toLowerCase());
>     }
> ```

Agreed to this ☝️ Let's just change the implementation.

jaxrs/src/test/java/org/killbill/billing/jaxrs/resources/TestJaxRsResourceBase.java
xsalefter: add `if (hasFailed()) { return; }` works.

xsalefter:
@sbrossie I took @pierre 's [approach](https://github.com/killbill/killbill/pull/1712#issuecomment-1149506140) instead of yours :sweat_smile: .

sbrossie:
> @sbrossie I took @pierre 's [approach](https://github.com/killbill/killbill/pull/1712#issuecomment-1149506140) instead of yours 😅 .

All good!

PR #1713, opened 2022-06-07
invoice: Add accountRId 'where' clause for getRepairMap query

sbrossie:


invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceItemSqlDao.sql.stg
```diff
@@ -139,6 +139,7 @@ getRepairMap(ids) ::= <<
   JOIN invoices i on i.id = ii2.invoice_id
   WHERE
   ii1.invoice_id in (<ids>)
+  and <accountRecordIdField("i.")> = :accountRecordId
```
pierre, line 142: Should we add the filter on all `ii*` tables (like we do for tenant below)?

pierre (approved):

PR #1714, opened 2022-06-09
invoice: fix regression from a47ba879138d7b2b966c4e8afb192c6a5bb14e6c

pierre:
The account balance wasn't always accurate when some invoices were written off.


sbrossie (approved):
👍

PR #1720, opened 2022-06-13
subscription: Remove unused date restriction on bundle transfer

sbrossie:
I removed the restriction because it prevents doing transfer bundle with a future date - and I am not sure why. This is old code from 2013, perhaps we had a good reason at the time?


pierre (approved):
Found the original commit: https://github.com/killbill/killbill/commit/752be1d398ebcf42223692ebd0561642e2955289

> The entitlement code always assumes that the first event for a
subscription is in the past

Apparently, in Kill Bill 0.1, we had a requirement to have start dates in the past. Things have changed a bit... 😄

PR #1728, opened 2022-06-21
1615 blocking state dao

xsalefter:
`BlockingStateDao.setBlockingStatesAndPostBlockingTransitionEvent()` use Guava's Optional instead of Java 8 Optional. This commit changes that, also fix affected classes. As result, Guava dependency in entitlement's `pom.xml` get removed.


pierre (approved):
🥳

sbrossie (approved):

Issue #1631, opened 2022-03-17
Plan change with subscription start date results in an error


reshmabidikar:
When  a plan is changed with a date equal to the subscription start date, the following error occurs:

```
Error while changing subscription: Requested date 2022-03-14T12:04:58.000Z is not allowed to be prior to the previous transition 2022-03-14T12:05:11.000Z
```
Steps to Reproduce:

1. Set date to `2022-03-14`.
2. Upload attached catalog.
3. Create a subscription corresponding to the `premium-monthly` plan.
4. Move clock by 10 days.
5. Change plan to `standard-monthly`  with date as `2022-03-14` - This results in the above error.

**Expected Behavior: The plan change should take effect immediately.**

[monthly-no-trial.zip](https://github.com/killbill/killbill/files/8281694/monthly-no-trial.zip)

sbrossie:
Verified as fixed in 0.22.30

PR #1734, opened 2022-06-28
Test and potential fix for 1631

reshmabidikar:
- Includes code changes and potential fix for #1631 
- Code changes caused `TestUserApiChangePlan.testChangePlanOnPendingSubscription` to fail, so had to change it [here](https://github.com/reshmabidikar/killbill/blob/7ec2a02332667f83bee1005e6deee076896f68a9/subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiChangePlan.java#L534).
- Code changes caused `TestDefaultEntitlementApi.testCreateBaseWithDifferentStartDate` to fail so had to change it [here](https://github.com/reshmabidikar/killbill/blob/7ec2a02332667f83bee1005e6deee076896f68a9/entitlement/src/test/java/org/killbill/billing/entitlement/api/TestDefaultEntitlementApi.java#L679).

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java
@@ -575,4 +575,42 @@ public class TestSubscription extends TestIntegrationBase {
 
     }
 
+    @Test(groups = "slow",description="https://github.com/killbill/killbill/issues/1631")
+    public void testChangePlanWithStartDate() throws Exception {
+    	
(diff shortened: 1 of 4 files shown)


sbrossie (approved):
The fix makes sense (I suggested it), but what's not great is that we keep having this kind if logic throughout the subscription code. Ideally, we should revisit for all these calls where such logic lives and try to factor in such code:
* Code that checks for date validity
* Code that re-aligns input date to correct default
* Code that truncate millisec
* ... and so on

PR #1736, opened 2022-06-29
remove Guava's cache and replace with LinkedHashMap

xsalefter:


util/src/main/java/org/killbill/billing/util/security/shiro/realm/KillBillAuth0Realm.java
```diff
@@ -161,10 +144,11 @@ public PublicKey call() throws Exception {
                                  }
                              })
                              .build();
-        this.keys = CacheBuilder.newBuilder()
-                                .maximumSize(CACHE_MAXIMUM_SIZE)
-                                .expireAfterWrite(CACHE_TIMEOUT_MINUTES, TimeUnit.MINUTES)
-                                .build();
+        this.keys = new LinkedHashMap<>(CACHE_MAXIMUM_SIZE, .75f, true) {
+            public boolean removeEldestEntry(final Map.Entry eldest) {
+                return size() > CACHE_MAXIMUM_SIZE;
```
sbrossie, line 149: ⚠️ Caching things for 15 minutes (what we had) is a very different thing than caching 15 different entries.

Ideally, we should have 2 settings:

- `CACHE_MAXIMUM_SIZE` -> Make sure this is bounded
- `CACHE_TIMEOUT_MINUTES` -> Don't remove elements unless they have been there for `CACHE_TIMEOUT_MINUTES` minutes *unless* we hit  `CACHE_MAXIMUM_SIZE` 

Unfortunately, the `Map.Entry` does not provide the timestamp we need, so we have 2 choices:

1. We could insert objects that provide this timestamp (but adds complexity)
2. We could simply have a `CACHE_MAXIMUM_SIZE` that is large enough to make sure that with an expected deployment entries stay long enough in the cache. If we assume 1000 /min => 10000/10 min. 

I would suggest to go with 2/ for now, and use a value of 10000.

@pierre Any thoughts?

sbrossie (changes requested):

util/src/main/java/org/killbill/billing/util/security/shiro/realm/KillBillAuth0Realm.java
pierre: FWIW I had just copied the default timeout from upstream (https://github.com/auth0/jwks-rsa-java#guavacachedjwkprovider).

The issue with this new implementation is that the keys will never be updated anymore in case they change (which was the point before).

It sounds like we would benefit from having a time-based cache in `killbill-utils` (we have 2 use-cases already...). We could be clever with `System. currentTimeMillis ()` and do some logic in `removeEldestEntry`?

sbrossie: Yes I agree, we could have our own implementation of a time-based cache in `killbill-utils`.

The impl. could reply on `LinkedHashMap` but the perhaps we can have our own subclass of `Map.Entry` that takes on a timestamp and that we can use to decide whether to evict or not.

@xsalefter Does this make sense?

xsalefter: @sbrossie Yeah I'll try to implements it in `killbill-utils`.

xsalefter:
Closed and replaced by #1740

PR #1744, opened 2022-07-22
Invoice translation custom properties

Jimmy-Newtron:
This code wants to fix an issue encountered while working to create an invoice template and the relative translations.

In my template I added something like 

```
<div><strong>{{text.payTo}}</strong></div>
```

Even if I added the ```payTo``` property in the translation file for both the default locale (en_US) and the account locale (fr_FR) I get the error from mustache

```
"stack_trace":"org.glassfish.jersey.server.internal.process.MappableException: com.samskivert.mustache.MustacheException$Context: No method or field with name 'text.payTo' on line 46
...
```

The code wants to fix this allowing to use a new field to access any property name even when there is no a named method

In the template we will be able to use the following:

```
<div><strong>{{text.properties.payTo}}</strong></div>
```

Properties will be a Map and should return the translated value for any custom key

Let me know if you see any problem to add this fix
Cheers,
Enrico

### DIFF
--- a/invoice/src/test/java/org/killbill/billing/invoice/TestHtmlInvoiceGenerator.java
+++ b/invoice/src/test/java/org/killbill/billing/invoice/TestHtmlInvoiceGenerator.java
@@ -66,6 +66,7 @@ public class TestHtmlInvoiceGenerator extends InvoiceTestSuiteNoDB {
         final HtmlInvoice output = g.generateInvoice(createAccount(), createInvoice(), false, internalCallContext);
         Assert.assertNotNull(output);
         Assert.assertNotNull(output.getBody());
+        Assert.assertTrue(output.getBody().contains("<!-- Test customKey value -->"));
         Assert.assertEquals(output.getSubject(), "Your invoice");
     }
 
--- a/invoice/src/test/resources/org/killbill/billing/util/template/translation/InvoiceTranslation_en_US.properties
+++ b/invoice/src/test/resources/org/killbill/billing/util/template/translation/InvoiceTranslation_en_US.properties
@@ -38,3 +38,5 @@ invoiceItemAmount=Amount
 
 processedPaymentCurrency=(*) The payment was processed in
 processedPaymentRate=The rate applied was
+
+customKey=Test customKey value
--- a/util/src/main/java/org/killbill/billing/util/template/translation/DefaultTranslatorBase.java
+++ b/util/src/main/java/org/killbill/billing/util/template/translation/DefaultTranslatorBase.java
@@ -16,8 +16,6 @@
 
 package org.killbill.billing.util.template.translation;
 
-import java.util.HashMap;
-import java.util.Map;
 import java.util.ResourceBundle;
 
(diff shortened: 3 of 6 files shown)


Jimmy-Newtron:
The problem still persist and I spend my afternoon trying to find a solution but I could not

sbrossie:
@Jimmy-Newtron Thanks for the contribution.

It looks like CI tests are broken, could you fix it?
Screen Shot 2022-07-22 at 4 35 19 PM

Also, could you add specific unit tests for the new feature?

reshmabidikar:
> The problem still persist and I spend my afternoon trying to find a solution but I could not

Thanks for submitting this PR, I'm a part of the Kill Bill team. From what I understand, you need to add a custom property to your invoice. Is that correct? 

If so, I think you will need to do the following:
1. Add the property to your [translation file](https://github.com/killbill/killbill/blob/6cef988a976cd357dae9f52c3c7b0cec9183fe46/util/src/main/resources/org/killbill/billing/util/invoice/translation/InvoiceTranslation_en_US.properties) and [invoice template](https://github.com/killbill/killbill/blob/6cef988a976cd357dae9f52c3c7b0cec9183fe46/util/src/main/resources/org/killbill/billing/util/invoice/templates/HtmlInvoiceTemplate.mustache)
2. Add a getter method corresponding to the property in the [DefaultInvoiceTranslator](https://github.com/killbill/killbill/blob/6cef988a976cd357dae9f52c3c7b0cec9183fe46/invoice/src/main/java/org/killbill/billing/invoice/template/translator/DefaultInvoiceTranslator.java)

Could you give this a try and let us know how it goes?

Jimmy-Newtron:
Thanks for your replies. As I commented before

> The problem still persist and I spend my afternoon trying to find a solution but I could not

@sbrossie I have implemented working unit test for the code and I moved to mustache unit test realizing that it does not search maps values, but only Object methods and fields

@reshmabidikar my goal was to add custom properties and not modify each time the code to add a new getter method

I tried to implement a Lamda function like in the [Mustache guide](https://mustache.github.io/mustache.5.html) to get any key without changing the code.... but I failed miserably

```
public Function<String, String> properties() {
    return input -> "this::getTranslation";
}
```

I think I will leave it aside for the moment, but I hope to find a solution .... because I got extremely frustrated to not solve this issue

Jimmy-Newtron:
@sbrossie and @reshmabidikar  I finally get over the issue and pushed a generic Lambda that allow to retrieve any customKey without adding new methods

Cheers

sbrossie (approved):

Jimmy-Newtron:
Any news about merging it?

Jimmy-Newtron:
@reshmabidikar I have finalized my work can you go further with the next steps?

reshmabidikar:
Hi @Jimmy-Newtron, the team was very busy last week but I'll let you know as soon as I find out more on this. Thanks for your patience!

sbrossie:
@Jimmy-Newtron Code merged and available in `0.22.30`.

PR #1747, opened 2022-07-28
Support phaseType change in invoices dry runs

dawidworobiec:


dawidworobiec:
Sorry, will create a topic on the mailing list before creating PR once again.

PR #1752, opened 2022-08-03
1030 - Changes for Step 1.1 (Added includeDeletedEvents flag to all methods)

reshmabidikar:
Changes for Step 1.1

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
@@ -104,9 +104,9 @@ public class DefaultSubscriptionBase extends EntityBase implements SubscriptionB
     // the call completes
     //
     private LinkedList<SubscriptionBaseTransition> transitions;
-    
+
     private LinkedList<SubscriptionBaseTransition> transitionsWithDeletedEvents;
-    
+
     private boolean includeDeletedEvents;
 
 	// Low level events are ONLY used for Repair APIs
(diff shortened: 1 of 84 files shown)


sbrossie:
@reshmabidikar I merged PR https://github.com/killbill/killbill/pull/1750, could you update your branch with latest work-for-release-0.23.x - hopefully this should help with the transient test failures.

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -812,6 +836,23 @@ public void rebuildTransitions(final List<SubscriptionBaseEvent> inputEvents, fi
 
         removeEverythingPastCancelEvent(events);
 
+        transitions = new LinkedList<SubscriptionBaseTransition>();
+        transitionsWithDeletedEvents = new LinkedList<SubscriptionBaseTransition>();
+
+        final List<SubscriptionBaseEvent> activeEvents = inputEvents.stream().filter(event -> event.isActive()).collect(Collectors.toList());
+        rebuildTransitionsInternal(activeEvents, catalog, true);
```
sbrossie, line 843: I think calling `rebuildTransitionsInternal` twice because we have to build 2 list, i.e. `transitions` and `transitionsWithDeletedEvents` makes sense, but I am not convinced
doing a first pass with just the active events and then if `includeDeletedEvents` is true doing a second pass to complete the list of `transitionsWithDeletedEvents` leads to the correct result as 'transitions' span events and they need to be ordered. Also, there is a lot of state between what `inputEvents` contains based on `includeDeletedEvents=true|false`, the state member variable `includeDeletedEvents` and the new boolean `activeEvents`.


Instead, could we:
1. If `includeDeletedEvents=false` (most common case), just call once `rebuildTransitionsInternal`, no filtering even required
2. If `includeDeletedEvents=true`, call `rebuildTransitionsInternal` once with filtering to build 'transitions'  and then call `rebuildTransitionsInternal` without filtering to build `transitionsWithDeletedEvents`.

Perhaps, we could make `rebuildTransitionsInternal` a static method and pass the correct List<SubscriptionBaseTransition> as an argument, i.e either `transitions` or `transitionsWithDeletedEvents` based on cases; it looks like you would need to pass `id`, `bundleId`, `bundleExternalKey` in order to do that. The benefit is that this method
would just build a list of transitons based on input and there is no additional boolean `activeEvents` nor use of  the state member variable `includeDeletedEvents` in this method anymore making it much simpler.

sbrossie:
@xsalefter I think the CI error from this branch is expected, but can you confirm:
Screen Shot 2022-08-15 at 12 07 48 PM

xsalefter:
@sbrossie yes, it is expected. If you or @reshmabidikar want to try to pass build locally without skipping dependency plugin, you can try what I did [here](https://github.com/killbill/killbill/pull/1754/files). But then this also not problem-free: You'll see Guice binding problem in beatrix `slow` tests.

sbrossie (approved):
Looks good, I think we should merge.

PR #1754, opened 2022-08-13
update various version

xsalefter:
update various version, and remove declared killbill-platform dependencies

- killbill parent to 0.145.3-8ec8d1f-SNAPSHOT
- killbill-commons to 0.25.1-2361a7d-SNAPSHOT
- killbill-platform to 0.41.0-642a42b-SNAPSHOT

Draft, since I'm wondering whether removing various killbill-platform artifact affecting the `slow` test or not.

Update: Build failed at beatrix because of Guice.TDLR: We need to find out exact replacement of `@com.google.guice.Inject(optional = true)`, because `@javax.inject.Inject` + `@Nullable` is not direct replacement and behave differently.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.kill-bill.billing</groupId>
         <artifactId>killbill-oss-parent</artifactId>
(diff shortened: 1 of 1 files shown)


xsalefter:
Replaced by PR #1760

PR #1760, opened 2022-08-24
Fix guice dependency problem

xsalefter:


profiles/killbill/pom.xml
```diff
@@ -61,15 +61,25 @@
...
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>jsr305</artifactId>
-            <scope>provided</scope>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>com.google.inject</groupId>
             <artifactId>guice</artifactId>
-            <scope>provided</scope>
+            <scope>compile</scope>
```
sbrossie, line 82: That seems reasonable - who was 'providing' such package before this?

sbrossie (approved):
I tried launching KB with the change using the `jetty:run` and it seems to be working fine on my end.

profiles/killbill/pom.xml
pierre: FYI `compile` is the default scope (and it isn't overridden in the base pom.xml: https://github.com/killbill/killbill-oss-parent/blob/17b5048cafb6fe6c80a58ce12c6163844dea214c/pom.xml#L361-L365), so you shouldn't even need a scope at all.

pom.xml
```diff
@@ -59,11 +59,12 @@
         <url>http://github.com/killbill/killbill</url>
     </issueManagement>
     <properties>
+        <check.skip-dependency-versions>true</check.skip-dependency-versions>
```
pierre, line 62: Let's not forget to fix this eventually...

pom.xml
```diff
@@ -106,43 +107,6 @@
                     </exclusion>
                 </exclusions>
             </dependency>
-            <!-- Temporary until 0.24.x is released -->
```
pierre, line 109: Wondering if I had added this specifically for that `check.skip-dependency-versions` 🤔 

I think even the `two-digits-backward-compatible` strategy isn't good enough for our `SNAPSHOT` scheme: https://github.com/killbill/killbill-oss-parent/blob/17b5048cafb6fe6c80a58ce12c6163844dea214c/pom.xml#L1712

The real fix would be:

1. Switch to https://github.com/basepom/dependency-versions-check-maven-plugin from the old Ning plugin (see https://github.com/killbill/killbill-oss-parent/pull/445 and https://github.com/killbill/killbill-oss-parent/pull/431#issuecomment-1112488332 for why I couldn't do it)
2. Submit a PR upstream to @hgschmie to support our naming scheme (`ignore-SNAPSHOT-with-sha` strategy)

Maybe take a look at this once the build is stable again?

profiles/killbill/src/main/java/org/killbill/billing/server/modules/KillbillServerModule.java
```diff
@@ -106,6 +116,15 @@ protected void configure() {
         configurePushNotification();
 
         bind(new TypeLiteral<OSGIServiceRegistration<Healthcheck>>() {}).to(DefaultHealthcheckPluginRegistry.class).asEagerSingleton();
+
+        // Needed because changes in KillbillHealthcheck from Guice @Inject(optional=true) to javax @Inject + @Nullable
```
pierre, line 120: Subtle! 😅 

We should do some additional manual testing to verify all of this plumbing still works as expected. For instance, service discovery can be tested using the Eureka bundle (https://github.com/killbill/killbill-platform/tree/work-for-release-0.23.x/osgi-bundles/bundles/eureka). @vnandwana can help if needed (he had to do similar testing when doing the metrics work).

pierre (approved):

profiles/killbill/pom.xml
xsalefter: > That seems reasonable - who was 'providing' such package before this?

[I did](https://github.com/killbill/killbill/commit/429ac9b1166f724cbd18a953279e6400be3fa91d#diff-030ec92e9612718b2d2bc8aa5b7a65ebd296c063e121df61c273b3b00530b926) :man_facepalming: . I think in attempt to fix something else.

profiles/killbill/src/main/java/org/killbill/billing/server/modules/KillbillServerModule.java
xsalefter: Noted.

pom.xml
xsalefter: Tracked [here](https://github.com/killbill/killbill/issues/1763). Meanwhile, I'll mark this as resolved.

profiles/killbill/pom.xml
xsalefter: > so you shouldn't even need a scope at all.

Fixed by this [commit](https://github.com/killbill/killbill/pull/1760/commits/aaedb5e4f3755ad87e74df93f946518e186d85e0).

profiles/killbill/src/main/java/org/killbill/billing/server/modules/KillbillServerModule.java
xsalefter: @pierre tracked [here](https://github.com/killbill/killbill/issues/1764), although I'm not sure what do you mean by ".... service discovery can be tested using the Eureka bundle".

sbrossie:
Merging as this solves the issue with the build 🎉

PR #1762, opened 2022-08-25
Work for release 0.23x 1030 blocking states

reshmabidikar:
Changes for blocking states


entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
```diff
@@ -88,11 +91,10 @@ public DefaultEventsStream(final ImmutableAccountData account,
                                final SubscriptionBase subscription,
                                final Collection<SubscriptionBase> allSubscriptionsForBundle,
                                @Nullable final Integer defaultBillCycleDayLocal,
-                               final InternalTenantContext contextWithValidAccountRecordId, final DateTime utcNow) {
+                               final InternalTenantContext contextWithValidAccountRecordId, final DateTime utcNow, final boolean includeDeletedEvents) {
```
sbrossie, line 94: I am not super excited about passing both the `blockingStates` (unclear what those are unless you look at how it's been called) and the `includeDeletedEvents` flag.

Ideally, we would want the impl. to be as symmetrical as what we have done in `DefaultSubscriptionBase`, but I understand it's not easy as 'events' get computed inside `DefaultSubscriptionBase` and here they are computed outside (so I understand why you did that).

Perhaps:
1. We pass both `blockingStates` and `blockingStatesWithDeletedEvents` 
2. We compute the value of `includeDeletedEvents ` based on the lists we receive - see comment below.
3. We have to define the semantics of what blockingStatesWithDeletedEvents contains when caller was called with `includeDeletedEvents=false` (In `DefaultSubscriptionBase`, I think this is indeed an empty list), hence my comment below for the `getBlockingStates()` method.

entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
```diff
@@ -239,7 +251,7 @@ public Integer getDefaultBillCycleDayLocal() {
 
     @Override
     public Collection<BlockingState> getBlockingStates() {
```
sbrossie, line 253: Can we make the code symmetrical with `DefaultSubscriptionBase#getAllTransitions(final boolean includeDeleted)`, i.e the method takes the  `includeDeleted` flag, which is only called from the `BlockingStateOrdering#getBlockingStates()` ?

sbrossie (changes requested):
Also, check out the `TODO_1030`, you have in the code. For instance, [here](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java#L275), the value should come from the `DefaultEventsStream#includeDeletedEvents` and not be hardcoded (this may explain why some of your test failed).

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
```diff
@@ -101,13 +101,14 @@ public class DefaultEntitlement extends EntityBase implements Entitlement {
     // Refresh-able
     protected EventsStream eventsStream;
 
+    //TODO_1030 - This constructor is not used anywhere in the code, can it be deleted?
```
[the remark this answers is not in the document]
reshmabidikar: Done

api/src/main/java/org/killbill/billing/junction/DefaultBlockingState.java
```diff
@@ -169,6 +203,8 @@ public int compareTo(final BlockingState arg0) {
             return comparison;
         }
     }
+    
+    //TODO_1030: Should isActive flag be used in equals and hashCode method?
```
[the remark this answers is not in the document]
reshmabidikar: Fixed

entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
reshmabidikar: I've now made changes as per my understanding of the above, please let me know if it still does not look good.

entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
reshmabidikar: Done

entitlement/src/main/java/org/killbill/billing/entitlement/dao/DefaultBlockingStateDao.java
```diff
@@ -189,16 +189,27 @@ public List<BlockingState> getBlockingActiveForAccount(final VersionedCatalog ca
...
-        return transactionalSqlDao.execute(true, entitySqlDaoWrapperFactory -> {
-            final List<BlockingStateModelDao> states = entitySqlDaoWrapperFactory
-                    .become(BlockingStateSqlDao.class)
-                    .getByBlockingIds(blockableIds, context);
-            return states.stream()
-                    .map(BlockingStateModelDao::toBlockingState)
-                    .collect(Collectors.toList());
-        });
+    public List<BlockingState> getByBlockingIds(final Iterable<UUID> blockableIds, final boolean includeDeletedEvents, final InternalTenantContext context) {
+        if (includeDeletedEvents) {
```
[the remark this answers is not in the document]
reshmabidikar: Fixed

reshmabidikar:
Hi @sbrossie, I have now fixed your review comments. The following commits contain the changes for this: 

1. https://github.com/killbill/killbill/pull/1762/commits/24656b4d9243b0c55145a47553a7881e59a706dc
2. https://github.com/killbill/killbill/pull/1762/commits/369fee560b59fd45f06bb9394462659ec14ee685

Please take a look and let me know what you think. 

With regards to the following: 

> Also, check out the `TODO_1030`, you have in the code. For instance, [here](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java#L275), the value should come from the `DefaultEventsStream#includeDeletedEvents` and not be hardcoded (this may explain why some of your test failed).

I made this change but found that the addon test still failed. I had to make some other modifications for this as mentioned below. 

**Addon test failure issue**

I was able to get this to work and created a separate commit [here](https://github.com/killbill/killbill/pull/1762/commits/31f26ddc7522f440b7f4c19327f6406ec7424af5). A few details about this:
1. [This ](https://github.com/reshmabidikar/killbill/blob/31f26ddc7522f440b7f4c19327f6406ec7424af5/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java#L1014)is the test that failed.
2. I got it to work by making [this change](https://github.com/reshmabidikar/killbill/blob/31f26ddc7522f440b7f4c19327f6406ec7424af5/entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java#L519).
4. However,  I think modifying the `subscriptionEntitlementStates` seems a bit risky (similar to the issue we discussed for modifying `DefaultSubscriptionBase#transitions`). The alternative would be to keep a separate list like `subscriptionEntitlementStatesWithDeletedEvents` and use that to compute `entitlementCancelEvent` and others. Please let me know what you think.

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
```diff
@@ -874,17 +875,20 @@ private List<DefaultSubscriptionBase> buildBundleSubscriptions(final List<Defaul
...
-
+                        
+                        ApiEventBuilder builder = new ApiEventBuilder()
+                                .setSubscriptionId(reloaded.getId())
+                                .setEffectiveDate(baseTriggerEventForAddOnCancellation.getEffectiveDate())
+                                .setCreatedDate(baseTriggerEventForAddOnCancellation.getCreatedDate())
+                                // This event is only there to indicate the ADD_ON is future canceled, but it is not there
+                                // on disk until the base plan cancellation becomes effective
+                                .setFromDisk(false);
+                        builder = cur.getIncludeDeletedEvents() ? builder.setActive(baseTriggerEventForAddOnCancellation.isActive()) : builder;
```
sbrossie, line 886: I am wondering if we really need this logic: We are creating an in memory cancel event for the add-on as a result of seeing a future cancelation for the base plan, so I would think such computed event always need to be active. Was there a reason (test failing?) why you added this?

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
```diff
@@ -837,7 +838,7 @@ private List<DefaultSubscriptionBase> buildBundleSubscriptions(final List<Defaul
             switch (cur.getCategory()) {
                 case BASE:
                     for (final SubscriptionBaseEvent event : events) {
-                        if (!event.isActive()) {
+                    	if (!cur.getIncludeDeletedEvents() && !event.isActive()) { //TODO_1030 - revisit to see if this is correct
```
sbrossie, line 841: I don't think this is correct, see next comment below:

entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
```diff
@@ -516,7 +516,8 @@ private void computeEntitlementBlockingStates() {
 
     private List<BlockingState> filterBlockingStatesForEntitlementService(final BlockingStateType blockingStateType,
                                                                           @Nullable final UUID blockableId) {
-    	return blockingStates.stream()
+    	final Collection<BlockingState> blockingStatesToConsider = includeDeletedEvents ? blockingStatesWithDeletedEvents : blockingStates;
```
sbrossie, line 519: Hum... The issue may be that this is called by `DefaultEventsStream#computeEntitlementBlockingStates` to compute `subscriptionEntitlementStates` in the `setup` method. So this would mean that we would alway return the `deleted` blocking states for `ENTITLEMENT_SERVICE` is the instance variable was set with `includeDeletedEvents`.

I think we want to never return any deleted blocking states (or subscription events) unless explicitly called from the `DefaultSubscription#getSubscriptionEvents`

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
reshmabidikar: I added this to prevent creating an active transition corresponding to a deleted event.

See this [test](https://github.com/reshmabidikar/killbill/blob/835a6a5cf9344da4effd058bd05167f6b6301794/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java#L1014). When the base subscription is uncancelled [here](https://github.com/reshmabidikar/killbill/blob/835a6a5cf9344da4effd058bd05167f6b6301794/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java#L1078), the base cancel event is deleted. However, when `includeDeleted=true` is passed, this returns the deleted cancel event and results in the `baseTriggerEventForAddOnCancellation` getting initialized [here](https://github.com/reshmabidikar/killbill/blob/835a6a5cf9344da4effd058bd05167f6b6301794/subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java#L844). This in turn results in a cancel event getting created for the addon. By default, it is created with `isActive=true` which results in it being used to create an active transition [here](https://github.com/reshmabidikar/killbill/blob/835a6a5cf9344da4effd058bd05167f6b6301794/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java#L845). Hence it needs to be set with the `isActive` value in the `baseTriggerEventForAddOnCancellation`.

entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
reshmabidikar: Yes, I'm not happy with this change either, I had to make this change to prevent [this](https://github.com/reshmabidikar/killbill/blob/31f26ddc7522f440b7f4c19327f6406ec7424af5/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java#L1014) addon test from failing as mentioned [here](https://github.com/killbill/killbill/pull/1762#issuecomment-1230094693). 

So basically, without this change, the [entitlementCancelEvent](https://github.com/reshmabidikar/killbill/blob/835a6a5cf9344da4effd058bd05167f6b6301794/entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java#L81) field does not get populated due to which the `STOP_ENTITELMENT` event is not returned for the addon subscription causing the test to fail [here](https://github.com/reshmabidikar/killbill/blob/31f26ddc7522f440b7f4c19327f6406ec7424af5/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java#L1118).


The alternative would be maybe to create a separate list like `subscriptionEntitlementStatesWithDeletedEvents` and use that to compute `entitlementCancelEvent` and other fields as required. Please let me know what you think.

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
sbrossie: I see what you are saying, and what you have is probably OK, but I think we could simplify even a bit more: Regardless of the mode, `includeDeletedEvents=true|false`, I think we should only add a new in-memory event in the stream for the add-on when there is an active event. So instead of just looking at whether `baseTriggerEventForAddOnCancellation` is not null we should check whether `baseTriggerEventForAddOnCancellation` is not null and is active [here](https://github.com/reshmabidikar/killbill/blob/work-for-release-0.23x-1030-blocking-states/subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java#L877).

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
sbrossie: This comment is linked to [this one](https://github.com/killbill/killbill/pull/1762#discussion_r962429438), same method and part of the same logic.

What I'd like to suggest as a behavior is that if we cancel a BASE and then uncancel it, then fetching the events with `includeDeletedEvents=true` does return the canceled event for the base, but it does not add any in-memory deleted events for the AOs (if any). So this assertion [here](https://github.com/reshmabidikar/killbill/blob/835a6a5cf9344da4effd058bd05167f6b6301794/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestSubscription.java#L1118) would need to be changed, as we would not see any in-memory deleted events.

I think reverting the code in SubscriptionDao would just work (to be verified):
Screen Shot 2022-09-08 at 6 23 22 PM

subscription/src/main/java/org/killbill/billing/subscription/engine/dao/DefaultSubscriptionDao.java
reshmabidikar: Fixed

entitlement/src/main/java/org/killbill/billing/entitlement/engine/core/DefaultEventsStream.java
reshmabidikar: Reverted these changes and updated test.

sbrossie (approved):

PR #1778, opened 2022-10-04
Merge master to work for release 0.23.x

xsalefter:
@sbrossie @reshmabidikar do you mind to review my merge changes especially for 
- [SubscriptionApiBase](https://github.com/killbill/killbill/pull/1778/files#diff-699f89a95013c79d65f3a2489c333b27c2dc0d176b9ef086afad26a71dbaf997)
- [DefaultSubscriptionBase](https://github.com/killbill/killbill/pull/1778/files#diff-6c9485b2174859a0d155b69e387e4b868e364c2608ba08fd0a2db9c49101fc46)
- [SubscriptionBaseApiService](https://github.com/killbill/killbill/pull/1778/files#diff-7db4ffeb96edd7d1fdfbbfb0193eb65ae5a88db6932092a7424c2e9bb2f330dd)

I also add `FIXME` for various changes that I made but not sure if that's what expected. For example, [this one](https://github.com/killbill/killbill/pull/1778/files#diff-39b9f6d3a9fd8f86565a96058830f6184e1c3d62938aa6e685e6bd420d91e08f).

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
@@ -83,6 +83,7 @@ import org.killbill.commons.utils.Preconditions;
 import org.killbill.billing.util.callcontext.CallContext;
 import org.killbill.billing.util.callcontext.InternalCallContextFactory;
 import org.killbill.billing.util.callcontext.TenantContext;
+import org.killbill.billing.util.config.definition.SubscriptionConfig;
 import org.killbill.commons.utils.collect.MultiValueHashMap;
 import org.killbill.commons.utils.collect.MultiValueMap;
 import org.killbill.clock.Clock;
@@ -100,6 +101,8 @@ public class DefaultSubscriptionBaseApiService implements SubscriptionBaseApiSer
     private final SubscriptionCatalogApi subscriptionCatalogApi;
     private final PlanAligner planAligner;
     private final AddonUtils addonUtils;
+
+    private final SubscriptionConfig config;
     private final InternalCallContextFactory internalCallContextFactory;
 
     @Inject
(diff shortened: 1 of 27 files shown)


subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
```diff
@@ -279,6 +281,7 @@ private boolean doCancelPlan(final Map<DefaultSubscriptionBase, DateTime> subscr
                     throw new SubscriptionBaseApiException(ErrorCode.SUB_CANCEL_BAD_STATE, subscription.getId(), "PENDING CANCELLED");
                 }
                 // Similarly, if subscription is cancelled with date past the expiry date (in case of a FIXEDTERM phase), we disallow the operation
+                // FIXME-MERGE-MASTER-TO-WFR-023x: Is this should be exist?
```
sbrossie, line 284: I am not sure I understand what is meant here?

sbrossie (approved):
@xsalefter I can't spot any issue with the merge. Regarding the `FIXME-MERGE-MASTER-TO-WFR-023x`:
* For FIXME-MERGE-MASTER-TO-WFR-023x: set 'includeDeletedEvents=false' parameter here => you can remove those, keeping a default of `false` for our tests is the right approach
* FIXME-MERGE-MASTER-TO-WFR-023x: set 'pluginProperty=emptyList()'  => you can remove those as well, this is also a sane default.
* I left a comment for the remaining FIXME-MERGE-MASTER-TO-WFR-023x

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBaseApiService.java
xsalefter: I actually asking whether these lines:

```java
  if (pendingTransition != null &&
    pendingTransition.getTransitionType() == SubscriptionBaseTransitionType.EXPIRED &&
    pendingTransition.getEffectiveTransitionTime().compareTo(effectiveDate) < 0) {
      throw new SubscriptionBaseApiException(ErrorCode.SUB_CANCEL_BAD_STATE, subscription.getId(), "PENDING EXPIRY");
  }
```

should be removed or keep them exist. Because the diff from you and Reshma is different, and I'm not quite sure which one that is correct one. But I think is should be there, and thus I'll remove the `FIXME` line.

Issue #1272, opened 2020-02-05
Account invoices pagination API


pierre:
When accounts have a large number of invoices, the call `/1.0/kb/accounts/{accountId}/invoices` isn't performant. We should expose a pagination-like API instead.

pierre:
The existing call could also be optimized today if we were to return shallow objects (`withInvoiceItems=false`):

https://github.com/killbill/killbill/blob/b74f63d7e6e085155a266fc4f2cbad4db4ed892e/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L218

reshmabidikar:
Quick question, do we want to keep the existing endpoint for backward compatibility? Or should the new pagination endpoint replace the existing one?

reshmabidikar:
Fixed in the following PRs:

- #1780
- #1789

PR #1780, opened 2022-10-07
Changes for 1272

reshmabidikar:


invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -164,43 +164,47 @@ protected InvoiceApiException generateAlreadyExistsException(final InvoiceModelD
...
-                                       (includeVoidedInvoices || !InvoiceStatus.VOID.equals(invoice.getStatus())))
-                    .sorted(INVOICE_MODEL_DAO_COMPARATOR)
-                    .collect(Collectors.toUnmodifiableList());
+                                                                            .filter(invoice -> !invoice.isMigrated() &&
+                                                                                               (includeVoidedInvoices || !InvoiceStatus.VOID.equals(invoice.getStatus())))
+                                                                            .sorted(INVOICE_MODEL_DAO_COMPARATOR)
+                                                                            .collect(Collectors.toUnmodifiableList());
 
-            invoiceDaoHelper.populateChildren(invoices, invoicesTags, entitySqlDaoWrapperFactory, context);
+            if (withInvoiceItems) {
```
sbrossie, line 179: The `populateChildren` does more than pulling the invoice items, so eliminating calling this based on the value of this flag seems incorrect (too restrictive). Are we trying to just return 'shell' invoices (i.e not populated), in which case we should change the name of this flag, or are we simply trying to not pull invoice items but keep everything else the way `populateChildren` does?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -491,7 +491,7 @@ public List<SubscriptionBundle> call() throws Exception {
         final Callable<List<Invoice>> invoicesCallable = new Callable<List<Invoice>>() {
             @Override
             public List<Invoice> call() throws Exception {
-                return invoiceApi.getInvoicesByAccount(accountId, false, false, tenantContext);
+                return invoiceApi.getInvoicesByAccount(accountId, false, false, true, tenantContext); //TODO_1272: should withInvoiceItems be set to false to optimize this call?
```
sbrossie, line 494: I believe fetching the invoice items is required: If you follow the code and look at how this `invoicesCallable` is being used, you end up `AccountTimelineJson` where such [items are being used](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/AccountTimelineJson.java#L82)

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -671,11 +672,11 @@ public Response getInvoicesForAccount(@PathParam("accountId") final UUID account
 
         final List<Invoice> invoices;
         if (unpaidInvoicesOnly) {
-            invoices = new ArrayList<Invoice>(invoiceApi.getUnpaidInvoicesByAccountId(accountId, startDate, endDate, tenantContext));
+            invoices = new ArrayList<Invoice>(invoiceApi.getUnpaidInvoicesByAccountId(accountId, startDate, endDate, tenantContext)); //TODO_1272:  withInvoiceItems not passed here as it is not possible to get unpaid invoices without retrieving invoice items. Maybe add a check at the start to disallow passing both the unpaidInvoicesOnly and withInvoiceItems flags?
```
sbrossie, line 675: Ideally, this api also allows to not fetch invoice items, but this is just an optimization issue, and hopefully the number of unpaid invoices is small, so fetching invoice items when not required is probably ok.

sbrossie (commented):
I think I may not have all the context required here: The issue https://github.com/killbill/killbill/issues/1272 talks about Account invoices pagination API, but what I am seeing is that we simply allow to control fetching invoices without items now. Perhaps this was discussed already, and I missed it: Could you re-explain how this solves the issue?

Regarding impl. I left a few comments that I think require attention.

reshmabidikar:
> The issue https://github.com/killbill/killbill/issues/1272 talks about Account invoices pagination API

This PR only addresses adding an additional flag as per the comment [here](https://github.com/killbill/killbill/issues/1272#issuecomment-582296393). I will be implementing the pagination logic as part of a separate PR. I should have mentioned this earlier, sorry for the confusion.

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
reshmabidikar: Taking a look at the code again, I see what you mean. So rather than using the flag here, it should be passed to the `populateChildren` method and used at [Line#255](https://github.com/killbill/killbill/blob/35d35aeffc512fc2052b7c65aa686a3a3865b4ba/invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java#L255) to determine whether `setInvoiceItemsWithinTransaction` should be invoked or not. Does that sound right?

sbrossie: If we want to really honor what this flag says, then yes. But more importantly, what is the behavior we really want? 
/cc @pierre Do we implement a way to retrieve shell invoices (i.e we don't populate anything), or we simply don't pull the invoice items? 

My thoughts: I think we really want the former.

pierre: > Do we implement a way to retrieve shell invoices (i.e we don't populate anything)

Yup, that's what we want. Pulling anything else and/or computing amount/balance/etc. will be a performance hit.

Skipping `invoiceDaoHelper.populateChildren` as it is now looks like the right approach to me.

reshmabidikar: Left the behavior as it is and renamed the flag to `includeInvoiceComponents`

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
reshmabidikar: Changed to code to add a check that disallows both `unpaidInvoicesOnly` and `includeInvoiceComponents` to be specified at once.

PR #1783, opened 2022-10-15
Updated TenantMetaDeta for Tenant Config and User Tag Change and Deletion

KBbitsP:
Updated Tenant Metadata for TENANT_CONFIG_CHANGE and TENANT_CONFIG_DELETION to save meta data as Object instead of string

@reshmabidikar @pierre


pierre:
One thought: I suspect https://github.com/killbill/killbill-plugin-framework-java/blob/69156e4100f55932d7a59ed46a3119f0b824100a/src/main/java/org/killbill/billing/plugin/api/notification/PluginConfigurationEventHandler.java#L38 needs to be changed as well (we should also test manually the reconfiguration of plugins works as expected after this change).

KBbitsP:
> One thought: I suspect https://github.com/killbill/killbill-plugin-framework-java/blob/69156e4100f55932d7a59ed46a3119f0b824100a/src/main/java/org/killbill/billing/plugin/api/notification/PluginConfigurationEventHandler.java#L38 needs to be changed as well (we should also test manually the reconfiguration of plugins works as expected after this change).

@pierre I think we don't need this check for tenant config. I should remove this check.

pierre:
> @pierre I think we don't need this check for tenant config. I should remove this check.

We totally need that check 😅 

`PluginConfigurationEventHandler` is used to monitor per-plugin configuration changes (https://docs.killbill.io/latest/plugin_development.html#_enabling_per_tenant_configuration). The config handlers specified by the plugin developer are run when these special config events are sent.

Since `extBusEvent.getMetaData()` isn't a simple string anymore, I suspect the underlying code needs some updates.

beatrix/src/main/java/org/killbill/billing/beatrix/extbus/BeatrixListener.java
```diff
@@ -293,31 +295,35 @@ private BusEvent computeExtBusEventEntryFromBusInternalEvent(final BusInternalEv
...
+                metaData = objectMapper.writeValueAsString(userTagDeletionMetadata);
                 break;
 
             case CONTROL_TAG_DELETION:
                 final ControlTagDeletionInternalEvent realTagEventDel = (ControlTagDeletionInternalEvent) event;
                 objectType = ObjectType.TAG;
                 objectId = realTagEventDel.getTagId();
                 eventBusType = ExtBusEventType.TAG_DELETION;
-                metaData = realTagEventDel.getTagDefinition().getName();
+                final TagMetadata controlTagDeletionMetadata = new TagMetadata(realTagEventDel.getTagId(), realTagEventDel.getObjectId(), realTagEventDel.getObjectType(), realTagEventDel.getTagDefinition());
```
reshmabidikar, line 325: Even if the `objectId` and `objectType` fields are included in the `TagMetadata` class, I do not think populating the `objectType` from the internal event would not be correct. So I would suggest the following:
```suggestion
final TagMetadata controlTagDeletionMetadata = new TagMetadata(realTagEventDel.getTagId(), realTagEventDel.getObjectId(), objectType, realTagEventDel.getTagDefinition());
```

KBbitsP: Hi @reshmabidikar 

I have changed the code as per the above comments

PR #1784, opened 2022-10-19
1272 - Dao changes for pagination logic

reshmabidikar:
@sbrossie, I've created this PR with only the DAO layer changes for the pagination logic for #1272. Could you please take a look just to confirm if this is the right approach? Also, the code still has a few TODO statements, please ignore these, for now. I will clean this up soon.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
@@ -185,9 +185,21 @@ public class DefaultInvoiceDao extends EntityDaoBase<InvoiceModelDao, Invoice, I
     }
 
     @Override
+    public Pagination<InvoiceModelDao> getPaginatedInvoicesByAccount(final Boolean includeVoidedInvoices, final Boolean includeInvoiceComponents, final Long offset, final Long limit,final InternalTenantContext context) {
+        final List<Tag> invoicesTags = getInvoicesTags(context);
+        return transactionalSqlDao.execute(true, entitySqlDaoWrapperFactory -> invoiceDaoHelper.getInvoicesByAccountFromTransactionPaginated(includeVoidedInvoices, includeInvoiceComponents, invoicesTags, entitySqlDaoWrapperFactory, offset, limit, paginationHelper, context));
+    }
+
+    @Override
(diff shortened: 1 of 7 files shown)


invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -184,12 +184,24 @@ public List<InvoiceModelDao> getInvoicesByAccount(final Boolean includeVoidedInv
         });
     }
 
+    @Override
+    public Pagination<InvoiceModelDao> getPaginatedInvoicesByAccount(final Boolean includeVoidedInvoices, final Boolean includeInvoiceComponents, final Long offset, final Long limit, final InternalTenantContext context) {
+        final List<Tag> invoicesTags = getInvoicesTags(context);
+        return transactionalSqlDao.execute(true, entitySqlDaoWrapperFactory -> invoiceDaoHelper.getInvoicesByAccountFromTransactionPaginated(includeVoidedInvoices, includeInvoiceComponents, invoicesTags, entitySqlDaoWrapperFactory, offset, limit, paginationHelper, context));
```
sbrossie, line 190: I think there is a problem with the approach: Most of our DAO methods are built using the `transactionalSqlDao.execute(true, wrapper)`, where the `wrapper` is of type `EntitySqlDaoTransactionWrapper`, i.e an interface whose implementation will run some code **meant to be run within a transaction**.

However, in your case, you pass `invoiceDaoHelper.getInvoicesByAccountFromTransactionPaginated(...)` which is not really a bunch of queries to be run in a transaction but rather a delegation to the pagination helper, which will itself [start a new transaction and run queries within the transaction](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/util/src/main/java/org/killbill/billing/util/entity/dao/DefaultPaginationSqlDaoHelper.java#L70) (so we end up with two transactions one inside the other, and perhaps it ends up working but this is not the implementation we want).

I think we need to simplify and do something similar for other pagination dao apis, where we simply invoke the pagination helper, as shown [here](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/payment/src/main/java/org/killbill/billing/payment/dao/DefaultPaymentDao.java#L607). The logic for the queries in embedded in the `PaginationIteratorBuilder#build` method.

reshmabidikar:
Closing this PR, will be opening a different one.

PR #1789, opened 2022-11-04
Work for release 0.23.x 1272 pagination shallow

reshmabidikar:


invoice/src/test/java/org/killbill/billing/invoice/api/user/TestDefaultInvoiceUserApi.java
```diff
@@ -701,4 +702,42 @@ public void testCreditWithoutDraftInvoice() throws Exception {
...
+    
+    @Test(groups = "slow")
+    public void testRetrieveInvoicesByAccount() throws Exception {
+
+        final Account account = invoiceUtil.createAccount(callContext);
+        final UUID accountId = account.getId();
+
+        // Create invoice1
+        final InvoiceItem externalCharge1 = new ExternalChargeInvoiceItem(null, accountId, null, UUID.randomUUID().toString(), clock.getUTCToday(), null, new BigDecimal(300), accountCurrency, null);
+        invoiceUserApi.insertExternalCharges(accountId, clock.getUTCToday(), List.of(externalCharge1), true, null, callContext).get(0);
```
pierre, line 714: What's the `get(0)` (same question below)?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -694,6 +693,58 @@ public Response getInvoicesForAccount(@PathParam("accountId") final UUID account
...
     }
 
+    @TimedResource
+    @GET
+    @Path("/{accountId:" + UUID_PATTERN + "}/" + INVOICES + "/" + PAGINATION)
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Retrieve account invoices", response = InvoiceJson.class, responseContainer = "List")
+    @ApiResponses(value = {@ApiResponse(code = 400, message = "Invalid account id supplied"),
+                           @ApiResponse(code = 404, message = "Account not found")})
+    public Response getInvoicesForAccountWithPagination(@PathParam("accountId") final UUID accountId,
```
pierre, line 703: Could we just nam it to `getInvoicesForAccount`? The name here will dictate the Java client API.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -694,6 +693,58 @@ public Response getInvoicesForAccount(@PathParam("accountId") final UUID account
...
+                                          @QueryParam(QUERY_INVOICES_FILTER) final String invoicesFilter,
+                                          @QueryParam(QUERY_SEARCH_OFFSET) @DefaultValue("0") final Long offset,
+                                          @QueryParam(QUERY_SEARCH_LIMIT) @DefaultValue("100") final Long limit,
+                                          @QueryParam(QUERY_AUDIT) @DefaultValue("NONE") final AuditMode auditMode,
+                                          @javax.ws.rs.core.Context final HttpServletRequest request) throws AccountApiException {
+
+        final TenantContext tenantContext = context.createTenantContextWithAccountId(accountId, request);
+
+        // Verify the account exists
+        accountUserApi.getAccountById(accountId, tenantContext);
```
pierre, line 713: Interesting -- we typically don't do that though. @sbrossie Any thought?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
```diff
@@ -694,6 +693,58 @@ public Response getInvoicesForAccount(@PathParam("accountId") final UUID account
...
+
+        final Pagination<Invoice> invoices = invoiceApi.getInvoicesByAccount(accountId, offset, limit, tenantContext);
+
+        //TODO_1272: Delete this filter?
```
pierre, line 717: Yes, I would remove it as the offset / count would be wrong in the Pagination object. If / When we want to introduce it, we should implement a `search` API instead (I don't suggest we introduce it just yet though).

util/src/main/java/org/killbill/billing/util/entity/dao/DefaultPaginationSqlDaoHelper.java
```diff
@@ -85,15 +106,16 @@ public Long inTransaction(final EntitySqlDaoWrapperFactory entitySqlDaoWrapperFa
             // The count to get maxNbRecords can be expensive on very large datasets. As a heuristic to check how large that number is,
             // we retrieve 1 record at offset simplePaginationThreshold (pretty fast). If we've found a record, that means the count is larger
             // than this threshold and we don't issue the full count query
-            final Long recordIdAtSimplePaginationOffset = sqlDao.getRecordIdAtOffset(simplePaginationThreshold);
+            final Long recordIdAtSimplePaginationOffset = withAccountRecordId ? sqlDao.getRecordIdAtOffsetWithAccountRecordId(simplePaginationThreshold, context) : sqlDao.getRecordIdAtOffset(simplePaginationThreshold);
```
pierre, line 109: Could you add a few more test cases in `TestDefaultPaginationSqlDaoHelper`? I think it's worth testing it (e.g. insert 10 records, and go through them 2 at a time to verify the `Pagination` fields).

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -184,6 +184,27 @@ public List<InvoiceModelDao> getInvoicesByAccount(final Boolean includeVoidedInv
...
     }
 
+    @Override
+    public Pagination<InvoiceModelDao> getInvoicesByAccount(final Long offset, final Long limit, final InternalTenantContext context) {
+
+        final Pagination<InvoiceModelDao> invoices = paginationHelper.getPaginationWithAccountRecordId(InvoiceSqlDao.class,
+                                                                                    new PaginationIteratorBuilder<InvoiceModelDao, Invoice, InvoiceSqlDao>() {
+                                                                                        @Override
+                                                                                        public Long getCount(final InvoiceSqlDao invoiceSqlDao, final InternalTenantContext context) {
+                                                                                            return null;
```
pierre, line 194: We should be able to get the count, no?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
reshmabidikar: This is already there in the existing `getInvoicesForAccount` method (See this [line](https://github.com/killbill/killbill/blob/c9ca6ca5d5f3310205d6baabee372bb2e007cdea/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java#L689)). I can delete this line from both methods if it is unnecessary.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
reshmabidikar: Using the `getInvoicesForAccount` name causes an issue in generating the next page URI [here](https://github.com/reshmabidikar/killbill/blob/c58d2684202706005f97cf3ef07daf21c73930db/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java#L730) since there is another `getInvoicesForAccount` method [here](https://github.com/reshmabidikar/killbill/blob/c58d2684202706005f97cf3ef07daf21c73930db/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java#L651). Would you like me to look into the logic of `uriBuilder.nextPage` to see if there is any way to work around this?

pierre: oh, I see. Never mind then!

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
pierre: oh, I had no idea! From 2013: https://github.com/killbill/killbill/commit/b15041255b6ffefdc95c6d65027f1f583002e09a#diff-412208bd13addbd00f2f80dde17031143060e334f320967752451977cf8c5e2aR388 😅

sbrossie: In don't remember the details, I suspect this was to make sure we were throwing the right `AccountException` and as a result returning the correct http error.

For simplicity, keep it the same - based on time, it would be interesting to try without it, and see what happens with a non existent accountId (if result is the same, remove it).

util/src/main/java/org/killbill/billing/util/entity/dao/DefaultPaginationSqlDaoHelper.java
sbrossie: I agree, spending time writing some good DAO tests to verify all these queries is a good investment.

invoice/src/test/java/org/killbill/billing/invoice/api/user/TestDefaultInvoiceUserApi.java
reshmabidikar: It's not required, deleted it.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
reshmabidikar: Sure, I'll try deleting that line. If it works, I will create a separate commit.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
reshmabidikar: Fixed.

util/src/main/java/org/killbill/billing/util/entity/dao/DefaultPaginationSqlDaoHelper.java
reshmabidikar: I have now added some tests to `TestDefaultPaginationSqlDaoHelper`. Please take a look and let me know how these look. Also, there are some tests in [TestInvoiceDao](https://github.com/reshmabidikar/killbill/blob/ac2e822b14844eea8f88cc5b8e775020975efa53/invoice/src/test/java/org/killbill/billing/invoice/dao/TestInvoiceDao.java#L248) as well.

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/AccountResource.java
reshmabidikar: When the endpoint is queried with an invalid `accountId` it displays the following both with and without the account exists check:

```
{"className":"org.killbill.billing.util.callcontext.InternalCallContextFactory$ObjectDoesNotExist","code":null,"message":"Object id=7edbcbfa-da58-492e-a8df-1baadb008485 type=ACCOUNT doesn't exist!","causeClassName":null,"causeMessage":null,"stackTrace":[]}
```

So, looks like the account exists check has no effect. So, I've deleted it as part of [this ](https://github.com/killbill/killbill/pull/1789/commits/f78e2a9278eae56fa45e6d0791c8e922683b9949)commit.

util/src/test/java/org/killbill/billing/util/entity/dao/TestDefaultPaginationSqlDaoHelper.java
```diff
@@ -83,6 +83,65 @@ public void testSearch1MatchOnLargeTable() {
         searchAndValidateKombuchas("jasmin", 0L, 100L, 4, 1L, 1L, null, 0L, null);
     }
 
+    @Test(groups = "slow")
+    public void testListKombuchasAll() {
+        insertKombuchas(4);
```
pierre, line 88: The kombuchas are all for the same account today, would be worth inserting rows with other `account_record_id` to validate we don't return some rows for other accounts (and that the count queries are correct).

pierre (commented):
Just a test thought, otherwise :shipit:!

Also, maybe regenerate the Java client and add server tests?

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
reshmabidikar: Fixed

util/src/test/java/org/killbill/billing/util/entity/dao/TestDefaultPaginationSqlDaoHelper.java
reshmabidikar: Fixed. The code uses the `accountRecordId` from `internalCallContext` while inserting a `kombucha`. I've added a hack to set the `internalCallContext.accountRecordId` to a dummy value. Let me know if this is okay or if there is a better way to do this.

reshmabidikar:
> Also, maybe regenerate the Java client and add server tests?

I've regenerated the client and also written a test, was just waiting for this PR to be merged in case there are additional changes 😁

pierre (approved):

PR #1794, opened 2022-11-14
entitlement: Remove hack introduced to set subscription externalKey while doing bundle transfer.

sbrossie:
The hack was introduced [here](https://github.com/killbill/killbill/commit/80d78a3cdfffc3f48705d6467813bd9a24f968da#diff-7693b3c207b4531f18416f247cde014ec778db5e0fd5055039661b45d4053d02R286). Instead we extended the api as shown in this [PR](https://github.com/killbill/killbill-api/pull/109) to allow passing a Map.

The JAXRS api does not change as we leverage the existing body to pass additional info (instead of relying on plugin properties).

This only removes the hack and does not implement additional functionalities. See #1627


reshmabidikar (approved):
Looks good!

sbrossie:
Merging the PR - hoping for the best with tests as they are currently unstable.

Issue #1792, opened 2022-11-11
Implement per-account subscriptions pagination


pierre:
Similar work than per-account invoices pagination. See #1272.

reshmabidikar:
Fixed in https://github.com/killbill/killbill/pull/1798

PR #1798, opened 2022-11-16
Changes for 1792

reshmabidikar:
- Includes Changes for #1792
- Minor change for #1272 (Renamed method to keep it consistent with the bundle pagination method)


util/src/main/java/org/killbill/billing/util/entity/dao/EntityDaoBase.java
```diff
@@ -221,6 +221,26 @@ public Iterator<M> build(final EntitySqlDao<M, E> sqlDao, final Long offset, fin
...
+                                                                     public Long getCount(final EntitySqlDao<M, E> sqlDao, final InternalTenantContext context) {
+                                                                         // Only need to compute it once, because no search filter has been applied (see DefaultPaginationSqlDaoHelper)
+                                                                         return null;
+                                                                     }
+
+                                                                     @Override
+                                                                     public Iterator<M> build(final EntitySqlDao<M, E> sqlDao, final Long offset, final Long limit, final Ordering ordering, final InternalTenantContext context) {
+                                                                         return sqlDao.getByAccountRecordIdWithPaginationEnabled(offset, limit, context); //TODO_1792: Is this correct or is a new query (similar to EntitySqlDao.sql.stg#Line147 but with accountRecordId) required? This query does not take into account the ordering parameter
```
pierre, line 236: I'm trying to understand why this is needed / why it's not symmetrical to https://github.com/killbill/killbill/pull/1789. Is it because we go through the internal API?

reshmabidikar: Good catch! [DefaultInvoiceDao.getInvoicesForAccount](https://github.com/killbill/killbill/blob/7ea3acf0a95a520991985904b682d7ddc8a34e48/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L188) is no longer required as the same logic is written in `EntityDaoBase.getByAccountRecordId` (for subscription bundles) and can be used for invoices as well. I've now deleted the code from `InvoiceDao`.

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultSubscriptionApi.java
```diff
@@ -239,6 +239,27 @@ public List<SubscriptionBundle> getSubscriptionBundlesForAccountId(final UUID ac
         return getSubscriptionBundlesForAccount(accountId, context);
     }
 
+    @Override
+    public Pagination<SubscriptionBundle> getSubscriptionBundlesForAccountId(final UUID accountId, final Long offset, final Long limit, final TenantContext context) throws SubscriptionApiException {
+        final InternalTenantContext internalContext = internalCallContextFactory.createInternalTenantContext(accountId, context);
+        return getEntityPaginationNoException(limit,
```
sbrossie, line 245: Since the `subscriptionBaseInternalApi#getBundlesForAccount`  throws an exception, should we instead use `getEntityPagination` instead of `getEntityPaginationNoException`. @pierre Thoughts?

reshmabidikar: The `subscriptionBaseInternalApi#getBundlesForAccount` method does not throw an exception. I think I had added a `throws` clause to the [SubscriptionApi.getSubscriptionBundlesForAccountId](https://github.com/killbill/killbill-api/blob/be2a8eb8c1657636c234864cd27cc5bcdfd543be/src/main/java/org/killbill/billing/entitlement/api/SubscriptionApi.java#L140) method to keep it consistent with the [non-paginated version](https://github.com/killbill/killbill-api/blob/be2a8eb8c1657636c234864cd27cc5bcdfd543be/src/main/java/org/killbill/billing/entitlement/api/SubscriptionApi.java#L130) of the method (which specifies a `throws` clause). But, now I think it can be removed since the paginated version would not be throwing an exception. Let me know what you think, if you agree, I will remove the `throws` clause from `DefaultSubscriptionApi.getSubscriptionBundlesForAccountId` and use `getEntityPaginationNoException`.

sbrossie: Yes, I see, the exception handling is in the mapper function below line 255. Our current framework does not expect such mapping function to throw exceptions, but here, we are trying to rehydrate the object (i.e additional queries, ...) and exception handling is therefore needed.

Isn't it similar to the discussion we had for paginating Invoices [here](https://github.com/killbill/killbill/pull/1789/files#diff-a75da8f7888dc03193606d82e152964f60580bf1fd8f56f44201695e3d1a7f34R177), and where we decided we would return shallow invoices? 

@pierre Do we want to rehydrate all the subscriptionBundle being paginated?

pierre: Agreed - we should make it symmetrical to invoices.

reshmabidikar: The mapper function converts `SubscriptionBaseBundle` to `SubscriptionBundle`, it does not rehydrate the subscription. 

This code is similar to the existing [getSubscriptionBundles](https://github.com/killbill/killbill/blob/1c7581cc0fd22fdbd23c95d83dd2e3f836bc2ec5/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultSubscriptionApi.java#L243) ( which returns paginated subscription bundles without `accountRecordId`).

If we would like to avoid the mapping code, we would also need to change the [internal API](https://github.com/reshmabidikar/killbill/blob/fb64e0a0094e1373c7a0940b18a8afb8f9f9d031/subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java#L210) (and dao) but then the code would looks different from the[ existing internal API code](https://github.com/killbill/killbill/blob/1c7581cc0fd22fdbd23c95d83dd2e3f836bc2ec5/subscription/src/main/java/org/killbill/billing/subscription/api/svcs/DefaultSubscriptionInternalApi.java#L215).

Let me know your thoughts.

entitlement/src/main/java/org/killbill/billing/entitlement/dao/ProxyBlockingStateDao.java
```diff
@@ -366,4 +366,10 @@ private static boolean isEntitlementCancellationBlockingState(final BlockingStat
                KILLBILL_SERVICES.ENTITLEMENT_SERVICE.getServiceName().equals(blockingState.getService()) &&
                DefaultEntitlementApi.ENT_STATE_CANCELLED.equals(blockingState.getStateName());
     }
+
+    @Override
+    public Pagination<BlockingStateModelDao> getByAccountRecordId(final Long offset, final Long limit,
+                                                                  final InternalTenantContext context) {
+        throw new RuntimeException("Not Implemented"); //TODO_1792 - Does this need to be handled in any other way?
```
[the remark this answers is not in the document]
reshmabidikar: Fixed

pierre (approved):

entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultSubscriptionApi.java
sbrossie: I had not realized the code was already like that for `DefaultSubscriptionApi #getSubscriptionBundles` - it's a bit weird to have to do an extra (set of) query(ies) for every item being paginated.

sbrossie (approved):

PR #1801, opened 2022-11-19
1799 fix utils module

xsalefter:


util/src/main/java/org/killbill/billing/util/audit/DefaultAccountAuditLogs.java
```diff
@@ -46,7 +46,7 @@ public DefaultAccountAuditLogs(final UUID accountId, final AuditLevel auditLevel
         this.accountId = accountId;
         this.auditLevel = auditLevel;
         // TODO pierre - lame, we should be smarter to avoid loading all entries in memory. It's a bit tricky though...
-        this.accountAuditLogs = Iterators.toUnmodifiableList(accountAuditLogsOrderedByTableName);
+        this.accountAuditLogs = accountAuditLogsOrderedByTableName;
```
pierre, line 49: I don't think this would work: `accountAuditLogsOrderedByTableName` essentially _points_ to an open DB connection and we don't want to keep that connection open for too long (i.e. for the lifetime of the `DefaultAccountAuditLogs` object -- hence my _it's tricky_ comment 😄).

PR #1803, opened 2022-11-22
close pagination after assertion completed

xsalefter:
During debugging https://github.com/killbill/killbill/issues/1799 , where I purposely set connection pool to 1, I found that closing pagination after use helpful to make connection available.

### DIFF
--- a/util/src/test/java/org/killbill/billing/util/entity/dao/TestDefaultPaginationSqlDaoHelper.java
+++ b/util/src/test/java/org/killbill/billing/util/entity/dao/TestDefaultPaginationSqlDaoHelper.java
@@ -17,6 +17,8 @@
 
 package org.killbill.billing.util.entity.dao;
 
+import java.io.Closeable;
+import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
 
@@ -156,6 +158,8 @@ public class TestDefaultPaginationSqlDaoHelper extends UtilTestSuiteWithEmbedded
     	Assert.assertEquals(pagination.getMaxNbRecords(), expectedMaxNbRecords);
     	Assert.assertEquals(pagination.getCurrentOffset(), expectedCurrentOffset);
     	Assert.assertEquals(pagination.getNextOffset(), expectedNextOffset);
+
+        closePagination(pagination);
     }
 
     private void searchAndValidateKombuchas(final String tea,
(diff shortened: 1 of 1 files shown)


sbrossie (commented):
Couple of questions:
1. How did you identify the source(s) for the leak - was this test the only place?
2. Did you try resetting the connection pool to 1 to see if this now works (more as an experiment, we can leave to the value you set as a default)?

xsalefter:
@sbrossie 

> How did you identify the source(s) for the leak - was this test the only place?

No, this test not the only place, but this tests are the easiest spot to fix. I found several leak (and probably more that I'm not aware), but the fix is not so obvious. This PR essentially neat version of my [chaotic PR here](https://github.com/killbill/killbill/pull/1801).

> Did you try resetting the connection pool to 1 to see if this now works

Yes. I re-run to make sure. Here' the detail:

- set connection pool = 1, use `work-for-release-0.23.x` branch:
  - `TestDefaultPaginationSqlDaoHelper` : fails.
  - `TestDefaultAccountUserApi` : fails.


- set connection pool = 1, use `close_pagination_after_use` branch:
  - `TestDefaultPaginationSqlDaoHelper` : pass.
  - `TestDefaultAccountUserApi` : fails (a.k.a, seems like I got false positive for this class :sweat_smile: )


- set connection pool = 2, use `work-for-release-0.23.x` branch:
  - `TestDefaultPaginationSqlDaoHelper` : fails.
  - `TestDefaultAccountUserApi` : pass.


- set connection pool = 2, use `close_pagination_after_use` branch:
  - `TestDefaultPaginationSqlDaoHelper` : pass.
  - `TestDefaultAccountUserApi` : pass.

So yeah, seems like  changes in `TestDefaultAccountUserApi` is not needed, but I'll keep close the `Pagination` after use, unless you tell me not to.

For `TestDefaultPaginationSqlDaoHelper`, seems like closing the pagination really make difference.

account/src/test/java/org/killbill/billing/account/api/user/TestDefaultAccountUserApi.java
```diff
@@ -75,20 +77,23 @@ public void testSearch() throws Exception {
         Assert.assertEquals(search1.getMaxNbRecords(), (Long) 2L);
         Assert.assertEquals(search1.getTotalNbRecords(), (Long) 2L);
         Assert.assertEquals(Iterables.toUnmodifiableList(search1).size(), 2);
+        closePagination(search1);
```
pierre, line 80: Unless I'm missing something, I'm pretty sure this isn't needed: `Iterables.toUnmodifiableList` should have consumed the whole stream, so the connection should be automatically closed at this point already (unless there is a bug).

sbrossie (approved):

Issue #1651, opened 2022-04-01
Support for quantity in APIs


pierre:
* When creating subscriptions, ability to specify a `quantity` (e.g. `price=$20/mo, quantity=$2 => invoiceItem recurring amount=$40`). Related: #1434 
* When creating external charges, specifying a `quantity` should be reflected into the invoice (e.g. `price=$20, quantity=$2 => invoiceItem amount=$40`)

shaun-forgie:
Awesome...work Pierre. This will allow for so many more customers to benefit from the Kill Bill project. Is this coming in 0.23?

sbrossie:
@shaun-forgie Could you describe the exact use case you are after?

Is this a model where you want to invoice/pay **in-advance** based on a well defined number of seats? In such a scenario, being able to create one `Plan` in the catalog with the price of 1 `seat` and allow to toggle the `quantity` when creating the subscription should be enough? In such a model, do we expect the number of seats associated with a subscription to vary, and if so, what is the expectation?

shaun-forgie:
@pierre Two scenarios I see
1. in-advance: with per seat cost defined in catalog and quantity in subscription. 
2. in-arrears: with usage records providing quantities for invoice item calculations.

In the catalog the all or top tiered policy can still be used for rating purposes. Usage records would be supplied each month to trigger changes to the base quantity defined in the subscription.

A subscription quantity update policy should be defined in the invoice plugin to update the default quantity based on the current value of the usage record received for the period.

Pricing should be driven out of the catalog with tiered usage models.

sbrossie:
@shaun-forgie  My question was only for *in-advance* use case, and I am not really asking about how it should be implemented, but more interested about 1/ the use case(s) and 2/ why price override is not a good option already (given a Plan, you can override the price on a per subscription level to adjust for the # seats) ?

shaun-forgie:
The problem I have is that prices and quantites are different things. Using price over rides to change a quantity is confusing at best.

sbrossie:
@shaun-forgie I agree it would be best to have support for `quantity`, but fundamentally, you send a `seat price` and a `quantity` and you expect the system to compute `total price` = `seat price` * `quantity` and then have that reflected on your invoice. We already have support to override `Plan#price`, so you could make the multiplication on the client side and leverage this mechanism. So aside from the convenience is there anything else missing?

[...]

PR #1805, opened 2022-11-28
Fix for 1651

sbrossie:
This PR implements the support for setting a `quantity` when creating a subscription: Setting a quantity=3 is equivalent to creating 3 times the same subscription (i.e any FIXED or RECURRING items would be billed 3 times their normal price). See #1651 

This is rather large PR, so looking at individual commits may make sense. The work is mostly at the `subscription` level and the implementation is very well aligned with the one we have for BCD support. New apis and bus events have been created as a result. This PR depends on:

* https://github.com/killbill/killbill-plugin-api/pull/29
* https://github.com/killbill/killbill-api/pull/112
* https://github.com/killbill/killbill-client-java/pull/116


subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
```diff
@@ -120,6 +120,10 @@ private SubscriptionBaseEvent createEvent(final boolean firstEvent, final Existi
                 newEvent = BCDEventData.createBCDEvent(subscription, effectiveDate, existingEvent.getBillCycleDayLocal());
                 break;
 
+            case QUANTITY_CHANGE:
+                // TODO
```
pierre, line 124: Maybe throw unsupported?

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -460,14 +460,26 @@ public ProductCategory getCategory() {
...
+        return getQuantityOrBCD(true);
+    }
+
+    @Override
+    public Integer getQuantity() {
+        return getQuantityOrBCD(false);
+    }
+
+
+    private Integer getQuantityOrBCD(final boolean bcd) {
```
pierre, line 472: Nit: I find the method a tad confusing.

Maybe `getNextValueForTransition(final SubscriptionBaseTransitionType transitionType)`?

invoice/src/main/resources/org/killbill/billing/invoice/ddl.sql
```diff
@@ -106,7 +106,7 @@ CREATE TABLE invoice_item_history (
     rate numeric(15,9) NULL,
     currency varchar(3) NOT NULL,
     linked_item_id varchar(36),
-    quantity int,
+    quantity decimal(18, 9),
```
pierre, line 109: Shouldn't this be symmetric with subscriptions?

pierre (approved):

subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
sbrossie: Just to be clear: With the current code we allow the transfer, but we lose the `quantity` as it is not 'ported' with the subscription. If I understand correctly you are proposing the prevent the transfer from happening in such case to make it clear?

Ideally, I would like to fix it in as part of #1627, which I kept opened.

invoice/src/main/resources/org/killbill/billing/invoice/ddl.sql
sbrossie: That was my thinking as well, but as I was looking for what we **already have** (in invoice), I discovered our scheme is inconsistent: [invoice_items](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/invoice/src/main/resources/org/killbill/billing/invoice/ddl.sql#L71) versus [invoice_items_history](https://github.com/killbill/killbill/blob/work-for-release-0.23.x/invoice/src/main/resources/org/killbill/billing/invoice/ddl.sql#L109)

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
sbrossie: Fixed in https://github.com/killbill/killbill/pull/1805/commits/9e05be1dc910b224d8bc28994c7d43a909225277

subscription/src/main/java/org/killbill/billing/subscription/api/transfer/DefaultSubscriptionBaseTransferApi.java
sbrossie: Followed your recommendation for now in https://github.com/killbill/killbill/pull/1805/commits/e375c3c360a421c90cebff5e262d5d9f7ca9c490

PR #1807, opened 2022-11-30
pom.xml: Experiment with shiro upgrade to 1.10.0

sbrossie:


sbrossie:
Already updated as part of parent pom.xml

PR #1809, opened 2022-12-01
Changes for 1674

reshmabidikar:
Includes changes for 1674

Following are the related PRs:

- https://github.com/killbill/killbill-api/pull/114
- https://github.com/killbill/killbill-client-java/pull/117

Note that until the above PRs are merged, tests will not be green (After merging, the correct version needs to be specified in the `killbill/pom.xml`).

PR #1815, opened 2022-12-07
1674 - Changes to clone catalog

reshmabidikar:
Changes to clone catalog as per @pierre's comment [here](https://github.com/killbill/killbill/pull/1809#discussion_r1040812211).


pierre (approved):
Much safer 👍 

I am wondering if this doesn't fix at the same time some of the "catalog corruption" issues that are sometimes reported...

PR #1820, opened 2022-12-12
catalog: Fix NPE for when there is no price

pierre:
It is not clear if plugin will return an empty list or a null list of prices when there is nothing (e.g fixed price), so this code change hardens for both situations.


sbrossie (approved):

reshmabidikar (approved):

sbrossie:
Looking at the test failure, I can't figure out why this is failing:

Screen Shot 2022-12-12 at 9 11 15 AM

@reshmabidikar @pierre If you have any thoughts/ ideas, let me know.

PR #1823, opened 2022-12-12
Emb project

sbrossie:


sbrossie:
We have implemented invoice grouping in a different way as shown in https://github.com/killbill/killbill/pull/1676

PR #1825, opened 2022-12-19
Fix for avatax plugin

reshmabidikar:
The `res.getAdditionalItems()`  returns the additional invoice items only the first time it is invoked in the case of the `Avatax` plugin. On subsequent calls, it returns an empty list which prevents the tax items from being created. With this fix in place, the tax items are created.


pierre (approved):

reshmabidikar:
With the [Avatax fix](https://github.com/killbill/killbill-avatax-plugin/pull/31), this change is not required, so closing this PR.

PR #1824, opened 2022-12-16
invoice: Introduce a greedy way to charge for in-arrear subscriptions

sbrossie:
In scenarios where we want to align RECURRING subscriptions on the day they start, and yet perform a 'bill run' later (i.e. generate all the invoices for the current month on the 1st), none of the `IN_ARREAR` nor `IN_ADVANCE` modes work very well. 

We are introducing a variation for the  `IN_ARREAR` mode which behaves just like the default use case we have, except that **when configured we allow to generate the invoice prior we reach the end of the period**. The configuration can be made at the tenant level using a new property `org.killbill.invoice.in-arrear.mode=GREEDY`.

The behavior has only been plugged for RECURRING subscriptions and not USAGE based.

Note: The PR is against our `0.22.x` `maint` branch.

PR #1827, opened 2022-12-22
In arrear new mode

sbrossie:
Re-opening [similar PR](https://github.com/killbill/killbill/pull/1824) against `master` (0.24.x).


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInArrearGreedySubscriptions.java
```diff
@@ -0,0 +1,200 @@
...
+import org.killbill.billing.entitlement.api.Entitlement.EntitlementState;
+import org.killbill.billing.invoice.api.InvoiceItemType;
+import org.killbill.billing.payment.api.PluginProperty;
+import org.killbill.billing.platform.api.KillbillConfigSource;
+import org.killbill.billing.util.config.definition.InvoiceConfig.InArrearMode;
+import org.killbill.billing.util.config.definition.InvoiceConfig.UsageDetailMode;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableList;
```
pierre, line 47: No guava on master 😄

invoice/src/test/java/org/killbill/billing/invoice/model/TestRecurringInArrear.java
```diff
@@ -152,7 +152,7 @@ private void verifyInvoiceItems(final LocalDate startDate, final LocalDate endDa
                                     final int billingCycleDayLocal, final BillingPeriod billingPeriod,
                                     final LinkedHashMap<LocalDate, LocalDate> expectedDates) throws InvalidDateSequenceException {
 
-        final RecurringInvoiceItemDataWithNextBillingCycleDate invoiceItemsWithDates = fixedAndRecurringInvoiceItemGenerator.generateInvoiceItemData(startDate, endDate, targetDate, billingCycleDayLocal, billingPeriod, BillingMode.IN_ARREAR);
+        final RecurringInvoiceItemDataWithNextBillingCycleDate invoiceItemsWithDates = fixedAndRecurringInvoiceItemGenerator.generateInvoiceItemData(startDate, endDate, targetDate, billingCycleDayLocal, billingPeriod, BillingMode.IN_ARREAR, null);
```
pierre, line 155: Nit: `internalCallContext` (and below)

pierre (changes requested):

reshmabidikar (approved):

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInArrearGreedySubscriptions.java
sbrossie: Indeed! Fix in https://github.com/killbill/killbill/pull/1827/commits/152903d48327e4ba4a381605d65b13fb6b1a1c62

@xsalefter I am wondering if we could find a way to make the build fail if/when we pull guava classes. There is no explicit dependency in the `pom.xml` - I suppose it must be working as it is a transitive dependency from `guice` ?

xsalefter: Yes, seems like it is transitive dependency from guice. 

> I am wondering if we could find a way to make the build fail if/when we pull guava classes.

I tried to simulate this locally (create class that contains `ImmutableList`) and get:
```
[ERROR] Used undeclared dependencies found:
[ERROR]    com.google.guava:guava:jar:31.1-jre:provided
```

Now why CI passed this? :thinking:

xsalefter: Maybe because [skipped here](https://github.com/killbill/gh-actions-shared/blob/main/.github/workflows/ci.yml#L15)?

pierre: CI **did** break (the `verify` job, not the test jobs). The reason I run these checks separately (in their own `verify` job) is because they take a long time (best to do it once vs doing it for every test suite).

image

PR #1831, opened 2023-01-17
cancel subscription with NO_BILLING_PERIOD. fix endless cycle

deepcitizen:
if the CTD is later than the current date and the plan has a billing period of NO_BILLING_PERIOD then this will cause an infinite loop

### DIFF
--- a/catalog/src/test/resources/org/killbill/billing/catalog/catalogTest.xml
+++ b/catalog/src/test/resources/org/killbill/billing/catalog/catalogTest.xml
@@ -778,6 +778,35 @@
                 </recurring>
             </finalPhase>
         </plan>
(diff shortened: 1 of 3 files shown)


pierre:
Thanks for your contribution! Could you add a regression test highlighting the current issue?

deepcitizen:
done

sbrossie:
@deepcitizen Just to make sure I understand the use case:  You defined an `EVERGREEN` plan with no `Usage` or `Recurring` section so It looks like you just want a one time (fixed) price. Is this the scenario?

sbrossie (approved):

deepcitizen:
Yes

PR #1835, opened 2023-02-27
Fix for 1340

reshmabidikar:


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
```diff
@@ -469,12 +469,8 @@ public Response generateDryRunInvoice(@Nullable final InvoiceDryRunJson dryRunSu
                     Preconditions.checkArgument(dryRunSubscriptionSpec.getBillingPeriod() == null, "DryRun subscription billing period should not be set when planName is specified");
                     Preconditions.checkArgument(dryRunSubscriptionSpec.getProductCategory() == null, "DryRun subscription product category should not be set when planName is specified");
                 }
-            } else if (SubscriptionEventType.CHANGE.equals(dryRunSubscriptionSpec.getDryRunAction())) {
-                verifyNonNullOrEmpty(dryRunSubscriptionSpec.getProductName(), "DryRun subscription product category should be specified");
```
pierre, line 473: I don't think you can collapse both branches, the preconditions are a bit different (e.g. `bundleId` / `subscriptionId` must be specified for change). The bug still stands though and we need to allow `planName` (but I think it only replaces `productName`/`billingPeriod` requirements.

Worth adding a few more tests, looks simple but tricky enough 😅

sbrossie: I think we should allow to specify all 3 actions, i.e  `START_BILLING`,  `CHANGE` and `STOP_BILLING`.
Then as you (started to do), we should add the proper checks to veirfy the correct argument for each case, i.e `effectiveDate` needs to be specified for  `STOP_BILLING` (*).

So, I would suggest:
1. Add back the `STOP_BILLING` case
2.  Either understand whether we really need the `effectiveDate` for the `STOP_BILLING` (it seems like a null date should be perfectly allowed and default to `now`) OR understand why the integration test fails and propose a fix.

reshmabidikar: > Add back the STOP_BILLING case

Sorry, but I do not follow. The `STOP_BILLING` check is still there. See [this](https://github.com/reshmabidikar/killbill/blob/24692dfe91bbfc776da096ff6275609f7da921be/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L473) line. LMK if you think anything else needs to be done.

> (it seems like a null date should be perfectly allowed and default to now)

I _think_ a null `targetDate` defaults to `now`, however the `InvoiceDryRun.effectiveDate` needs to be explicitly specified. However, I will investigate this further and look more into the test failure and provide a detailed analysis.

reshmabidikar: I investigated this further and I was right. The `targetDate` defaults to the current date if not specified, the `InvoiceDryRun.effectiveDate` does not. When `InvoiceDryRun.effectiveDate` is specified, an additional billing event is returned at [this line](https://github.com/killbill/killbill/blob/b469bf01e0a6dc6da8743ef40865168f5cc9148b/invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java#L390). Thus, this results in invoice generation. However, when `InvoiceDryRun.effectiveDate` is not specified, the billing event is not returned and no invoice is generated. 

There are two ways that this can be handled:

1. Force a user to specify an `effectiveDate` when the `STOP_BILLING` action is specified as done in [this commit](https://github.com/killbill/killbill/commit/4db7ae94bed442537cf55c2aa0925d8341d1483b). This causes [this integration test](https://github.com/killbill/killbill-integration-tests/blob/ea878b772a9e2a119910ae7b89a5081a3dfe8c11/killbill-integration-tests/core/test_invoice.rb#L219) to fail [here](https://github.com/killbill/killbill-integration-tests/blob/ea878b772a9e2a119910ae7b89a5081a3dfe8c11/killbill-integration-tests/core/test_invoice.rb#L240). In order to fix this, we would need to update `requested_date` to a non-nil value [here](https://github.com/killbill/killbill-integration-tests/blob/ea878b772a9e2a119910ae7b89a5081a3dfe8c11/killbill-integration-tests/core/test_invoice.rb#L238).

2. Add logic on the Kill Bill side to default `InvoiceDryRun.effectiveDate` to the current date if not specified. This can be done within the [InvoiceResource.generateDryRun](https://github.com/reshmabidikar/killbill/blob/24692dfe91bbfc776da096ff6275609f7da921be/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L441) method.

Let me know what you think and I will make the changes accordingly.

sbrossie: This code is all about doing assertions on the input based on the various use cases, and there are lots of things 😓

1. We want to check that `subscriptionId` is specified in all cases I believe - It seems like for `ADD_ON` `START_BILLING` we also need the `bundleId`. I am unclear if this is instead of `subscriptionId`  or on top of it. Also I am unclear if this also applies for `CHANGE`. So that's one thing to verify and I believe this was the original comment from Pierre
2. For `START_BILLING` and `CHANGE`, we need the logic that allows either `planName` or {`ProductName `, `BillingPeriod`, `Category`}, which is the core of the issue we are trying to fix
3. Regarding the `InvoiceDryRun.effectiveDate`, it seems like we need it for all 3 cases, but I don't see any `verifyNonNullOrEmpty` statement. Either we need to add it or we should default to `now`. If this makes some tests fail we should fix them to match the new behavior. I believe this last case addresses the concern with `STOP_BILLING`

reshmabidikar: Thanks, @sbrossie, it is indeed more complicated than it looks 😅 Just a few comments:

> We want to check that subscriptionId is specified in all cases I believe 

`subscriptionId` is not required for `START_BILLING` (Since there is no existing subscription). See [this line](https://github.com/killbill/killbill/blob/b469bf01e0a6dc6da8743ef40865168f5cc9148b/junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java#L185). 

> It seems like for ADD_ON START_BILLING we also need the bundleId

That is correct. For `ADD_ON START_BILLING`,  the `bundleId` is required. Again, `subscriptionId` is not required since there is no existing subscription yet.

> For START_BILLING and CHANGE, we need the logic that allows either planName or {ProductName , BillingPeriod, Category

Agreed, the check for `CHANGE` was missed out earlier and is added as part of this PR

> Regarding the InvoiceDryRun.effectiveDate, it seems like we need it for all 3 cases,

I took a closer look at the code and looks like the logic is in the [SubscriptionApiBase. populateDryRunEvents](https://github.com/killbill/killbill/blob/b469bf01e0a6dc6da8743ef40865168f5cc9148b/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionApiBase.java#LL122C17-L122C38) method:
* If `InvoiceDryRun.effectiveDate` is not specified for `START_BILLING` it [defaults ](https://github.com/killbill/killbill/blob/b469bf01e0a6dc6da8743ef40865168f5cc9148b/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionApiBase.java#L159)to `now`
* If `InvoiceDryRun.effectiveDate` is not specified for `CHANGE`, it [applies the change](https://github.com/killbill/killbill/blob/b469bf01e0a6dc6da8743ef40865168f5cc9148b/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionApiBase.java#L188) as per the catalog change policy.
* If `InvoiceDryRun.effectiveDate` is not specified for `STOP_BILLING`, it [applies the change](https://github.com/killbill/killbill/blob/b469bf01e0a6dc6da8743ef40865168f5cc9148b/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionApiBase.java#L206) as per the catalog cancel policy.

So, we can either:

  * Leave it as it is (As this is not a bug as such)
  * Force the user to always specify an  `InvoiceDryRun.effectiveDate` via a `verifyNonNullOrEmpty` statement
  * Default `InvoiceDryRun.effectiveDate` to `now` if not specified.

So the following table summarises the various combinations:

||subscriptionId|bundleId|effectiveDate|planName or ProductName , BillingPeriod, Category
| -------------: | ----------------: |----------------: |----------------: |----------------: 
|START_BILLING|Not Required|Not Required|Not Required, defaults to `now`|Either One|
|CHANGE|Required|Required|Not Required (Defaults to catalog `CHANGE` policy)| Either One
|STOP_BILLING|Required|Required|Not Required (Defaults to catalog `CANCEL` policy)| Not Required

So I think we are good as far as the `subscriptionId`, `bundleId` and `planName` or `{ProductName , BillingPeriod, Category}` combination are concerned. The only thing we need to do is decide if we want to change the `effectiveDate` behavior or leave it as it is as I mentioned above. Please let me know what you think and I will make changes if we decide to change the behavior.

sbrossie: Thank you for the very detailed explanation. Couple of thoughts though:
1. `START_BILLING`: Isn't the `bundleId` required in the case of an ADD_ON?
2. `CHANGE`,  `STOP_BILLING` : Do we actually need `bundleId` for use this action? And if so, do we really need it for both ADD_ON and non ADD_ON (e.g BASE) ?


Now as far as the `effectiveDate` is concerned, i think we should have a consistent behavior across all 3 operations. Whatever is easiest as long as it is consistent. Now, adding a precondition to check it has been set, creates a limitation because we typically interpret in our apis a `null date` as being equivalent to `now`. So, my preference would be to keep this behavior for all 3 operations.

reshmabidikar: > START_BILLING: Isn't the bundleId required in the case of an ADD_ON?

Yes, that is true, I missed adding it in the table above which only takes the base subscription case into consideration.

> CHANGE, STOP_BILLING : Do we actually need bundleId for use this action

Yes, the existing subscription bundle (to which the `CHANGE`/`STOP_BILLING` action is to be applied) is retrieved based on the `bundleId` passed in. See [this line](https://github.com/killbill/killbill/blob/b469bf01e0a6dc6da8743ef40865168f5cc9148b/junction/src/main/java/org/killbill/billing/junction/plumbing/billing/DefaultInternalBillingApi.java#L198).

> So, my preference would be to keep this behavior for all 3 operations.

Got it, I will make changes to the code to default `InvoiceDryRun.effectiveDate` to `now` if not specified.

Here is the updated table:

||subscriptionId|bundleId|effectiveDate|planName or ProductName , BillingPeriod, Category
| -------------: | ----------------: |----------------: |----------------: |----------------: 
|**START_BILLING**|Not Required|Not Required for BASE, Required for ADD_ON|Not Required, defaults to `now`|Either One|
|**CHANGE**|Required|Required|Not Required ~~(Defaults to catalog `CHANGE` policy)~~ Defaults to `now`| Either One
|**STOP_BILLING**|Required|Required|Not Required ~~(Defaults to catalog `CANCEL` policy)~~ Defaults to `now`| Not Required


One new issue comes to my mind: The code [currently checks](https://github.com/reshmabidikar/killbill/blob/24692dfe91bbfc776da096ff6275609f7da921be/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L464) if the `bundleId` is specified for the `START_BILLING` action for `ADDON`. However, if the `planName` is specified (instead of `productName`, `category`, `billingPeriod` combination), there is no way to detect that this is an `ADDON` plan. So wondering how to enforce the `bundleId` check.

reshmabidikar: Made the change to default `InvoiceDryRun.effectiveDate` to `now` if not specified in https://github.com/killbill/killbill/pull/1835/commits/742c3ca1ce01ff54e517475217fbd4e0aa539dc7

reshmabidikar: Made changes to the code to default `InvoiceDryRun.effectiveDate` to `now` in https://github.com/killbill/killbill/pull/1835/commits/742c3ca1ce01ff54e517475217fbd4e0aa539dc7. However, this causes the following test failures:
* [TestUserApiChangePlan.testChangePlanOnPendingSubscription](https://github.com/killbill/killbill/actions/runs/4382732680/jobs/7672115103#step:7:3585)
* [test_invoice.test_dry_run_cancel_bp_with_multiple_policies](https://github.com/killbill/killbill/actions/runs/4382732680/jobs/7672114914#step:18:108)

The `TestUserApiChangePlan.testChangePlanOnPendingSubscription` test failure can be prevented in one of the following ways:
* by moving the [logic to default to null](https://github.com/reshmabidikar/killbill/blob/742c3ca1ce01ff54e517475217fbd4e0aa539dc7/subscription/src/main/java/org/killbill/billing/subscription/api/SubscriptionApiBase.java#L183) to the `InvoiceResource` class. 
* By removing [this](https://github.com/reshmabidikar/killbill/blob/742c3ca1ce01ff54e517475217fbd4e0aa539dc7/subscription/src/test/java/org/killbill/billing/subscription/api/user/TestUserApiChangePlan.java#L504:L510) check ( I think we should go for this option).

The `test_invoice.test_dry_run_cancel_bp_with_multiple_policies`, test failure can be eliminated by removing [this](https://github.com/killbill/killbill-integration-tests/blob/ea878b772a9e2a119910ae7b89a5081a3dfe8c11/killbill-integration-tests/core/test_invoice.rb#L141:L146) check. 

Let me know what you think.

sbrossie: I think we are converging - thanks for your patience :-)

Regarding https://github.com/killbill/killbill/commit/742c3ca1ce01ff54e517475217fbd4e0aa539dc7 we are 👍 

Regarding the test failures, if I understand correctly before by not passing a date (nor a policy) we would end up defaulting the the catalog default rules (e.g `EOT`), but now we default to `now` which changes the behavior and makes the assertions fail. Is this is the correct assessment?

So at this point:
1. Go ahead with the suggestions you proposed and let's merge this PR.
2. Let's file a GH issue to track this discussion. When I had suggested to default 'null' -> 'now' I had not thought of the default catalog rules. What I'd like this GH issue to track  is the following:
- In non-dry mode how is 'null' date being handled for all operations and what is the way to end up using the default catalog rules. It is true that if we don't pass a policy and don't pass en affective date, then probably we should use default catalog policies. So there is confusion between default catalog policies and 'now'
- In dry-run mode, do we have the same behavior as non-dry run mode?

reshmabidikar: > Regarding the test failures, if I understand correctly before by not passing a date (nor a policy) we would end up defaulting to the catalog default rules (e.g EOT), but now we default to now which changes the behavior and makes the assertions fail. Is this is the correct assessment

Yes, that is correct. 

> Go ahead with the suggestions you proposed and let's merge this PR.

I've updated the tests that failed (See 742c3ca, 649eb7d,  and 98f4bda). I've also updated the integration tests (See https://github.com/killbill/killbill-integration-tests/pull/67)

> Let's file a GH issue to track this discussion.

I've filed https://github.com/killbill/technical-support/issues/95

PR #1836, opened 2023-03-13
Update version

reshmabidikar:


profiles/killbill/src/main/resources/update-checker/killbill-server-update-list.properties
```diff
@@ -206,314 +248,284 @@
...
+0.18.7.notices           = We recommend upgrading to 0.24.0, our latest GA release.
 0.18.7.release-notes     = https://github.com/killbill/killbill/releases/tag/killbill-0.18.7
 
 # 0.18.6
 0.18.6.updates           = 0.18.22
-0.18.6.notices           = We recommend upgrading to 0.22.25, our latest GA release.
+0.18.6.notices           = We recommend upgrading to 0.24.0, our latest GA release.
 0.18.6.release-notes     = https://github.com/killbill/killbill/releases/tag/killbill-0.18.6
-
-# 0.18.5
```
pierre, line 491: You can keep these old versions (still a few folks on 0.18.x unfortunately!).

pierre (approved):
Just one comment, otherwise :shipit:

profiles/killbill/src/main/resources/update-checker/killbill-server-update-list.properties
reshmabidikar: These got deleted when I ran the `gen-updater.rb` script, I'll add these back in 👍

pierre: Looks like the call doesn't return all tags (and changing the `per_page` argument doesn't help): https://github.com/killbill/killbill/blob/b469bf01e0a6dc6da8743ef40865168f5cc9148b/bin/gen_updater.rb#LL15C26-L15C89

That's ok, we can fix it manually.

reshmabidikar: Fixed

sbrossie (approved):

Issue #1853, opened 2023-05-10
Address Spotbugs errors


xsalefter:
1. Enable `check.fail-spotbugs` in each module (do not in root `killbill` project, otherwise, it wont build
2. Find a way to fix the error
3. If (2) is either: 
    - too costly
    - almost impossible to fix
    - it is as is as designed/intended

   then exclude it to `<killbill-root-project>/spotbugs-exclude.xml` file.
4. Remove the `@SuppressFBWarnings` annotations and use `spotbugs-excude.xml` file instead.

xsalefter:
Fix order:
```
killbill-api
killbill-util
killbill-tenant       
killbill-account      
killbill-catalog      
killbill-currency     
killbill-subscription 
killbill-entitlement  
killbill-junction     
killbill-usage        
killbill-invoice      
killbill-overdue      
killbill-payment      
killbill-beatrix      
killbill-jaxrs        
killbill-profiles         
killbill-profiles-killbill 
killbill-profiles-killpay 
```

PR #1861, opened 2023-05-18
fix spotbugs in beatrix and jaxrs

xsalefter:
Fix this issue: https://github.com/killbill/killbill/issues/1853 for `beatrix` and `jaxrs` modules. Each commits represent what its attempt to fix. Some commits may need more attention:

- [remove unused local variables](https://github.com/killbill/killbill/commit/d83c13e637bab0862505282ae23fbf470e897062)
- [use DateTimeZone.UTC if account.getTimeZone() is null](https://github.com/killbill/killbill/commit/576803ed6a0b89970203c21e52c44ff6ade7a0ad)
- [remove try-catch block to NullPointerException](https://github.com/killbill/killbill/commit/09781d15977a513975945a75e31998317a1ac2e1)

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
@@ -47,7 +47,6 @@ import javax.ws.rs.core.StreamingOutput;
 import javax.ws.rs.core.UriInfo;
 
 import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
@@ -90,9 +89,6 @@ import org.killbill.billing.payment.api.PaymentTransaction;
 import org.killbill.billing.payment.api.PluginProperty;
 import org.killbill.billing.payment.api.TransactionStatus;
 import org.killbill.billing.payment.api.TransactionType;
-import org.killbill.commons.utils.Joiner;
-import org.killbill.commons.utils.Preconditions;
-import org.killbill.commons.utils.Strings;
 import org.killbill.billing.util.api.AuditLevel;
 import org.killbill.billing.util.api.AuditUserApi;
 import org.killbill.billing.util.api.CustomFieldApiException;
@@ -105,7 +101,6 @@ import org.killbill.billing.util.audit.AuditLog;
 import org.killbill.billing.util.audit.AuditLogWithHistory;
 import org.killbill.billing.util.callcontext.CallContext;
 import org.killbill.billing.util.callcontext.TenantContext;
-import org.killbill.commons.utils.collect.Iterables;
 import org.killbill.billing.util.customfield.CustomField;
 import org.killbill.billing.util.customfield.StringCustomField;
 import org.killbill.billing.util.entity.Entity;
@@ -114,6 +109,10 @@ import org.killbill.billing.util.jackson.ObjectMapper;
 import org.killbill.billing.util.tag.Tag;
 import org.killbill.billing.util.tag.TagDefinition;
 import org.killbill.clock.Clock;
+import org.killbill.commons.utils.Joiner;
+import org.killbill.commons.utils.Preconditions;
+import org.killbill.commons.utils.Strings;
+import org.killbill.commons.utils.collect.Iterables;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -489,12 +488,9 @@ public abstract class JaxRsResourceBase implements JaxrsResource {
         return toLocalDateDefaultToday(account, inputDate, context);
     }
 
-    protected LocalDate toLocalDateDefaultToday(final Account account, @Nullable final String inputDate, final TenantContext context) {
+    protected LocalDate toLocalDateDefaultToday(@Nullable final Account account, @Nullable final String inputDate, final TenantContext context) {
         // TODO Switch to cached normalized timezone when available
-        final DateTimeZone dateTimeZone = (account == null || account.getTimeZone() == null) ?
-                                          DateTimeZone.UTC :
-                                          account.getTimeZone();
-        return Objects.requireNonNullElse(toLocalDate(inputDate), clock.getToday(dateTimeZone));
+        return Objects.requireNonNullElse(toLocalDate(inputDate), clock.getToday(account == null ? null : account.getTimeZone()));
     }
 
     // API for subscription and invoice generation: keep null, the lower layers will default to now()
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
@@ -47,6 +47,7 @@ import javax.ws.rs.core.StreamingOutput;
 import javax.ws.rs.core.UriInfo;
 
 import org.joda.time.DateTime;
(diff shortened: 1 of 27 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/JaxRsResourceBase.java
```diff
@@ -490,7 +491,10 @@ protected LocalDate toLocalDateDefaultToday(final UUID accountId, @Nullable fina
 
     protected LocalDate toLocalDateDefaultToday(final Account account, @Nullable final String inputDate, final TenantContext context) {
```
pierre, line 492: Maybe mark `account` as `@Nullable`?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java
```diff
@@ -136,7 +133,6 @@ public Response recordUsage(final SubscriptionUsageRecordJson json,
         final Entitlement entitlement = entitlementApi.getEntitlementForId(json.getSubscriptionId(), false, callContext);
         if (entitlement.getEffectiveEndDate() != null) {
             final DateTime highestRecordDate = getHighestRecordDate(json.getUnitUsageRecords());
-            final Account account = accountUserApi.getAccountById(entitlement.getAccountId(), callContext);
```
pierre, line 139: I would rename `callContext` to `callContextNoAccountId` and recreate `callContext` with the `accountId` right after retrieving `entitlement` (we use this pattern in a few other places in jax-rs).

jaxrs/src/main/java/org/killbill/billing/jaxrs/util/Context.java
```diff
@@ -59,19 +59,15 @@ public CallContext createCallContextNoAccountId(final String createdBy, final St
...
-            final Tenant tenant = getTenantFromRequest(request);
-            final UUID tenantId = tenant == null ? null : tenant.getId();
-            final CallContext callContext = contextFactory.createCallContext(accountId, tenantId, createdBy, origin, userType, reason,
-                                                                             comment, getOrCreateUserToken());
-
-            populateMDCContext(callContext);
-
-            return callContext;
-        } catch (final NullPointerException e) {
-            throw new IllegalArgumentException(e.getMessage());
```
pierre, line 73: I think this changes the behavior as the NPE will be returned as 500 while it should be a 400 (user forgot the header):

https://github.com/killbill/killbill/blob/3a52a09da73c4e6d06cdb23272e5c070789ca345/jaxrs/src/main/java/org/killbill/billing/jaxrs/mappers/IllegalArgumentExceptionMapper.java#L39

pierre (changes requested):

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java
xsalefter: @pierre Do you mind explaining why? This commit is about removing this line:
```java
final Account account = accountUserApi.getAccountById(entitlement.getAccountId(), callContext)
```
Because `account` not used anymore. FYI:

1. that `account` introduced by [this commit](https://github.com/killbill/killbill/commit/5ded768eaf13082d107eec92d3fdc7c7ed716392#diff-01bacc5f459c602e5c3c753b9f99dc8a6f47c5e8cf5458e97025f451e2264c25R138) (because `account` attributes needed to create `TimeAwareContext` instance).
2. And then, in sub-sequence commit (08f9b913) the `TimeAwareContext` [get removed](https://github.com/killbill/killbill/commit/08f9b913b886b8e497b6edde02d51f4166137e69#diff-01bacc5f459c602e5c3c753b9f99dc8a6f47c5e8cf5458e97025f451e2264c25L139), but `account` still there, and have no use.

Or, do you mean that `accountUserApi.getAccountById(entitlement.getAccountId(), callContext)` should be called no matter what because it manage/changing any state? :thinking:

pierre: Actually, you don't need to retrieve the account. I am suggesting doing something like that: 

```
diff --git a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java
index 1546e86d48..c2caa820f1 100644
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java
@@ -131,9 +131,12 @@ public class UsageResource extends JaxRsResourceBase {
                 verifyNonNull(usageRecordJson.getRecordDate(), "UsageRecordJson recordDate needs to be set");
             }
         }
-        final CallContext callContext = context.createCallContextNoAccountId(createdBy, reason, comment, request);
+        final CallContext callContextNoAccountId = context.createCallContextNoAccountId(createdBy, reason, comment, request);
+
         // Verify subscription exists..
-        final Entitlement entitlement = entitlementApi.getEntitlementForId(json.getSubscriptionId(), false, callContext);
+        final Entitlement entitlement = entitlementApi.getEntitlementForId(json.getSubscriptionId(), false, callContextNoAccountId);
+        final CallContext callContext = context.createCallContextWithAccountId(entitlement.getAccountId(), createdBy, reason, comment, request);
+
         if (entitlement.getEffectiveEndDate() != null) {
             final DateTime highestRecordDate = getHighestRecordDate(json.getUnitUsageRecords());
             final Account account = accountUserApi.getAccountById(entitlement.getAccountId(), callContext);
```

In general, the context should always have the accountId populated (this is used in other layers below). But sometimes we cannot construct it directly (like here) and we need to construct it in multiple stages. Another example:
https://github.com/killbill/killbill/blob/3a52a09da73c4e6d06cdb23272e5c070789ca345/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoicePaymentResource.java#L186-L189

xsalefter: Ah, so in `UsageResource.recordUsage()` use case, the `callContext` that [passed to usageUserApi.recordRolledUpUsage()](https://github.com/killbill/killbill/blob/09781d15977a513975945a75e31998317a1ac2e1/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/UsageResource.java#LL142C22-L142C41) need to contains valid `accountId`. This is the body of `UsageResource.recordUsage()` method (validation part removed):

```java
        final CallContext callContextNoAccount = context.createCallContextNoAccountId(createdBy, reason, comment, request);
        // Verify subscription exists..
        final Entitlement entitlement = entitlementApi.getEntitlementForId(json.getSubscriptionId(), false, callContextNoAccount);
        if (entitlement.getEffectiveEndDate() != null) {
            final DateTime highestRecordDate = getHighestRecordDate(json.getUnitUsageRecords());
            if (entitlement.getEffectiveEndDate().compareTo(highestRecordDate) < 0) {
                return Response.status(Status.BAD_REQUEST).build();
            }
        }

        final SubscriptionUsageRecord record = json.toSubscriptionUsageRecord();
        // See this discussion: https://github.com/killbill/killbill/pull/1861#discussion_r1199581600
        final CallContext callContext = context.createCallContextWithAccountId(entitlement.getAccountId(), createdBy, reason, comment, request);
        usageUserApi.recordRolledUpUsage(record, callContext);
        return Response.status(Status.CREATED).build();
```

Is this correct?

pierre: Yes, that works (my suggested patch above was similar). 👍

pierre (approved):

Issue #1863, opened 2023-05-20
Missing Child invoices on the Parent account


KBbitsP:
Description : The child invoices are not linked to the Parent invoice when there are a large number of closely placed calls to create the child account invoice. Child invoices are generated but the link to the Parent invoice is missing. 

Steps to reproduce : 

* Create a parent account with the tag auto_invoice_draft
* Create child accounts, for example, 500 or 1000 accounts
* For all child accounts, set payment_delegated_to_parent as true and the tag auto_invoice_off
* Create subscription for each child account
* Generate invoices for all child accounts for a target date

Below script( python ) can be used to replicate this issue.

```
import killbill
from killbill.configuration import Configuration
from killbill.models.account import Account
from killbill.models.tenant import Tenant
from killbill.models.payment_method import PaymentMethod
from killbill.models.subscription import Subscription
from killbill.models.invoice_dry_run import InvoiceDryRun
from killbill.models.invoice_item import InvoiceItem
from killbill.models.payment_transaction import PaymentTransaction
from random import choice
from string import ascii_lowercase
import time
import datetime
from killbill.rest import ApiException

#configuration block
config = Configuration()
apiKey = '14thMayTenant2key' #Replace this with your Api Key
apiSecret= '14thMayTenant2secret'#Replace this with Api Secret
config.api_key['X-Killbill-ApiKey'] = apiKey
config.api_key['X-Killbill-ApiSecret'] =  apiSecret

#Calling all API's
api_tenant = killbill.TenantApi()
api_account = killbill.AccountApi()
api_subscription = killbill.SubscriptionApi()
api_invoice = killbill.InvoiceApi()

childAccounts = []

#Create the parent account, add tag auto_invoice_draft. 
#Calls createChildAccount function and createChildInvoice function for all child accounts
#You can set the external key per choice on lines 34,38. 
def createParentAccount():

    body = Account(name='John', external_key='random_external_key_parent12000', currency='USD', state='CA', country='USA',bill_cycle_day_local=25)

    api_account.create_account(body, 'test')

    parentAccountID = api_account.get_account_by_key('random_external_key_parent12000').account_id

    tag = ["00000000-0000-0000-0000-000000000008"]

    api_account.create_account_tags(parentAccountID, tag, 'test')

    for x in range(1000):        
        createChildAccount(parentAccountID,x)
        
    for x in childAccounts:
        createChildInvoice(x)
        #time.sleep(1)

[...]

PR #1864, opened 2023-05-20
https://github.com/killbill/killbill/issues/1863

KBbitsP:
Hi @pierre @reshmabidikar


sbrossie (commented):
@KBbitsP  Did you have a chance to run a load test prior and after the fix to check if this works?

I suspect if there is a lot of concurrency, we may end reschedule a lot of entries for the same parent account, i.e all at the same time, and then when they get processed, we may again incur a lot of concurrency leading to the same result.

sbrossie (approved):

Issue #1862, opened 2023-05-19
Incorrect invoicing with versioned catalogs and multi-phase plans


reshmabidikar:
Steps to reproduce:

1. Set the `org.killbill.subscription.align.effectiveDateForExistingSubscriptions` to true
2. Upload [Insurance-v1](https://github.com/killbill/killbill/files/11515449/Insurance-v1.zip) and [Insurance-v2](https://github.com/killbill/killbill/files/11515450/Insurance-v2.zip)
3. Set the clock to `2023-03-18`
4. Create a new account 
5. Create a subscription to the `liability-monthly-no-trial` plan - An invoice is generated for `$0`  for the duration `2023-03-18` to `2023-04-01` corresponding to the `DISCOUNT` phase as expected
6. Move the clock to `2023-04-01` - An invoice is generated for `$6.99`  for the duration `2023-04-01` to `2023-05-01` corresponding to the `EVERGREEN` phase as expected
7. Move the clock to `2023-05-01` - An invoice is generated for `$6.99`  for the duration `2023-05-01` to `2023-06-01` corresponding to the `EVERGREEN` phase as expected
8. Move the clock to `2023-06-01` - An invoice is generated for `$0`  for the duration `2023-06-01` to `2023-07-01` . **Expected behavior** - The invoice should be generated for `$10.5`  as per `v2` as the `effectiveDateForExistingSubscriptions` is now passed.
9. Move the clock to `2023-07-01` -  An invoice is generated for `$0`  for the duration `2023-07-01` to `2023-08-01` . **Expected behavior** - The invoice should be generated for `$10.5`  as per `v2` as the `effectiveDateForExistingSubscriptions` is now passed.

pierre:
Difference between catalogs:

```
--- Insurance-v1.xml	2023-05-19 10:23:12
+++ Insurance-v2.xml	2023-05-19 10:23:16
@@ -17,7 +17,7 @@

 <catalog xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="CatalogSchema.xsd ">

-    <effectiveDate>2022-05-18T00:00:00+00:00</effectiveDate>
+    <effectiveDate>2023-05-18T00:00:00+00:00</effectiveDate>
     <catalogName>Insurance</catalogName>

     <recurringBillingMode>IN_ADVANCE</recurringBillingMode>
@@ -70,6 +70,7 @@

     <plans>
[...]

PR #1868, opened 2023-05-25
Maint for 0.22.x fix 1862

reshmabidikar:
Test and fix for https://github.com/killbill/killbill/issues/1862


sbrossie:
We updated the original fix on 0.24, so closing the one.

PR #1870, opened 2023-05-25
Test 1867

reshmabidikar:


sbrossie:
Closing as we modified the fix in https://github.com/killbill/killbill/pull/1872

Re-opened https://github.com/killbill/killbill/pull/1873 instead

PR #1872, opened 2023-05-26
subscription: Rework billing events computation. See #1862

sbrossie:
This undo the fix introduced in 42e9b205900d.

Also note that the diff in the test `TestSubscriptionBillingEvents.java` just reverts it exactly to what it currently is in `master`, so the code introduces no known behavior change aside from fixing the issue #1862

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
@@ -35,7 +35,6 @@ import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
 import org.joda.time.Period;
-import org.killbill.billing.ErrorCode;
 import org.killbill.billing.callcontext.InternalTenantContext;
 import org.killbill.billing.catalog.api.BillingActionPolicy;
 import org.killbill.billing.catalog.api.BillingAlignment;
@@ -470,7 +469,6 @@ public class DefaultSubscriptionBase extends EntityBase implements SubscriptionB
         return getPrevValue(false);
     }
 
-
     private Integer getPrevValue(final boolean bcd) {
 
         final SubscriptionBaseTransitionDataIterator it = new SubscriptionBaseTransitionDataIterator(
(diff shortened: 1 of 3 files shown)


subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -686,26 +683,18 @@ public List<SubscriptionBillingEvent> getSubscriptionBillingEvents(final Version
...
-                        } catch (final CatalogApiException e) {
-                            if (e.getCode() == ErrorCode.CAT_NO_SUCH_PLAN.getCode()) {
-                                // Retired plan
-                                continue;
-                            } else {
-                                throw e;
-                            }
-                        }
+                        // If we are moving to a new Plan, we use the latest active catalog version at the time this operation took place.
+                        final DateTime billingTransitionEffectiveDate = isPhaseEvent ? lastPlanTransition.getEffectiveTransitionTime() : billingTransition.getEffectiveDate();
```
reshmabidikar, line 687: Is this required? Or could we just set `billingTransitionEffectiveDate=billingTransition.getEffectiveDate()` every time? Not saying this is wrong, just trying to understand the code better  ;-)

reshmabidikar (approved):

subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
sbrossie: It is not easy to understand, I added this fix after I discovered that one of our test, i.e `TestCatalogNewEntries#testNewPlanWithRetiredOldPlan` was failing:

Given 2 catalog versions, V1 effDt = `2011-01-01` and V2 `effDt 2015-12-01`: V1 has a plan `pistol-monthly` with a 30 days TRIAL  but such plan got deprecated in V2.

If we start a subscription on `2015-11-01` with plan `pistol-monthly`: The code above processes the PHASE event on `2015-12-01` (`2015-11-01` + `30` days), and unless we look for the Plan in the original catalog version V1, it will fail because such plan got deprecated in V2, hence the computation of the `billingTransitionEffectiveDate`.

Issue #1869, opened 2023-05-25
Plan#effectiveDateForExistingSubscriptions does not work for weekly plans


reshmabidikar:
Steps to reproduce:

1. Set the `org.killbill.subscription.align.effectiveDateForExistingSubscriptions` to true
2. Upload [Insurance-v1](https://github.com/killbill/killbill/files/11561167/Insurance-v1.zip) and [Insurance-v2](https://github.com/killbill/killbill/files/11561168/Insurance-v2.zip)
3. Set the clock to `2023-05-05`
4. Create a new account 
5. Create a subscription to the `liability-monthly-with-trial` plan - An invoice is generated for `$1.5`  for the duration `2023-05-05` to `2023-05-12` corresponding to the `TRIAL` phase as expected
6. Move the clock to `2023-05-12` - An invoice is generated for `$4.5`  for the duration `2023-05-12` to `2023-05-19` corresponding to the `EVERGREEN` phase as expected
7. Move the clock to `2023-05-19` - An invoice is generated for `$3.48`  for the duration `2023-05-19` to `2023-05-25`.  **Expected behavior** - The invoice should be generated for `$4.5` for the duration `2023-05-19` to `2023-05-26` since  `org.killbill.subscription.align.effectiveDateForExistingSubscriptions=true` is set.

reshmabidikar:
Addressed in https://github.com/killbill/killbill/pull/1877 and https://github.com/killbill/killbill/pull/1891. Closing this issue.

PR #1877, opened 2023-06-07
Initial fix for 1869

reshmabidikar:
Initial fix for #1869

### DIFF
--- a/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
+++ b/subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
@@ -744,7 +744,7 @@ public class DefaultSubscriptionBase extends EntityBase implements SubscriptionB
         }
 
         final BillingPeriod billingPeriod = curPlanPhase.getRecurring() != null ? curPlanPhase.getRecurring().getBillingPeriod() : BillingPeriod.NO_BILLING_PERIOD;
-        final LocalDate resultingLocalDate = BillCycleDayCalculator.alignProposedNextBillCycleDate(originalTransitionDate, bcd, billingPeriod, context);
+        final LocalDate resultingLocalDate = BillCycleDayCalculator.alignProposedNextBillCycleDate(originalTransitionDate, lastTransitionDate, bcd, billingPeriod, context);
         final DateTime candidateResult = context.toUTCDateTime(resultingLocalDate);
         return candidateResult;
     }
--- a/util/src/main/java/org/killbill/billing/util/bcd/BillCycleDayCalculator.java
+++ b/util/src/main/java/org/killbill/billing/util/bcd/BillCycleDayCalculator.java
@@ -76,22 +76,27 @@ public abstract class BillCycleDayCalculator {
     }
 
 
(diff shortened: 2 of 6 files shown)


subscription/src/main/java/org/killbill/billing/subscription/api/user/DefaultSubscriptionBase.java
```diff
@@ -744,7 +744,7 @@ private DateTime alignToNextBCDIfRequired(final Plan curPlan, final PlanPhase cu
         }
 
         final BillingPeriod billingPeriod = curPlanPhase.getRecurring() != null ? curPlanPhase.getRecurring().getBillingPeriod() : BillingPeriod.NO_BILLING_PERIOD;
-        final LocalDate resultingLocalDate = BillCycleDayCalculator.alignProposedNextBillCycleDate(originalTransitionDate, bcd, billingPeriod, context);
+        final LocalDate resultingLocalDate = BillCycleDayCalculator.alignProposedNextBillCycleDate(originalTransitionDate, lastTransitionDate, bcd, billingPeriod, context);
```
sbrossie, line 747: Unrelated to your change, but the name are so confusing. Would it make sense to rename
* `originalTransitionDate` -> `curTransitionDate` # Really the transition time associated with the `Plan#EffectiveDateForExistingSubscriptions`
* `lastTransitionDate` -> `prevTransitionDate`

And perhaps swap the 2 in the method signature?

util/src/main/java/org/killbill/billing/util/bcd/BillCycleDayCalculator.java
```diff
@@ -76,22 +76,27 @@ public static LocalDate alignProposedBillCycleDate(final DateTime proposedDate,
     }
 
 
-    public static LocalDate alignProposedNextBillCycleDate(final LocalDate proposedDate, final int billingCycleDay, final BillingPeriod billingPeriod) {
+    public static LocalDate alignProposedNextBillCycleDate(final LocalDate proposedDate, final LocalDate originalDate, final int billingCycleDay, final BillingPeriod billingPeriod) {
```
sbrossie, line 79: Same thing here - rename and swap:
* `originalDate` -> `prevTransitionDate`
* `proposedDate` -> `curTransitionDate`
*

util/src/test/java/org/killbill/billing/util/bcd/TestBillCycleDayCalculator.java
```diff
@@ -172,7 +172,7 @@ public void testAlignProposedNextBillCycleDate1() {
         // BCD > proposed day of the month
```
sbrossie, line 172: Could we add a few unit tests here to cover for non monthly billing periods - at least the one matching the scenario?

sbrossie:
@reshmabidikar I am merging your PR as-is - optionally we could have another small one to clean things up and address the comments.

sbrossie (approved):

util/src/main/java/org/killbill/billing/util/bcd/BillCycleDayCalculator.java
sbrossie: What about we also change the method name to `alignProposedNextBillCycleDate` -> `alignToNextBillCycleDate`. It would take as a first argument `prevTransitionDate`, and then `curTransitionDate` with a comment /* date to be aligned */  ?

PR #1888, opened 2023-06-24
Global lock error resiliency

sbrossie:
Add resiliency in the system when Global Account lock could not be acquired in an async flow, either as the system processes a bus event, or a notification. see https://github.com/killbill/killbill/issues/1889

Most of the implementation relies on the wrapping of our original handlers so we can handle the `QueueRetryException` and reschedule based on default schedule. The one exception is the main invoice processing path where we leverage the existing deduplication of events already existing inside the invoice code. 

We already had some logic in invoice to handle such issues but part of if was behind a feature flag. In this change https://github.com/killbill/killbill/pull/1888/commits/19878ee0d6bdcf94d8b3ed1eca7a6c4e1ad7e918, we remove the feature flag to make it the default behavior.

In the payment code, there are 2 places where we need to handle such issues:
* In the `PaymentBusEventHandler`, the bus handler is now wrapped in a `SubscriberQueueHandler` similarly to what we did with our invoice handlers. This has been implemented in https://github.com/killbill/killbill/pull/1888/commits/d3d8635dc00372b97f743103603a9219a4107ccd and tested in https://github.com/killbill/killbill/pull/1888/commits/89cb8e41d7a20db197a0723a9c9327981d9e3d12
* In the `AUTO_PAY_OFF` removal logic, the notification queue handler is also wrapped in a new retry queue. This was implemented and tested in https://github.com/killbill/killbill/pull/1888/commits/e1b1e6c9c4b288c8cd148e922841f9c5fe6ad0c7

The overdue code also needs to be fixed and will be part of a subsequent PR if there is agreement on the proposed strategy.


beatrix/pom.xml
```diff
@@ -295,6 +295,10 @@
             <groupId>org.kill-bill.commons</groupId>
             <artifactId>killbill-jdbi</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.kill-bill.commons</groupId>
+            <artifactId>killbill-locker</artifactId>
```
pierre, line 300: Technically `<scope>test</scope>` although this might confuse Maven (since it's also required via runtime transitive dependencies).

payment/src/main/java/org/killbill/billing/payment/bus/PaymentBusEventHandler.java
```diff
@@ -34,42 +34,112 @@
...
+                                                                              internalContext);
+                } catch (final AccountApiException e) {
+                    log.warn("Failed to process invoice payment", e);
+                } catch (final PaymentApiException e) {
+                    // Look for a LockFailedException cause to decide if we should reschedule
+                    // (It should be the first cause of the PaymentApiException, but we support additional levels mostly for testing)
+                    Throwable curCause = e.getCause();
+                    while (curCause != null) {
+                        if (curCause instanceof LockFailedException) {
+                            throw new QueueRetryException();
```
pierre, line 131: Should we make the retry schedule configurable (shared config for all of these), instead of relying on `DEFAULT_RETRY_SCHEDULE`? To your point that we may eventually timeout, we probably want to have a default with more retries (e.g. `1m,1m,1m,5m,5m,5m,1h,1h,1h,6h,6h,6h,24h,24h,24h`).

pierre (approved):

payment/src/main/java/org/killbill/billing/payment/bus/PaymentBusEventHandler.java
sbrossie: Updated the implementation with per module config as discussed: https://github.com/killbill/killbill/pull/1888/commits/b2f74ab2a35e0145f64a5f2275c0a629ce3ab96f

invoice/src/main/java/org/killbill/billing/invoice/InvoiceDispatcher.java
```diff
@@ -346,6 +347,21 @@ public List<Invoice> processAccount(final boolean isApiCall,
...
+
+    private boolean rescheduleProcessAccount(final UUID accountId, final InternalCallContext context) {
+        // Anything below 1sec, we would ignore
+        final TimeSpan timeSpan = invoiceConfig.getRescheduleIntervalOnLock(context);
+        final int delaySec = (int) TimeUnit.SECONDS.convert(timeSpan.getMillis(), TimeUnit.MILLISECONDS);
+        if (delaySec <= 0) {
+            return false;
+        }
+        final DateTime nextRescheduleDt = clock.getUTCNow().plusSeconds(delaySec);
+        log.info("Rescheduling invoice call at time {}", nextRescheduleDt);
```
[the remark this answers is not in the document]
sbrossie: Yeah I don't know: It is an expected scenario as we have seen it multiple times, and if our retry logic works well ansd invoice gets generated, does it make sense to warn?

beatrix/pom.xml
sbrossie: Yeah, I'd rather not touch it :-)

payment/src/main/java/org/killbill/billing/payment/bus/PaymentBusEventHandler.java
pierre: > providing an empty list does not seem handled

Yup, agreed that empty list should result in no retry. That being said, I don't see why we would want to disable these specific retries.

pierre (approved):

PR #1890, opened 2023-06-28
Fix ts 103

reshmabidikar:
Fix for custom keywords/key-value patterns for obfuscation.  In order to use this feature, the following additional system properties need to be set:

- killbill.server.log.obfuscate.keywords (defaults to `accountnumber,authenticationdata,banknumber,bic,cardvalidationnum,cavv,ccFirstName,ccLastName,ccNumber,ccTrackData,ccVerificationValue,ccvv,cvNumber,cvc,cvv,email,iban,name,number,password,xid,externalKey`)
- killbill.server.log.obfuscate.patterns (defaults to `\\s*=\\s*'([^']+)',\\s*=\\s*"([^"]+)"`)
- killbill.server.log.obfuscate.patterns.separator (defaults to `,`)

### DIFF
--- a/profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
+++ b/profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
@@ -43,8 +43,8 @@ public class TestPatternObfuscator {
     public void testLoadKeywords() {
         //without setting the killbill.server.log.obfuscate.patterns property
         System.clearProperty("killbill.server.log.obfuscate.keywords");
-        Collection<String> keywords = PatternObfuscator.loadKeywords();
-        Assert.assertEquals(22, keywords.size());
+        Collection<String> keywords = obfuscator.loadKeywords();
+        Assert.assertEquals(keywords.size(), 22);
         Collection<String> actual = List.of(
                 "accountnumber",
                 "authenticationdata",
@@ -72,8 +72,8 @@ public class TestPatternObfuscator {
 
         //after setting the the killbill.server.log.obfuscate.patterns property
         System.setProperty("killbill.server.log.obfuscate.keywords", "accountnumber,authenticationdata");
-        keywords = PatternObfuscator.loadKeywords();
-        Assert.assertEquals(2, keywords.size());
+        keywords = obfuscator.loadKeywords();
+        Assert.assertEquals(keywords.size(), 2);
         actual = List.of(
                 "accountnumber",
                 "authenticationdata");
@@ -83,24 +83,45 @@ public class TestPatternObfuscator {
 
     @Test
     public void testLoadPatterns() {
-        //without setting the killbill.server.log.obfuscate.patterns property
+        //without setting any properties
         System.clearProperty("killbill.server.log.obfuscate.patterns");
-        Collection<String> patterns = PatternObfuscator.loadPatterns();
-        Assert.assertEquals(2, patterns.size());
+        System.clearProperty("killbill.server.log.obfuscate.patterns.separator");
+        Collection<String> patterns = obfuscator.loadPatterns();
+        Assert.assertEquals(patterns.size(), 2);
         Collection<String> actual = List.of("\\s*=\\s*'([^']+)'",
                                             "\\s*=\\s*\"([^\"]+)\"");
         Assert.assertEquals(actual, patterns);
 
-        // after setting the killbill.server.log.obfuscate.patterns property
-        System.setProperty("killbill.server.log.obfuscate.patterns.separator", "#");
-        System.setProperty("killbill.server.log.obfuscate.patterns", "\\s*=\\s*'([^']+)'#\\s*=\\s*\"([^\"]+)#\\s*=\\s*([^ '\",{}]+)#\\s*:\\s*'([^'\",{}]+)'");
-        patterns = PatternObfuscator.loadPatterns();
-        Assert.assertEquals(4, patterns.size());
+        //without setting the killbill.server.log.obfuscate.patterns.separator property
+        System.setProperty("killbill.server.log.obfuscate.patterns", "\\s*=\\s*'([^']+)',\\s*=\\s*\"([^\"]+)\"");
+        patterns = obfuscator.loadPatterns();
+        Assert.assertEquals(patterns.size(), 2);
+        actual = List.of("\\s*=\\s*'([^']+)'",
+                         "\\s*=\\s*\"([^\"]+)\"");
+        Assert.assertEquals(actual, patterns);
+
+        // after setting the killbill.server.log.obfuscate.patterns property and killbill.server.log.obfuscate.patterns.separator property
+        System.setProperty("killbill.server.log.obfuscate.patterns.separator", "~");
+        System.setProperty("killbill.server.log.obfuscate.patterns", "\\s*=\\s*'([^']+)'~\\s*=\\s*\"([^\"]+)~\\s*=\\s*([^ '\",{}]+)~\\s*:\\s*'([^'\",{}]+)'");
+        patterns = obfuscator.loadPatterns();
+        Assert.assertEquals(patterns.size(), 4);
         actual = List.of("\\s*=\\s*'([^']+)'",
                          "\\s*=\\s*\"([^\"]+)",
                          "\\s*=\\s*([^ '\",{}]+)",
                          "\\s*:\\s*'([^'\",{}]+)'");
         Assert.assertEquals(actual, patterns);
+    }
+
+    @Test
+    public void testLoadPatternSeparator() {
+        //without setting the killbill.server.log.obfuscate.patterns.separator property
+        System.clearProperty("killbill.server.log.obfuscate.patterns.separator");
+        String patternSeparator = obfuscator.loadPatternSeparator();
+        Assert.assertEquals(patternSeparator, ",");
+        // after setting the killbill.server.log.obfuscate.patterns.separator property
+        System.setProperty("killbill.server.log.obfuscate.patterns.separator", "#");
+        patternSeparator = obfuscator.loadPatternSeparator();
+        Assert.assertEquals(patternSeparator, "#");
     }
 
     @Test(groups = "fast")
--- a/profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
+++ b/profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
@@ -18,26 +18,32 @@
 package org.killbill.billing.server.log.obfuscators;
 
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.List;
-import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
-import org.killbill.billing.server.log.ServerTestSuiteNoDB;
 import org.mockito.Mockito;
 import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
 import ch.qos.logback.classic.spi.ILoggingEvent;
 
-public class TestPatternObfuscator extends ServerTestSuiteNoDB {
+public class TestPatternObfuscator {
 
-    private final PatternObfuscator obfuscator = new PatternObfuscator();
+    private PatternObfuscator obfuscator;
+
+    @BeforeClass(groups = "fast")
+    public void beforeClass() throws Exception {
+        System.setProperty("killbill.server.log.obfuscate.keywords", "req.requestURI,accountnumber,authenticationdata,banknumber,bic,cardvalidationnum,cavv,ccFirstName,ccLastName,ccNumber,ccTrackData,ccVerificationValue,ccvv,cvNumber,cvc,cvv,email,iban,name,number,password,xid,cell_phone");
+        System.setProperty("killbill.server.log.obfuscate.patterns.separator", "#");
+        System.setProperty("killbill.server.log.obfuscate.patterns", "\\s*=\\s*'([^']+)'#\\s*=\\s*\\\"([^\\\"]+)#\\s*=\\s*([^ '\\\",]+)#\\s*:\\s*'([^'\\\",{}]+)'");
+        obfuscator = new PatternObfuscator();
+    }
 
     @Test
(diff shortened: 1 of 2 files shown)


reshmabidikar:
I'd like your inputs on a few issues @sbrossie:

- I have written some additional tests [here](https://github.com/reshmabidikar/killbill/blob/2968faed64ea2c3e810d85eaedae963a9bd51fb9/profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java#L447:L470). However, these are currently disabled. Since the keywords/key-value patterns are passed as system properties, the corresponding properties need to be set for the tests to pass.  I cannot set them in the `beforeMethod`/`beforeClass` methods as it would be too late in the sequence, by then the properties are already set [here](https://github.com/reshmabidikar/killbill/blob/2968faed64ea2c3e810d85eaedae963a9bd51fb9/profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/PatternObfuscator.java#L40:L41). Any thoughts as to how these properties can be set before tests are run so that tests can be enabled?
- I have introduced a property for the pattern separator called `killbill.server.log.obfuscate.patterns.separator` as discussed on slack. I could not find a good way of escaping the comma in the regular expression pattern. Please let me know if you have any better suggestions.

sbrossie:
> I have written some additional tests [here](https://github.com/reshmabidikar/killbill/blob/2968faed64ea2c3e810d85eaedae963a9bd51fb9/profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java#L447:L470). However, these are currently disabled. Since the keywords/key-value patterns are passed as system properties, the corresponding properties need to be set for the tests to pass. I cannot set them in the beforeMethod/beforeClass methods as it would be too late in the sequence, by then the properties are already set [here](https://github.com/reshmabidikar/killbill/blob/2968faed64ea2c3e810d85eaedae963a9bd51fb9/profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/PatternObfuscator.java#L40:L41). Any thoughts as to how these properties can be set before tests are run so that tests can be enabled?

@reshmabidikar Right now, the logger system gets initialized very early as we load the base class `GuicyKillbillTestSuite` and run the static [Logger initializer](https://github.com/killbill/killbill/blob/master/util/src/test/java/org/killbill/billing/GuicyKillbillTestSuite.java#L74). I would suggest 2 things - I have not tried them though:
1. Make your test `TestPatternObfuscator` **not** `extends ServerTestSuiteNoDB` - it breaks our KB pattern but I suspect for this type of test extending the base class may not be relevant. 
2. Move away from the `static` initialization of all your variables `KILLBILL_OBFUSCATE_*` inside `PatternObfuscator` and move them in the CTOR using `final` but non `static` variable.

With this, I am hoping we can use ` @BeforeClass` to set those system properties. To be tried...

reshmabidikar:
>With this, I am hoping we can use  @BeforeClass to set those system properties. To be tried...

Thanks, @sbrossie, this approach worked, the tests are now enabled. 

I've also added some additional tests for the `killbill.server.log.obfuscate.patterns.separator property` property.

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
```diff
@@ -17,16 +17,112 @@
...
+
 import org.mockito.Mockito;
 import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
 import ch.qos.logback.classic.spi.ILoggingEvent;
 
-public class TestPatternObfuscator extends ServerTestSuiteNoDB {
+public class TestPatternObfuscator {
```
sbrossie, line 30: Could you add the following comments so we don't forget and don't try to 'fix it':

1. On top of the class: `// Test does not extend base class on purpose so as to not pull the static Logger that would be initialized too early`
2.  `public class TestPatternObfuscator { /* extends ServerTestSuiteNoDB */`

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
```diff
@@ -17,16 +17,112 @@
...
+
+    private PatternObfuscator obfuscator;
+
+    @BeforeClass(groups = "fast")
```
sbrossie, line 34: Should we have an `@AfterClass` to remove the system properties for cleanliness?

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
```diff
@@ -17,16 +17,112 @@
...
+                "email",
+                "iban",
+                "name",
+                "number",
+                "password",
+                "xid");
+        Assert.assertEquals(actual, keywords);
+
+        //after setting the the killbill.server.log.obfuscate.patterns property
+        System.setProperty("killbill.server.log.obfuscate.keywords", "accountnumber,authenticationdata");
```
sbrossie, line 74: Actually, should we even have an `@AfterMethod` to clean up system properties?

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
```diff
@@ -17,16 +17,112 @@
...
+
+    @BeforeClass(groups = "fast")
+    public void beforeClass() throws Exception {
+        System.setProperty("killbill.server.log.obfuscate.keywords", "req.requestURI,accountnumber,authenticationdata,banknumber,bic,cardvalidationnum,cavv,ccFirstName,ccLastName,ccNumber,ccTrackData,ccVerificationValue,ccvv,cvNumber,cvc,cvv,email,iban,name,number,password,xid,cell_phone");
+        System.setProperty("killbill.server.log.obfuscate.patterns.separator", "#");
+        System.setProperty("killbill.server.log.obfuscate.patterns", "\\s*=\\s*'([^']+)'#\\s*=\\s*\\\"([^\\\"]+)#\\s*=\\s*([^ '\\\",]+)#\\s*:\\s*'([^'\\\",{}]+)'");
+        obfuscator = new PatternObfuscator();
+    }
+
+    @Test
```
sbrossie, line 42: Can you add the `groups = "fast"` for all tests?

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
```diff
@@ -17,16 +17,112 @@
...
 import ch.qos.logback.classic.spi.ILoggingEvent;
 
-public class TestPatternObfuscator extends ServerTestSuiteNoDB {
+public class TestPatternObfuscator {
+
+    private PatternObfuscator obfuscator;
+
+    @BeforeClass(groups = "fast")
+    public void beforeClass() throws Exception {
+        System.setProperty("killbill.server.log.obfuscate.keywords", "req.requestURI,accountnumber,authenticationdata,banknumber,bic,cardvalidationnum,cavv,ccFirstName,ccLastName,ccNumber,ccTrackData,ccVerificationValue,ccvv,cvNumber,cvc,cvv,email,iban,name,number,password,xid,cell_phone");
```
sbrossie, line 36: I would suggest to only have the shared properties for all tests here: If those are test dependent, then they should probably be set inside the test itself, and cleanup afterwards.

sbrossie (commented):
Looks great, left a few minor comments.

Also, can you do a pass, and remove any customer (perhaps sensitive) provided data, e.g `spa-ico-plugin-cb` ?

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
reshmabidikar: The System properties need to be set before invoking the `PatternObfuscator` constructor as the constructor then loads the keywords/patterns [here](https://github.com/reshmabidikar/killbill/blob/916d4a2454a2b36084d522f69289c0493651adbb/profiles/killbill/src/main/java/org/killbill/billing/server/log/obfuscators/PatternObfuscator.java#L53). If we set the properties in the test, we would need to create a new instance of the obfuscator in the test after the properties are set. 

So, the test would need to be modified as follows:
```
    public void testJSONWithCustomKeyWord() throws Exception {
        System.setProperty("killbill.server.log.obfuscate.keywords","accountnumber,authenticationdata,banknumber,bic,cardvalidationnum,cavv,ccFirstName,ccLastName,ccNumber,ccTrackData,ccVerificationValue,ccvv,cvNumber,cvc,cvv,email,iban,name,number,password,xid,cell_phone");
        PatternObfuscator obfuscator = new PatternObfuscator();
        verify(obfuscator,"{\n" +
               "  \"card\": {\n" +
               "    \"id\": \"card_483etw4er9fg4vF3sQdrt3FG\",\n" +
               "    \"object\": \"card\",\n" +
               "    \"banknumber\": 4111111111111111,\n" +
               "    \"cvv\" : 111,\n" +
               "    \"cvv\": 111,\n" +
               "    \"cvv\": \"111\",\n" +
               "    \"data\": {\"cvv\" : 111 },\n" +
               "    \"last4\": \"0000\",\n" +
               "    \"brand\": \"Visa\",\n" +
               "    \"funding\": \"credit\",\n" +
               "    \"exp_month\": 6,\n" +
               "    \"exp_year\": 2019,\n" +
               "    \"fingerprint\": \"HOh74kZU387WlUvy\",\n" +
               "    \"country\": \"US\",\n" +
               "    \"name\": \"Bob Smith\",\n" +
               "    \"cell_phone\": \"212-421-1278\",\n" +
               "    \"address_line1\": null,\n" +
               "    \"address_line2\": null,\n" +
               "    \"address_city\": null,\n" +
               "    \"address_state\": null,\n" +
               "    \"address_zip\": null,\n" +
               "    \"address_country\": null,\n" +
               "    \"dynamic_last4\": \"4242\",\n" +
               "    \"customer\": null,\n" +
               "    \"type\": \"Visa\"}\n" +
               "}",
               "{\n" +
               "  \"card\": {\n" +
               "    \"id\": \"card_483etw4er9fg4vF3sQdrt3FG\",\n" +
               "    \"object\": \"card\",\n" +
               "    \"banknumber\": ****************,\n" +
               "    \"cvv\" : ***,\n" +
               "    \"cvv\": ***,\n" +
               "    \"cvv\": *****,\n" +
               "    \"data\": {\"cvv\" : ****},\n" +
               "    \"last4\": \"0000\",\n" +
               "    \"brand\": \"Visa\",\n" +
               "    \"funding\": \"credit\",\n" +
               "    \"exp_month\": 6,\n" +
               "    \"exp_year\": 2019,\n" +
               "    \"fingerprint\": \"HOh74kZU387WlUvy\",\n" +
               "    \"country\": \"US\",\n" +
               "    \"name\": ***********,\n" +
               "    \"cell_phone\": **************,\n" +
               "    \"address_line1\": null,\n" +
               "    \"address_line2\": null,\n" +
               "    \"address_city\": null,\n" +
               "    \"address_state\": null,\n" +
               "    \"address_zip\": null,\n" +
               "    \"address_country\": null,\n" +
               "    \"dynamic_last4\": \"4242\",\n" +
               "    \"customer\": null,\n" +
               "    \"type\": \"Visa\"}\n" +
               "}");

    }
```
If you agree, I'll make this change for all the tests, let me know what you think.

sbrossie: If you look at your existing tests, they already set some system properties within the test. For example in `testLoadKeywords`, I see:

```
    @Test
    public void testLoadKeywords() {
        //without setting the killbill.server.log.obfuscate.patterns property
        System.clearProperty("killbill.server.log.obfuscate.keywords");
        Collection<String> keywords = obfuscator.loadKeywords();

        ...

        System.setProperty("killbill.server.log.obfuscate.keywords", "accountnumber,authenticationdata");
        keywords = obfuscator.loadKeywords();
        Assert.assertEquals(keywords.size(), 2);
```


Anyhow, it's really your choice if you want to reuse the same `obfuscator` instance through the tests or create a new one each time. All I am asking is 1/ consistency across the tests, 2/ proper logic to make sure we clean things up after each tests to avoid weird behavior on subsequent tests and ideally 3/ set the minimum we need.

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
reshmabidikar: Done

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
reshmabidikar: Not required now as I've removed the `@BeforeClass` and added an `@AfterMethod` method that clears the system properties.

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
reshmabidikar: Done

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
reshmabidikar: Done

profiles/killbill/src/test/java/org/killbill/billing/server/log/obfuscators/TestPatternObfuscator.java
reshmabidikar: The `testLoadKeywords` test is different from the other tests in that it loads the keywords within the test and is not concerned about when the `PatternObfuscator` instance is created.

I've anyhow modified the tests to set the system property within the test and create an `obfuscator` instance within the test after setting the property.  I've also added an `@AfterMethod` to clear the properties after each test, so hopefully this is good now.

sbrossie (approved):

Issue #1889, opened 2023-06-28
Add resiliency around global account lock


sbrossie:
In all the async paths where Kb tries to grab the global Account lock, we should have some rescheduling logic for when the operation fails. This includes:
* Invoice bus handlers
* Payment bus handlers
* Overdue bus handlers

The following issues would be addressed by this change:
* https://github.com/killbill/killbill/issues/1863
* https://github.com/killbill/killbill/issues/1555

PR #1894, opened 2023-07-11
Global lock error resiliency

sbrossie:
This is a continuation of the work started in https://github.com/killbill/killbill/pull/1888

This PR adds the retry logic inside the overdue module when we receive a `LockFailedException`. This should conclude all the existing paths where we grab the global lock in Bill, and therefore fully address https://github.com/killbill/killbill/issues/1889

* https://github.com/killbill/killbill/pull/1894/commits/42109e4407bc51333644a9076617b03fb8b30509 adds a test to verify we correctly handle an empty config, i.e we default to not having any retry schedule. No behavior change, just a test.
* https://github.com/killbill/killbill/pull/1894/commits/8501c2a07b7c693305e51c23a75b1f453288b834 is the retry implementation for the overdue module. It adds a new `OverdueConfig` to allow configuring the retry schedule.
* https://github.com/killbill/killbill/pull/1894/commits/13d8209b51e37df6cc4a19de33e2e6c492e96369 simplifies the configuration to make it unique across modules
* https://github.com/killbill/killbill/pull/1894/commits/1cbeeedbcb31ae150261fc23f0295b6f35966d59 adds Beatrix test to verify the global lock schedule is correctly implemented for overdue.

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/config/definition/KillbillConfig.java
+++ b/util/src/main/java/org/killbill/billing/util/config/definition/KillbillConfig.java
@@ -16,9 +16,28 @@
  */
 package org.killbill.billing.util.config.definition;
 
+import java.util.List;
+
+import org.killbill.billing.callcontext.InternalTenantContext;
+import org.skife.config.Config;
+import org.skife.config.Default;
+import org.skife.config.Description;
+import org.skife.config.Param;
+import org.skife.config.TimeSpan;
+
 /*
  * Marker interface for killbill config files
  */
 public interface KillbillConfig {
+
+    @Config("org.killbill.rescheduleIntervalOnLock")
+    @Default("30s, 1m, 1m, 3m, 3m, 10m")
+    @Description("Tme delay to reschedule an invoice run when lock is held")
+    List<TimeSpan> getRescheduleIntervalOnLock();
+
+    @Config("org.killbill.rescheduleIntervalOnLock")
+    @Default("30s, 1m, 1m, 3m, 3m, 10m")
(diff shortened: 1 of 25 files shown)


pierre (approved):

util/src/main/java/org/killbill/billing/util/config/definition/KillbillConfig.java
```diff
@@ -16,9 +16,28 @@
...
+
 /*
  * Marker interface for killbill config files
  */
 public interface KillbillConfig {
 
+    @Config("org.killbill.rescheduleIntervalOnLock")
+    @Default("30s, 1m, 1m, 3m, 3m, 10m")
+    @Description("Tme delay to reschedule an invoice run when lock is held")
+    List<TimeSpan> getRescheduleIntervalOnLock();
```
pierre, line 36: NACK - `KillbillConfig` is the marker interface for all of our configs. These configs won't apply to `RbacConfig` or `EhCacheConfig` for instance.

You could introduce an intermediate interface though (`LockAwareConfig extends KillbillConfig`).

pierre (changes requested):

util/src/main/java/org/killbill/billing/util/config/definition/KillbillConfig.java
sbrossie: Ok, fair enough, that was my first thought as well -  but then I had convinced myself differently...

Fixed in https://github.com/killbill/killbill/pull/1894/commits/6f37abe76378cba90c91c3695d04411bed84d1b9

pierre (approved):

PR #1935, opened 2023-11-09
Fix for 1934

sbrossie:
We introduce a different logic for usage code to convert dates to guarantee some stability across accounts created with the same TZ but at different dates during the year (daylight saving or standard). We isolated the code in such a way that we could easily revert (or make it configurable) if needed.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/config/MultiTenantInvoiceConfig.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/config/MultiTenantInvoiceConfig.java
@@ -223,12 +223,28 @@ public class MultiTenantInvoiceConfig extends MultiTenantLockAwareConfigBase imp
         if (result != null){
             return UsageDetailMode.valueOf(result);
         }
-
         if (mode == UsageDetailMode.AGGREGATE || mode == UsageDetailMode.DETAIL) {
             return mode;
         }
-
         return UsageDetailMode.AGGREGATE;
+    }
+
+    @Override
+    public AccountTzOffset getAccountTzOffsetMode() {
+        final AccountTzOffset mode = staticConfig.getAccountTzOffsetMode();
+        if (mode == AccountTzOffset.FIXED || mode == AccountTzOffset.VARIABLE) {
+            return mode;
+        }
+        return AccountTzOffset.FIXED;
+    }
+
+    @Override
+    public AccountTzOffset getAccountTzOffsetMode(final InternalTenantContext tenantContext) {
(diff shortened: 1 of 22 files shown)


pierre:
Essentially, we ignore daylight savings in the context of usage? This is the opposite of what we typically do, so I would suggest we introduce an explicit property for it.

sbrossie:
> Essentially, we ignore daylight savings in the context of usage? This is the opposite of what we typically do, so I would suggest we introduce an explicit property for it.

Introduced config in https://github.com/killbill/killbill/pull/1935/commits/72fd3450ff380f598b515561f3041e42d357db73

util/src/main/java/org/killbill/billing/util/config/definition/InvoiceConfig.java
```diff
@@ -167,6 +181,17 @@ public enum InArrearMode {
     @Description("How the result for an item will be reported (aggregate mode or detail mode). ")
     UsageDetailMode getItemResultBehaviorMode(@Param("dummy") final InternalTenantContext tenantContext);
 
+
+    @Config("org.killbill.invoice.usage.tz.mode")
+    @Default("FIXED")
+    @Description("Behavior to include usage points with respect to day light saving")
```
pierre, line 187: @reshmabidikar Could you add this to our docs?

invoice/src/main/java/org/killbill/billing/invoice/config/MultiTenantInvoiceConfig.java
```diff
@@ -223,14 +223,30 @@ public UsageDetailMode getItemResultBehaviorMode(final InternalTenantContext ten
...
             return mode;
         }
-
         return UsageDetailMode.AGGREGATE;
     }
 
+    @Override
+    public AccountTzOffset getAccountTzOffsetMode() {
+        final AccountTzOffset mode = staticConfig.getAccountTzOffsetMode();
+        if (mode == AccountTzOffset.FIXED || mode == AccountTzOffset.VARIABLE) {
```
pierre, line 235: I see we're doing this in other enum configs as well. Do you remember why? Don't we just need a null check?

pierre (approved):
FYI integration tests are failing.

invoice/src/main/java/org/killbill/billing/invoice/config/MultiTenantInvoiceConfig.java
sbrossie: I don 't think this is really required, I was trying to be consistent -  I suspect because we start from a 'string' we want double check it is one of the expected value and if not default.

PR #1942, opened 2024-01-04
Changes for fixed proration

reshmabidikar:
**Changes for fixed proration:**
1. Introduced a new property: `org.killbill.invoice.proration.fixed.days`
2. Added new tests: `TestInvoiceDateUtilsWithFixedProration` and `TestInvoiceDateUtilsWithoutFixedProration`
3. Tests run fine. I also did a round of manual testing via Kaui and that works fine too.

**Need your input on the following:**
The [Item](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/tree/Item.java) class also invokes the `InvoiceDateUtils.calculateProrationBetweenDates` [here](https://github.com/killbill/killbill/blob/38e4d3f0ea6d60282b4b87d7ac35341fcdd47290/invoice/src/main/java/org/killbill/billing/invoice/tree/Item.java#L154). Since this class does not have access to `InvoiceConfig`, I have passed the value `0` corresponding to `fixedDaysInMonth`. See  [this](https://github.com/killbill/killbill/compare/master...reshmabidikar:killbill:fixed-proration-2#diff-94c7eb4ba0f8059614e7f6c0c53fcc2cb5163035860ed880d5ce5ddc4ec48a8eL154) comment. I did not spend much time investigating what this class does and whether it requires reading the config parameter but if you think this is important, I'll look into it. Let me know.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
@@ -95,7 +95,7 @@ public class FixedAndRecurringInvoiceItemGenerator extends InvoiceItemGenerator 
 
         final InvoicePruner invoicePruner = new InvoicePruner(existingInvoices);
         final Set<UUID> toBeIgnored = invoicePruner.getFullyRepairedItemsClosure();
-        final AccountItemTree accountItemTree = new AccountItemTree(account.getId(), invoiceId);
+        final AccountItemTree accountItemTree = new AccountItemTree(account.getId(), invoiceId, config);
         for (final Invoice invoice : existingInvoices.getInvoices()) {
             for (final InvoiceItem item : invoice.getInvoiceItems()) {
                 if (toBeIgnored.contains(item.getId())) {
--- a/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
@@ -347,7 +347,7 @@ public class FixedAndRecurringInvoiceItemGenerator extends InvoiceItemGenerator 
         // is to charge for that period
         //
         if (endDate != null && !endDate.isAfter(billingIntervalDetail.getFirstBillingCycleDate())) {
-            final BigDecimal leadingProRationPeriods = calculateProRationBeforeFirstBillingPeriod(startDate, endDate, billingPeriod);
+            final BigDecimal leadingProRationPeriods = calculateProRationBeforeFirstBillingPeriod(startDate, endDate, billingPeriod, config);
             final RecurringInvoiceItemData itemData = new RecurringInvoiceItemData(startDate, endDate, leadingProRationPeriods);
             results.add(itemData);
             return new RecurringInvoiceItemDataWithNextBillingCycleDate(results, billingIntervalDetail);
@@ -359,7 +359,7 @@ public class FixedAndRecurringInvoiceItemGenerator extends InvoiceItemGenerator 
         // ii) The endDate is is not null and is strictly after our firstBillingCycleDate (previous check)
         //
         if (billingIntervalDetail.getFirstBillingCycleDate().isAfter(startDate)) {
-            final BigDecimal leadingProRationPeriods = calculateProRationBeforeFirstBillingPeriod(startDate, billingIntervalDetail.getFirstBillingCycleDate(), billingPeriod);
+            final BigDecimal leadingProRationPeriods = calculateProRationBeforeFirstBillingPeriod(startDate, billingIntervalDetail.getFirstBillingCycleDate(), billingPeriod, config);
             if (leadingProRationPeriods != null && leadingProRationPeriods.compareTo(BigDecimal.ZERO) > 0) {
                 // Not common - add info in the logs for debugging purposes
                 final RecurringInvoiceItemData itemData = new RecurringInvoiceItemData(startDate, billingIntervalDetail.getFirstBillingCycleDate(), leadingProRationPeriods);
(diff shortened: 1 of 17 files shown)


invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceDateUtils.java
```diff
@@ -26,7 +26,9 @@
 import org.joda.time.Weeks;
 import org.joda.time.Years;
 import org.killbill.billing.catalog.api.BillingPeriod;
+import org.killbill.billing.util.config.definition.InvoiceConfig;
 import org.killbill.billing.util.currency.KillBillMoney;
+import org.killbill.commons.utils.annotation.VisibleForTesting;
 
 public class InvoiceDateUtils {
```
sbrossie, line 33: This class currently only offers `static` method, because there is no state that is kept. However, now that we need the config (or at least one param of the config), this makes less sense. It would be better to have a real instance - perhaps injected with guice, that takes a CTOR with the `InvoiceConfig`, in such a way that we don't mix request parameters (`startDate`, ...) and static state from the system (`config`).

1. Could you experiment to see if this is not too much trouble doing this - I think the result would be much cleaner
2. If this proves difficult, then at least let's not pass the `InvoiceConfig`, but instead just the Integer for the `prorationFixedDays`

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceDateUtils.java
```diff
@@ -69,23 +71,38 @@ public static BigDecimal calculateProRationAfterLastBillingCycleDate(final Local
...
     }
 
+    @VisibleForTesting
+    public static int daysBetweenWithFixedDaysInMonth(final LocalDate startDate, final LocalDate endDate, final int fixedDaysInMonth) {
+        final int daysBetween = Days.daysBetween(startDate, endDate).getDays();
+        if(startDate.getMonthOfYear() == endDate.getMonthOfYear()) { //same month, no need for extra logic
+            return daysBetween;
+        }
+        final int lastDayOfMonth = startDate.dayOfMonth().withMaximumValue().getDayOfMonth();
+        if (lastDayOfMonth == fixedDaysInMonth) {
```
sbrossie, line 99: The `if` is not even needed because your `else` would return `daysBetween` in this case.

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceDateUtils.java
```diff
@@ -69,23 +71,38 @@ public static BigDecimal calculateProRationAfterLastBillingCycleDate(final Local
...
+            return daysBetween;
+        } else {
+            return daysBetween - (lastDayOfMonth - fixedDaysInMonth);
```
sbrossie, line 102: It's 🎉  that it works for both when `lastDayOfMonth > fixedDaysInMonth` or the reverse (and even the equality has commented above).

invoice/src/main/java/org/killbill/billing/invoice/tree/AccountItemTree.java
```diff
@@ -57,13 +58,16 @@ public class AccountItemTree {
...
+    private InvoiceConfig invoiceConfig;
+
+    public AccountItemTree(final UUID accountId, final UUID targetInvoiceId, final InvoiceConfig invoiceConfig) {
         this.accountId = accountId;
         this.targetInvoiceId = targetInvoiceId;
         this.subscriptionItemTree = new HashMap<UUID, SubscriptionItemTree>();
         this.isBuilt = false;
         this.allExistingItems = new LinkedList<InvoiceItem>();
         this.pendingItemAdj = new LinkedList<InvoiceItem>();
+        this.invoiceConfig = invoiceConfig;
```
sbrossie, line 70: Similar remark as for the class `InvoiceDateUtils`. Here we don't offer static method, but these objects are created on the fly with request parameters, and I think it would be best to mix the `InvoiceConfig` - not that most methods would not be available anyways because they need a context, which we don't pass. So here, instead I suggest to pass the integer for `prorationFixedDays`.

sbrossie (commented):
This is great, I just have a few comments around code organization and specifically passing or not the `InvoiceConfig`. Take a look and let me know what you think.

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceDateUtils.java
reshmabidikar: As discussed, I've updated the methods in `InvoiceDateUtils` to accept `prorationFixedDays` instead of `invoiceConfig`.

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceDateUtils.java
reshmabidikar: Sorry, but I do not follow this comment, which `if` statement are you referring to?

invoice/src/main/java/org/killbill/billing/invoice/tree/AccountItemTree.java
reshmabidikar: As discussed, I've updated `AccountItemTree` (and the other related classes) to accept `prorationFixedDays` instead of `invoiceConfig`.

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceDateUtils.java
sbrossie: The one on line `99` above my comment. What I am saying is that:

```
if (lastDayOfMonth == fixedDaysInMonth) {
           return daysBetween;
        } else {
            return daysBetween - (lastDayOfMonth - fixedDaysInMonth);
```

Is the same as: `return daysBetween - (lastDayOfMonth - fixedDaysInMonth);` so we could simplify - not need for `if ... else`

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
```diff
@@ -95,7 +95,7 @@ public InvoiceGeneratorResult generateItems(final ImmutableAccountData account,
 
         final InvoicePruner invoicePruner = new InvoicePruner(existingInvoices);
         final Set<UUID> toBeIgnored = invoicePruner.getFullyRepairedItemsClosure();
-        final AccountItemTree accountItemTree = new AccountItemTree(account.getId(), invoiceId);
+        final AccountItemTree accountItemTree = new AccountItemTree(account.getId(), invoiceId, config.getProrationFixedDays());
```
sbrossie, line 98: ⚠️ We need to call the version of the config API that takes a context - whenever, we offer a version that takes a context, it needs to be used. Unfortunately, we don't have a way to statically check for this, so this requires careful attention in the code review... Could you make sure we *only* use this version for `src/main` files - for tests, it's ok to use the other form.

reshmabidikar: fixed in 6f6206c

invoice/src/main/java/org/killbill/billing/invoice/generator/InvoiceDateUtils.java
reshmabidikar: Understood, fixed in 6f6206c

sbrossie (approved):

Issue #1950, opened 2024-01-19
Perf issue when committing an invoice


sbrossie:
For any code path where `ExistingInvoiceMetadata` is empty, we end up doing a loop to load all the invoice items associated with one - or multiple invoices as shown [here](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L436).

The idea behind `ExistingInvoiceMetadata` was to provide a cache of invoices and items in such a way that we can safely iterate over them without performance issues, but there are paths where the cache is empty.

sbrossie:
We should check all the code paths where we call `DefaultInvoiceDao#createInvoices` with a null `existingInvoiceMetadata` parameter, and either check whether those are:
1.  Already available to the caller - and up to date
2. Not available, but we could do one query to populate the cache, or as an alternative, have more logic in the `ExistingInvoiceMetadata` class to load items on a per invoice basis when we detect there are absent.

reshmabidikar:
From what I understand, `ExistingInvoiceMetadata` represents a cache of invoice and invoice items. Since the invoice items cache is empty, each invoice item is fetched from the DB, causing a performance issue. Does that sound right?

A few questions:

> The idea behind ExistingInvoiceMetadata was to provide a cache of invoices and items.

Does this refer to a cache of existing invoices for a particular account?

sbrossie:
> From what I understand, ExistingInvoiceMetadata represents a cache of invoice and invoice items. Since the invoice items cache is empty, each invoice item is fetched from the DB, causing a performance issue. Does that sound right?

Yes.

PR #1954, opened 2024-01-22
Changes for 1950

reshmabidikar:
Draft PR to demonstrate the changes for #1950. If this looks good, I will go through the code to check other places where  `DefaultInvoiceDao#createInvoices` is invoked with a null `existingInvoiceMetadata` parameter and make similar changes.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
@@ -120,9 +120,18 @@ public class InvoiceApiHelper {
                 invoiceModelDaos.add(invoiceModelDao);
             }
 
-            List<Invoice> invoices = new LinkedList<>();
-            invoicesForPlugins.forEach(invoice -> invoices.add((Invoice)invoice));
-            ExistingInvoiceMetadata existingInvoiceMetadata = new ExistingInvoiceMetadata(invoices);
+            final List<Invoice> invoices = new LinkedList<>();
+            for (Invoice invoice : invoicesForPlugins) {
+                try {
+                    final Invoice invoiceFromDB = new DefaultInvoice(dao.getById(invoice.getId(), internalCallContext));
+                    if (invoiceFromDB != null) {
+                        invoices.add(invoiceFromDB);
+                    }
+                }
+                catch (InvoiceApiException e) { //invoice not present in DB, do nothing
+                }
+            }
+            final ExistingInvoiceMetadata existingInvoiceMetadata = new ExistingInvoiceMetadata(invoices);
             final List<InvoiceItemModelDao> createdInvoiceItems = dao.createInvoices(invoiceModelDaos, null, Collections.emptySet(), null, existingInvoiceMetadata, true,  internalCallContext);
             success = true;
 
--- a/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java
@@ -38,6 +38,7 @@ import org.killbill.billing.catalog.api.Currency;
 import org.killbill.billing.invoice.InvoicePluginDispatcher;
 import org.killbill.billing.invoice.InvoicePluginDispatcher.AdditionalInvoiceItemsResult;
 import org.killbill.billing.invoice.InvoicePluginDispatcher.PriorCallResult;
+import org.killbill.billing.invoice.dao.ExistingInvoiceMetadata;
 import org.killbill.billing.invoice.dao.InvoiceDao;
 import org.killbill.billing.invoice.dao.InvoiceItemModelDao;
 import org.killbill.billing.invoice.dao.InvoiceModelDao;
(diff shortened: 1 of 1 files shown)


sbrossie (changes requested):
Some background: 
1. `InvoiceApiHelper#dispatchToInvoicePluginsAndInsertItems` is only called from API calls (commit, void, refund, ...) and no during a invoice run. I believe we only manipulate one invoice and technically we should be passing all invoices to the plugin but [in reality we pass null](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java#L81). Also, the [withAccountLock#prepareInvoices](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java#L103) call in all these cases only return the current invoice (the one on which we are doing an operation).
2. There is a parallel path for invoice run that calls directly `invoicePluginDispatcher#updateOriginalInvoiceWithPluginInvoiceItems`.

The purpose of `invoicePluginDispatcher#updateOriginalInvoiceWithPluginInvoiceItems` is to add extra items into the input [originalInvoice](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/InvoicePluginDispatcher.java#L315).

So, with that being said:
1. I don't think we can simply iterate through `invoicesForPlugins` and just do a `dao.getById` because it would miss the additional items added by the plugin
2. I believe in all cases `invoicesForPlugins` is only a list of one invoice, the one we are doing the update on - so maybe we can simplify; if this is true, then we already have the invoice so we can simply add it in our cache `ExistingInvoiceMetadata`.


So, can you verify double check and verify all `API paths` calling `InvoiceApiHelper#dispatchToInvoicePluginsAndInsertItems` only manipulate one invoicem and if so, simply add that Invoice in the cache.

In terms of testing:
1. It might be worth re-reading the test [TestWithInvoicePlugin](https://github.com/killbill/killbill/blob/master/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestWithInvoicePlugin.java#L96), and make sure we correctly ensure that no regression happened.
2. We should make sure the [invoice grouping](https://github.com/killbill/killbill-invgrp-demo/blob/master/src/test/java/org/killbill/billing/plugin/invgrp/TestIntegration.java) test still works
3. Also, verifying that prior fix, doing a commit on an invoice will indeed end fetching all invoice ietms one by one and after the fix,  the cache can be used so this is solved.



Once all this is done let's investigate the following:
1. Can we revisit if we really need these query path: a/ either we can make sure cache is always populated, so no need to query inside, or b/ we could still query a particular invoice, but then can we at least avoid this [one](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/dao/ExistingInvoiceMetadata.java#L60), i.e as we load an invoice, can we load the invoice ietms with it in one GO.
2. Can we revisit the way this [line](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L437) to make sure we never end up in a case like the one reported where we do 1000X queries to fetch all items one by one.

reshmabidikar:
>  believe in all cases invoicesForPlugins is only a list of one invoice, the one we are doing the update on - so maybe we can simplify; if this is true, then we already have the invoice so we can simply add it in our cache ExistingInvoiceMetadata.

So, from what I understand, you are suggesting avoiding the call to `dao.getById` and instead adding the invoice on which the operation is being performed to the `ExistingMetadata` cache. Is that correct?

I think this will work for the endpoints like `commitInvoice`/`voidInvoice`, etc. where the invoice already exists in the DB.

The `InvoiceApiHelper#dispatchToInvoicePluginsAndInsertItems` method is also invoked indirectly via methods that create an invoice/invoice items like `DefaultUserApi.insertExternalCharges`, `DefaultUserApi.insertTaxItems`, etc. ([DefaultUserApi.insertExternalCharges](https://github.com/killbill/killbill/blob/cf829df6d04e9c24c2b5d871b57085e918713741/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java#L349) --> [DefaultUserApi.insertItems](https://github.com/killbill/killbill/blob/cf829df6d04e9c24c2b5d871b57085e918713741/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java#L533) -->  [InvoiceApiHelper#dispatchToInvoicePluginsAndInsertItems](https://github.com/killbill/killbill/blob/cf829df6d04e9c24c2b5d871b57085e918713741/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java#L667)). In this case, the invoice/invoice items being operated on is not present in the DB yet, it is still in memory. So, we cannot add it to the cache. If at all we add such invoices to the cache, they will not be created in the database as [this line](https://github.com/killbill/killbill/blob/cf829df6d04e9c24c2b5d871b57085e918713741/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L399) would be false.

So to reply to the following:

> So, can you double-check and verify all API paths calling InvoiceApiHelper#dispatchToInvoicePluginsAndInsertItems only manipulate one invoice, and if so, simply to just add that Invoice in the cache.

I'll check this, but even if all the API paths calling `InvoiceApiHelper#dispatchToInvoicePluginsAndInsertItems` only manipulate one invoice, we cannot directly add it to the cache, we will need to have a check to ensure it exists in the DB.

Now to come to your other point:

> I don't think we can simply iterate through invoicesForPlugins and just do a dao.getById because it would miss the additional items added by the plugin

Yes, I agree, doing a `dao.getById` would indeed miss the additional plugin items. 

So, perhaps to handle both the issues mentioned above, we keep the `dao.getById` check and if the invoice exists in the DB, we [add ](https://github.com/reshmabidikar/killbill/blob/5e20c203742be75e3084ee1bf3bfd60c3237edd8/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java#L128)the `invoicesForPlugin` to the cache instead of `invoiceFromDB`? 

Just thinking aloud at this point, let me know if all the above makes sense.

sbrossie:
Let's forget some of my previous comments, and resume from your original fix. 

> Yes, I agree, doing a dao.getById would indeed miss the additional plugin items.


I think at the very least we need to address this:
1. Could you look at the various operations calling `invoicePluginDispatcher#updateOriginalInvoiceWithPluginInvoiceItems` (void, commit, credits, ... ) and for each see if it even makes sense for plugins to add items: Clearly for operations like `void` and `commit` it does not make sense.
2. If we find cases where it makes sense, then can you suggest a fix to ensure such items have been added in the `ExistingMetadata` cache - and ideally write and reuse a test case to verify it.

Issue #1951, opened 2024-01-19
Perf issue for computing repair map


sbrossie:
It was reported that the call `InvoiceItemSqlDao#getRepairMap` that is used when we populate the invoice may take a long time, i.e several seconds - although cardinality is not very large - and may be called multiple times for the same invoice.

1. Do we really need this query - could we compute such repaired invoices based on the invoice items that we have already fetched?
2. If we do, we need to review 1/ performance and indices 2/ code path to understand whether this is called too many times.

Note: The report came from a commit invoice path.

reshmabidikar:
Addressed in #1964. Closing this issue.

PR #1964, opened 2024-02-05
Changes for 1951

reshmabidikar:
Attempt to address #1951. Some comments:

1. This is the code path: [DefaultInvoiceDao.getById](https://github.com/killbill/killbill/blob/4153f85a3e84af28c3f748dad2da50e1845eb298/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java#L229) --> [InvoiceDaoHelper.populateChildren](https://github.com/killbill/killbill/blob/4153f85a3e84af28c3f748dad2da50e1845eb298/invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java#L249) --> [InvoiceDaoHelper.setInvoiceRepaired](https://github.com/killbill/killbill/blob/4153f85a3e84af28c3f748dad2da50e1845eb298/invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java#L392) --> [InvoiceItemSqlDao.getRepairMap](https://github.com/killbill/killbill/blob/4153f85a3e84af28c3f748dad2da50e1845eb298/invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceItemSqlDao.java#L74). So basically, the `InvoiceItemSqlDao.getRepairMap` is invoked every time `DefaultInvoiceDao.getById` is invoked. 
2. The only place the repair status is read is in [DefaultInvoiceUserApi.java.voidInvoice](https://github.com/killbill/killbill/blob/4153f85a3e84af28c3f748dad2da50e1845eb298/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java#L759). In all other places, the repair status is not used. In fact, `DefaultInvoice` does not even have a field corresponding to the repair status
3. Due to this, I've updated the code to invoke `InvoiceDaoHelper.setInvoiceRepaired` only for the void invoice code path.

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/dao/CBADao.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/dao/CBADao.java
@@ -132,7 +132,7 @@ public class CBADao {
         final List<InvoiceModelDao> invoices = new ArrayList<>();
         for (UUID id : invoiceIds) {
             final InvoiceModelDao invoice = transInvoiceDao.getById(id.toString(), context);
-            invoiceDaoHelper.populateChildren(invoice, invoicesTags, false, entitySqlDaoWrapperFactory, context); //TODO_1951 - should this be true?
+            invoiceDaoHelper.populateChildren(invoice, invoicesTags, false, entitySqlDaoWrapperFactory, context);
             invoices.add(invoice);
         }
 
--- a/invoice/src/main/java/org/killbill/billing/invoice/dao/CBADao.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/dao/CBADao.java
@@ -132,7 +132,7 @@ public class CBADao {
         final List<InvoiceModelDao> invoices = new ArrayList<>();
         for (UUID id : invoiceIds) {
             final InvoiceModelDao invoice = transInvoiceDao.getById(id.toString(), context);
-            invoiceDaoHelper.populateChildren(invoice, invoicesTags, entitySqlDaoWrapperFactory, context);
+            invoiceDaoHelper.populateChildren(invoice, invoicesTags, false, entitySqlDaoWrapperFactory, context); //TODO_1951 - should this be true?
             invoices.add(invoice);
         }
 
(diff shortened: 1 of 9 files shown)


invoice/src/main/java/org/killbill/billing/invoice/dao/CBADao.java
```diff
@@ -132,7 +132,7 @@ public Set<UUID> doCBAComplexityFromTransaction(final Set<UUID> invoiceIds,
         final List<InvoiceModelDao> invoices = new ArrayList<>();
         for (UUID id : invoiceIds) {
             final InvoiceModelDao invoice = transInvoiceDao.getById(id.toString(), context);
-            invoiceDaoHelper.populateChildren(invoice, invoicesTags, entitySqlDaoWrapperFactory, context);
+            invoiceDaoHelper.populateChildren(invoice, invoicesTags, false, entitySqlDaoWrapperFactory, context); //TODO_1951 - should this be true?
```
sbrossie, line 135: I think it should be ok with `false` - you can remove TODO: Since `isRepaired` is only needed when doing void operations, and those are not called during void, should be fine.

invoice/src/test/java/org/killbill/billing/invoice/dao/MockInvoiceDao.java
```diff
@@ -134,6 +134,13 @@ public InvoiceModelDao getById(final UUID id, final InternalTenantContext contex
         }
     }
 
+    @Override //TODO_1951 - revisit to check if anything else needs to be done
```
sbrossie, line 137: It's fine, you can remove TODO - if not of the tests break, this mock does not use `is_repair` functionality. If later we need it, we can add it.

sbrossie (approved):
Looks good.

I am still curious though about the 2 following things:
1. Why is this query slow - if nothing obvious, this is fine.
2. Do we even need the query, could we compute the `is_repair` state without the query simply by 'joining' in memory the items we have fetched?

reshmabidikar:
Deleted the TODO statements.

With regards to the comment below:

> Do we even need the query, could we compute the is_repair state without the query simply by 'joining' in memory the items we have fetched?

We do not need the query. The void operation however only loads the invoice being voided from the database. So,  we will need to fetch the other invoices corresponding to the account to check if there is an invoice with a `REPAIR_ADJ` item. Let me know if this makes sense and if you would like to me go ahead with the change.

Issue #1959, opened 2024-01-31
Inconsistency in catalog endpoints


sbrossie:
Our catalog rest api GET /1.0/kb/catalog vs GET /1.0/kb/catalog/xml behave differently when some products don't have plans. In XML case, everything is returned, but for json endpoint, we don't return empty products.

The issue seems to be [here](https://github.com/killbill/killbill/blob/ba491dbf8561e582580d2956a224b10d6a8a7ff0/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/CatalogJson.java#L99-L99).

PR #1970, opened 2024-02-10
https://github.com/killbill/killbill/issues/1959

KBbitsP:
Add products that are not part of any plan to the fetch catalogJson endpoint output

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/CatalogJson.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/CatalogJson.java
@@ -28,9 +28,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import org.joda.time.DateTime;
-import org.killbill.billing.catalog.api.BillingPeriod;
 import org.killbill.billing.catalog.api.BillingMode;
+import org.killbill.billing.catalog.api.BillingPeriod;
 import org.killbill.billing.catalog.api.CatalogApiException;
 import org.killbill.billing.catalog.api.Currency;
 import org.killbill.billing.catalog.api.CurrencyValueNull;
@@ -55,7 +54,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 import io.swagger.annotations.ApiModel;
 
-@ApiModel(value="Catalog")
+@ApiModel(value = "Catalog")
 public class CatalogJson {
 
     private final String name;
@@ -66,8 +65,7 @@ public class CatalogJson {
     private final List<PriceListJson> priceLists;
 
     @JsonCreator
-    public CatalogJson(@JsonProperty("name") final String name,
-                       @JsonProperty("effectiveDate") final Date effectiveDate,
+    public CatalogJson(@JsonProperty("name") final String name, @JsonProperty("effectiveDate") final Date effectiveDate,
                        @JsonProperty("currencies") final List<Currency> currencies,
                        @JsonProperty("units") final List<UnitJson> units,
                        @JsonProperty("products") final List<ProductJson> products,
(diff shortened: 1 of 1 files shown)


jaxrs/src/main/java/org/killbill/billing/jaxrs/json/CatalogJson.java
```diff
@@ -66,8 +65,7 @@ public class CatalogJson {
     private final List<PriceListJson> priceLists;
 
     @JsonCreator
-    public CatalogJson(@JsonProperty("name") final String name,
-                       @JsonProperty("effectiveDate") final Date effectiveDate,
+    public CatalogJson(@JsonProperty("name") final String name, @JsonProperty("effectiveDate") final Date effectiveDate,
```
sbrossie, line 68: Nit: Can we keep one argument per line - the way it was. In general we chose to break one per line or have all on he same line, not a mix of the two.

sbrossie (changes requested):
Can you please add a test to verify this works as expected?

Also, most of the diffs in this PR are spacing (and touching files that don't have any relevant changes), which makes it really hard to review. I believe the only file that should be changed is `CatalogJson.java`, and nothing else, is this correct?

reshmabidikar (commented):
Same comments as @sbrossie. With regards to writing a test,  maybe you can write one in [TestCatalog](https://github.com/killbill/killbill/blob/master/profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestCatalog.java).

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestCatalog.java
```diff
@@ -62,9 +63,22 @@ public void testMultiTenantCatalog() throws Exception {
...
-        // Standalone catalog and ids (JAXB name) are not unique across the various catalogs so deserialization would fail
-        //
+        // We can't deserialize the VersionedCatalog using our JAXB models because it
+        // contains several
+        // Standalone catalog and ids (JAXB name) are not unique across the various
+        // catalogs so deserialization would fail
+    }
+
+    @Test(groups = "slow", description = "Match the number of products in JSON and XML catalogs")
+    public void matchJsonXmlCatalogProductSize() throws Exception {
```
sbrossie, line 73: Our test methods typically start with `test...`

sbrossie (approved):

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/CatalogJson.java
```diff
@@ -87,6 +87,7 @@ public CatalogJson(final StaticCatalog catalog) throws CatalogApiException {
         currencies = Arrays.asList(catalog.getSupportedCurrencies());
         priceLists = new ArrayList<PriceListJson>();
 
+        final Collection<Product> XMLproducts = catalog.getProducts();
```
reshmabidikar, line 90: The variable is not named as per coding conventions, typically variable name should start with a lowercase letter and follow camelcase. Also, it has nothing to do with XML. So consider renaming to `catalogProducts` for clarity.

jaxrs/src/main/java/org/killbill/billing/jaxrs/json/CatalogJson.java
```diff
@@ -112,6 +113,17 @@ public CatalogJson(final StaticCatalog catalog) throws CatalogApiException {
             final PlanJson planJson = new PlanJson(plan);
             productJson.getPlans().add(planJson);
         }
+        // Add products that are not part of any plan
+        for (final Product product : XMLproducts) {
+            final String productName = product.getName();
+            ProductJson productJson = productMap.get(product.getName());
+            if (!productMap.containsKey(productName)) {
+                if (productJson == null) {
```
reshmabidikar, line 121: is this `if` condition required? If the `productMap` does not contain the `productName`, wouldn't the `productJson` be null?

reshmabidikar (approved):
Left some minor comments.

PR #1973, opened 2024-02-14
Deleted the org.killbill.template.invoiceFormatterFactoryClass property

reshmabidikar:
While documenting the new properties, I realized that the `org.killbill.template.invoiceFormatterFactoryClass` property is no longer needed.  If at all a CustomInvoiceFormatter is required, it will be injected via the plugin. So only the `org.killbill.template.invoiceFormatterFactoryPluginName` property should suffice.


pierre (approved):
Technically, this is a breaking change. That being said, I don't think this feature has been used over the years. Let's just make sure to add a note in the releases notes.

Issue #1990, opened 2024-03-21
Invoice plugins not called on refund


pierre:
70ebae3608a362028d486cea5d57c3cf03aaeb91 introduced a regression. Invoice plugins aren't called anymore for refunds (we don't adjust taxes on refunds anymore).

reshmabidikar:
For [recordRefund](https://github.com/killbill/killbill/blob/b7dc94ada8f18513b49f1e5f58c930d49969ae22/invoice/src/main/java/org/killbill/billing/invoice/api/svcs/DefaultInvoiceInternalApi.java#L130), I think the [getAdditionalInvoiceItems](https://github.com/killbill/killbill-avatax-plugin/blob/08a56cc65fbe01736b1b7931731b88f32a9e59ed/src/main/java/org/killbill/billing/plugin/avatax/api/AvaTaxInvoicePluginApi.java#L81C34-L81C59) method needs to be invoked so that the plugin can return additional invoice items corresponding to refund. So, we would simply need to undo the change to `recordRefund` (that is invoke [InvoiceApiHelper#dispatchToInvoicePluginsAndInsertItems](https://github.com/killbill/killbill/blob/b7dc94ada8f18513b49f1e5f58c930d49969ae22/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java#L75C30-L75C68) from the `recordRefund` method). If this makes sense, I will go ahead with the change.

sbrossie:
Fixed in https://github.com/killbill/killbill/pull/1996

Issue #1989, opened 2024-03-21
Invoice plugins not called on commit/void


pierre:
0311d5351f9b187a7137e628925f25eef0fd299b introduced a regression. For Avatax for instance, the plugin needs to be notified to commit the document in Avalara.

PR #1962, opened 2024-02-01
Changes for recordRefund and createItemAdjustment

reshmabidikar:
Round 2 changes for #1950 

- Removed call to `invoiceApiHelper.dispatchToInvoicePluginsAndInsertItems` for `recordRefund`
- Updated `invoiceApiHelper.dispatchToInvoicePluginsAndInsertItems` to query the database and add existing invoices to the cache before invoking `dao.createInvoices`. This is so that the cache is queried for the [DefaultInvoiceUserApi.insertInvoiceItemAdjustment](https://github.com/killbill/killbill/blob/84d1e887d3203dedd381ac9b3ff3282bad155381/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java#L409C24-L409C51) code path.

PR #1991, opened 2024-03-24
Fix for 1989 and 1990

reshmabidikar:
Fix for #1989 (See [433577b](https://github.com/killbill/killbill/pull/1991/commits/433577b3d8dd2042995b43cd4942884ce8c0f2ff)).

- As part of the fix for #1950, we had removed the call to [InvoiceApiHelper.dispatchToInvoicePluginsAndInsertItems](https://github.com/killbill/killbill/blob/b7dc94ada8f18513b49f1e5f58c930d49969ae22/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java#L75C30-L75C68) from the `commitInvoice`/`voidInvoice` methods. (See #1960). The reason for doing this was that [we thought](https://killbillio.slack.com/archives/G01C2P3V8G7/p1706662205027009?thread_ts=1706593497.615939&cid=G01C2P3V8G7) it was unnecessary to invoke invoice plugins as part of the commit/void calls. However, we now know that it is required to call invoice plugins in case of the void/commit calls, otherwise the Avatax integration tests fail.

- Thus, 433577b reverts the changes done to the  `commitInvoice`/`voidInvoice` methods. 

- This will not impact the performance improvements for #1950. As part of #1962, we have added [this change](https://github.com/killbill/killbill/blob/b7dc94ada8f18513b49f1e5f58c930d49969ae22/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java#L123:L134).  This will ensure that the `ExistingInvoiceMetadata` cache is populated and thus not cause the performance issue reported in https://github.com/killbill/killbill/issues/1950.

Fix for #1990 (See 4aed737)

- As part of the fix for #1950, we had removed the call to [InvoiceApiHelper.dispatchToInvoicePluginsAndInsertItems](https://github.com/killbill/killbill/blob/b7dc94ada8f18513b49f1e5f58c930d49969ae22/invoice/src/main/java/org/killbill/billing/invoice/api/InvoiceApiHelper.java#L75C30-L75C68) from the [recordRefund](https://github.com/killbill/killbill/blob/b7dc94ada8f18513b49f1e5f58c930d49969ae22/invoice/src/main/java/org/killbill/billing/invoice/api/svcs/DefaultInvoiceInternalApi.java#L130) method (See #1962). However, this causes Avatax integration tests to fail. 
- 4aed737 reverts this change. 
- This will not impact the performance improvements for #1950.

### DIFF
--- a/catalog/pom.xml
+++ b/catalog/pom.xml
@@ -204,8 +204,8 @@
                             <shadedClassifierName>load-tool</shadedClassifierName>
                             <createDependencyReducedPom>false</createDependencyReducedPom>
                             <transformers>
-                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer" />
-                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer" />
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer"/>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer"/>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.DontIncludeResourceTransformer">
                                     <resources>
                                         <resource>MANIFEST.MF</resource>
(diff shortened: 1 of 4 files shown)


sbrossie (commented):
I am not convinced we need to call `invoicePlugin#getAdditionalInvoiceItems` for operations like `void` and `commit` - it still makes no sense that a plugin would 'add' items for these operations.

However, looking quickly at the avalara code, I see that the plugin expects an `INVOICE_OPERATION` [to be passed as a plugin property for invoicePlugin#onSuccess/invoicePlugin#onFailure calls](https://github.com/killbill/killbill-avatax-plugin/blob/master/src/main/java/org/killbill/billing/plugin/avatax/api/AvaTaxInvoicePluginApi.java#L100), and this part is definitely a regression we introduced, i.e. we can see that the variable is still [defined but unused in our latest code base](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/api/user/DefaultInvoiceUserApi.java#L98).

I would recommend to simply pass this additional plugin property and check if this solves the test issue we are seeing - and if not investigate, if we are missing something else.

reshmabidikar:
Closing this PR, will open a separate one.

PR #1995, opened 2024-03-28
search: allow advanced filters

pierre:
The search key is opaque to the system. Interpret the prefix marker `_q=1` as the beginning of an advanced query, allowing to filter on specific fields, e.g.:

```
_q=1&email=john@acme.com&currency=USD
```

It is the responsibility of the user to add relevant indexes.


util/src/main/resources/org/killbill/billing/util/entity/dao/EntitySqlDao.sql.stg
```diff
@@ -332,27 +332,28 @@ where <targetRecordIdField("t.")> = :targetRecordId
 ;
 >>
 
-searchQuery(prefix) ::= <<
-1 = 1
+/** The extra space between "|  (:" matters below */
+searchQuery(prefix, searchAttributes, logicalOperator) ::= <<
+<searchAttributes:{ x |  (:<x.bindingKey> IS NULL OR <prefix><x.column> <x.operator> :<x.bindingKey>) }; separator=logicalOperator>
```
sbrossie, line 337: I see... tricky indeed 😅

account/src/main/java/org/killbill/billing/account/dao/DefaultAccountDao.java
```diff
@@ -167,18 +171,59 @@ public Pagination<AccountModelDao> searchAccounts(final String searchKey, final
...
+                                                 "notes",
+                                                 "migrated",
+                                                 "created_by",
+                                                 "created_date",
+                                                 "updated_by",
+                                                 "updated_date"));
+        } else {
+            searchQuery = new SearchQuery(SqlOperator.OR);
+
+            final String likeSearchKey = String.format("%%%s%%", searchKey);
```
sbrossie, line 207: So if I understand, this is the backward compatible search which hardcodes the columns and associated operators?

account/src/main/java/org/killbill/billing/account/dao/DefaultAccountDao.java
```diff
@@ -167,18 +171,59 @@ public Pagination<AccountModelDao> searchAccounts(final String searchKey, final
                                                           accountModelDao == null ? Collections.emptyIterator() : List.of(accountModelDao).iterator());
         }
 
+        final SearchQuery searchQuery;
+        if (searchKey.startsWith(SEARCH_QUERY_MARKER)) {
+            searchQuery = new SearchQuery(searchKey,
```
sbrossie, line 176: In the 'new free form', we necessarily impose and `AND` ?

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
```diff
@@ -95,61 +89,78 @@ getInvoiceByAccountRecordIdAfter() ::= <<
...
+SELECT
+      x.invoice_id
+    , x.tenant_record_id
+    , SUM(x.amount) as balance
+FROM (
+    SELECT
+          i.id as invoice_id
+        , i.tenant_record_id
+        , 0 as amount
+    FROM invoices i
```
sbrossie, line 102: As discussed, we need to make sure:
- Invoice is Committed (not draft/void)
- Invoice is not migrated
- Invoice is not marked as written_off
- For HA, we need to look at the parent invoice

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
```diff
@@ -95,61 +89,78 @@ getInvoiceByAccountRecordIdAfter() ::= <<
...
+        , SUM(ii.amount) as amount
+    FROM invoice_items ii
+    WHERE
+        ii.type in ('TAX', 'EXTERNAL_CHARGE', 'FIXED', 'USAGE', 'RECURRING', 'ITEM_ADJ', 'REPAIR_ADJ', 'PARENT_SUMMARY', 'CBA_ADJ', 'CREDIT_ADJ')
+        <AND_CHECK_TENANT("ii.")>
+    GROUP BY
+          ii.invoice_id
+        , ii.tenant_record_id
+
+    UNION ALL
```
sbrossie, line 120: Interesting way to write this query...

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -546,17 +554,47 @@ public Pagination<InvoiceModelDao> searchInvoices(final String searchKey, final
 
         final boolean isSearchKeyCurrency = isSearchKeyCurrency(searchKey);
 
+        final SearchQuery searchQuery;
+        if (isSearchKeyCurrency) {
+            searchQuery = new SearchQuery(SqlOperator.OR);
+            searchQuery.addSearchClause("currency", SqlOperator.EQ, searchKey);
+        } else if (searchKey.startsWith(SEARCH_QUERY_MARKER)) {
+            final Matcher matcher = BALANCE_QUERY_PATTERN.matcher(searchKey);
```
sbrossie, line 562: Is the use case that we have to show (return) invoices matching specific balance criteria for the given tenant - and not for a specific account?

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
```diff
@@ -95,61 +89,78 @@ getInvoiceByAccountRecordIdAfter() ::= <<
...
+            , ip.tenant_record_id
+    ) AS payments_and_refunds
+) x
+GROUP BY
+      x.invoice_id
+    , x.tenant_record_id
 >>
 
-searchByCurrency(ordering) ::= <<
+searchInvoicesByBalance(ordering, comparisonOperator) ::= <<
```
sbrossie, line 147: The result(s) we return will match the criteria, but since we just return the columns (fileds), the exact balance will not be returned, is this correct - i.e. balance computation is just used for search criteria and not to populate the objects being returned?

util/src/main/java/org/killbill/billing/util/entity/dao/SearchQuery.java
```diff
@@ -0,0 +1,112 @@
...
+
+import javax.annotation.Nullable;
+
+public class SearchQuery {
+
+    public static final String SEARCH_QUERY_MARKER = "_q=1&";
+
+    private final Map<String, Object> bindMap = new HashMap<>();
+    private final List<SearchAttribute> attributes = new LinkedList<>();
+    private final SqlOperator sqlOperator;
```
sbrossie, line 36: This confused me first, as the `SqlOperator` is used for different purposes. (just a comment)

util/src/main/java/org/killbill/billing/util/entity/dao/SearchQuery.java
```diff
@@ -0,0 +1,112 @@
...
+                return null;
+            }
+        }
+    }
+
+    public void addSearchClause(final String column,
+                                final SqlOperator operator,
+                                final Object bindingValue) {
+        // Must be unique in case of multiple conditions for a given column, e.g. WHERE age >= 18 AND age < 21
+        final String bindingKey = "s_attr_" + column + globalIdx;
```
sbrossie, line 94: Is there a reason why we need this prefix?

sbrossie (commented):
👏

account/src/main/java/org/killbill/billing/account/dao/DefaultAccountDao.java
pierre: Correct. I just wanted to keep backward compatibility for now. But maybe we remove that in 0.26?

account/src/main/java/org/killbill/billing/account/dao/DefaultAccountDao.java
pierre: Yes, I think that's the intuitive behavior. For instance, on GitHub, the `AND` is implicit too:

image

That being said, we could add support for `OR` (just need to define the right syntax from an "api" perspective).

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
pierre: Any concern? It's essentially charges - payments/refunds.  The first subquery which grabs all invoices is to support shell invoices (no invoice items -- at least required for tests).

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
pierre: Yes, the use-case is to filter invoices for that tenant and balance criteria, e.g., show me all unpaid invoices.

There is a branch in the code because it doesn't support additional filtering capabilities: it's either balance filter or any other filter. Cannot do: show me all unpaid invoices in the last month unfortunately. We might be able to do it, it's just a bit hard with the balance specific query...

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
pierre: > balance computation is just used for search criteria and not to populate the objects being returned

Correct.

> since we just return the columns (fileds)

That's a good point, I need to check if the objects are re-hydrated at higher levels of the stack or if I should select more columns here.

util/src/main/java/org/killbill/billing/util/entity/dao/SearchQuery.java
pierre: Yup. I had different objects at first but then just merged them...

util/src/main/java/org/killbill/billing/util/entity/dao/SearchQuery.java
pierre: I just wanted to make sure that if we end up hardcoding a specific binding in the template, this doesn't cause a conflict.

Note that this is never seen: it's replaced by `?` as soon as jdbi renders the Stringtemplate.

pierre:
Note to self: unresolved comments must be addressed before merging.

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
sbrossie: No concerns, LGTM.

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
sbrossie: I am resolving this, since this was addressed in https://github.com/killbill/killbill/pull/2016#issue-2320212379

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
```diff
@@ -95,61 +89,86 @@ getInvoiceByAccountRecordIdAfter() ::= <<
...
-<andCheckSoftDeletionWithComma("t.")>
-<AND_CHECK_TENANT("t.")>)) as count
-;
+invoiceBalanceQuery() ::= <<
+SELECT
+      x.invoice_id
+    , x.tenant_record_id
+    , SUM(x.amount) as balance
+FROM (
+    SELECT
```
sbrossie, line 98: Do we need that first query - is it because we want to return also that `DRAFT` or `VOID` invoices?

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
```diff
@@ -95,61 +89,86 @@ getInvoiceByAccountRecordIdAfter() ::= <<
...
+
+    SELECT
+          ii.invoice_id
+        , ii.tenant_record_id
+        , CASE WHEN inv.status in ('DRAFT', 'VOID') OR inv.migrated = true OR tg.tag_definition_id = '00000000-0000-0000-0000-000000000004' THEN 0 ELSE SUM(ii.amount) END as amount
+ FROM invoice_items ii
+    JOIN invoices inv ON ii.invoice_id = inv.id
+    LEFT OUTER JOIN tags tg ON ii.invoice_id = tg.object_id
+    WHERE
+        ii.type in ('TAX', 'EXTERNAL_CHARGE', 'FIXED', 'USAGE', 'RECURRING', 'ITEM_ADJ', 'REPAIR_ADJ', 'PARENT_SUMMARY', 'CBA_ADJ', 'CREDIT_ADJ')
```
sbrossie, line 116: It looks like all the `InvoiceItemType` are present here, so is this needed?

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
sbrossie: I am just wondering if we could use the `amount` computed from the underlying `invoiceBalanceQuery()` so we don;'t have to rehydrate the invoices. as it is very expensive. However, not returning the balance is an issue as well.

sbrossie: I am just wondering if we could use the `amount` computed from the underlying `invoiceBalanceQuery()` so we don't have to rehydrate the invoices. as it is very expensive. However, not returning the balance is an issue as well.

pierre: We could:

* Modify `InvoiceModelDao` to have a `balance` field. This is set only in this new codepath, when computing the balance in SQL (the `balance` column from the SQL query should automatically populate it if the getter and setter for balance are added in `InvoiceModelDao`.
* Modify `DefaultInvoice` to have a private `balance` field. If set during construction of the object (pulled from `InvoiceModelDao`), `getBalance()` returns it, otherwise, it goes through the computation like today.

@reshmabidikar Could you give this a try? When doing searches, the invoices will still be shallow, but when `searchInvoicesByBalance` is called, the balance should be returned. If this work, we could also look at updating the other search queries to do the same?

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
pierre: We could:

* Modify `InvoiceModelDao` to have a `balance` field. This is set only in this new codepath, when computing the balance in SQL (the `balance` column from the SQL query should automatically populate it if the getter and setter for balance are added in `InvoiceModelDao`.
* Modify `DefaultInvoice` to have a private `balance` field. If set during construction of the object (pulled from `InvoiceModelDao`), `getBalance()` returns it, otherwise, it goes through the computation like today.

@reshmabidikar Could you give this a try? When doing searches, the invoices will still be shallow, but when `searchInvoicesByBalance` is called, the balance should be returned. If this work, we could also look at updating the other search queries to do the same?

sbrossie:
Moving this as 'Ready to review' as we merged https://github.com/killbill/killbill/pull/2023

sbrossie (approved):

sbrossie:
@reshmabidikar Could you create a doc PR to add this new search capability - probably slate doc. 

@pierre Do we have a matching kaui task to use this ?

invoice/src/main/resources/org/killbill/billing/invoice/dao/InvoiceSqlDao.sql.stg
reshmabidikar: Implemented in https://github.com/killbill/killbill/pull/2023. See https://github.com/killbill/killbill/pull/2021#discussion_r1627965080.

reshmabidikar:
> @reshmabidikar Could you create a doc PR to add this new search capability - probably slate doc.

Noted, will do.

PR #2021, opened 2024-06-04
Return balance for search by balance queries

reshmabidikar:


invoice/src/main/java/org/killbill/billing/invoice/model/DefaultInvoice.java
```diff
@@ -268,6 +270,9 @@ public BigDecimal getRefundedAmount() {
 
     @Override
     public BigDecimal getBalance() {
+        if (this.balance != null) {
```
sbrossie, line 273: What you did is well aligned wih what we asked for.

Some (after) thoughts, though: Having this new `balance` field in the `DefaultInvoice` is not too great, because now we need extra logic to check if field is null and if not compute it dynamically.

I think a nicer approach would be to create another implementation of the `Invoice` interface (e.g `SearchInvoice`) which extends `DefaultInvoice` and override the `getBalance()` method. Could you take a quick look - maybe create a branch on top of this one - and if then we can compare. If this is harder than it looks, I am also happy to keep this implementation.

reshmabidikar:
Opened https://github.com/killbill/killbill/pull/2023 to address https://github.com/killbill/killbill/pull/2021#discussion_r1627965080. Closing this PR.

Issue #1952, opened 2024-01-19
Performance issues around tracking ids when populate large invoice (or all invoices for an account)


sbrossie:
The code inside invoice will populate the shell invoice with its items, payments, tags and tracking ids. However there are many code paths where such of these elements are not useful.

In particular, we should look at whether for internal use (e.g. invoice run), we could bypass tracking ids.

A more elaborate approach would be to make such loading lazy and only call it if such data is being used.

sbrossie:
Anytime we rehydrate an invoice from disk, we 'populate' all the elements of the invoice, [including the trackingids](https://github.com/killbill/killbill/blob/killbill-0.24.10/invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java#L241).

I don't believe such trackingIds are used anywhere internally in the code, except for when this needs to be returned to the client; in which case this is first set in the `InvoiceModelDao#trackingIds`, and then in the `DefaultInvoice#trackingIds`.

Since we don't have an api flag to bypass returning such `trackingIds`, the first goal would be to not populate those for any internal KB invoice - i.e we only populate when fetching an `Invoice` from [api](https://github.com/killbill/killbill-api/blob/master/src/main/java/org/killbill/billing/invoice/api/InvoiceUserApi.java#L51).


In a subsequent Kill Bill version (e.g. `0.26.x`) we could consider adding a flag to decide whether to return or not such items based on client needs.

PR #2022, opened 2024-06-06
Round 1 changes for 1952

reshmabidikar:
Round 1 changes for #1952:

- Tracking ids disabled for internal method calls
- Tracking ids returned for API facing method calls

I will implement the following as part of Round 2:

> In addition to those, we should check for all the get paths that are used for apis, whether we also use those internally and possibly introduce a flag to discriminate between internal and external (invoice api) use cases.


invoice/src/main/java/org/killbill/billing/invoice/dao/InvoiceDaoHelper.java
```diff
@@ -281,7 +285,7 @@ public List<InvoiceModelDao> getAllInvoicesByAccountFromTransaction(final Boolea
                                                        .filter(invoice -> includeVoidedInvoices || !InvoiceStatus.VOID.equals(invoice.getStatus()))
                                                        .collect(Collectors.toUnmodifiableList());
         if (includeInvoiceComponents) {
-            populateChildren(filtered, invoicesTags, false, entitySqlDaoWrapperFactory, context);
+            populateChildren(filtered, invoicesTags, false, true, entitySqlDaoWrapperFactory, context);
```
sbrossie, line 288: Good. For pass #2, this probably one option where we could optimize based on use cases, internal vs externals.

sbrossie (approved):

sbrossie (approved):

PR #2024, opened 2024-06-11
Work for 1952 - DRAFT PR for running tests

reshmabidikar:
Round 2 changes:

- `includeTrackingIds` flag added to `getByNumber`, `getByBroup`, `getByInvoiceItem` DAO methods
-  `includeTrackingIds` flag added to `getInvoicesByAccount` DAO methods

### DIFF
--- a/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
+++ b/invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
@@ -167,7 +167,7 @@ public class DefaultInvoiceDao extends EntityDaoBase<InvoiceModelDao, Invoice, I
     }
 
     @Override
-    public List<InvoiceModelDao> getInvoicesByAccount(final Boolean includeVoidedInvoices, final Boolean includeInvoiceComponents, final InternalTenantContext context) {
+    public List<InvoiceModelDao> getInvoicesByAccount(final Boolean includeVoidedInvoices, final Boolean includeInvoiceComponents, final Boolean includeTrackingIds, final InternalTenantContext context) {
         final List<Tag> invoicesTags = getInvoicesTags(context);
 
         return transactionalSqlDao.execute(true, entitySqlDaoWrapperFactory -> {
@@ -180,7 +180,7 @@ public class DefaultInvoiceDao extends EntityDaoBase<InvoiceModelDao, Invoice, I
                                                                             .collect(Collectors.toUnmodifiableList());
 
             if (includeInvoiceComponents) {
-                invoiceDaoHelper.populateChildren(invoices, invoicesTags, false, true, entitySqlDaoWrapperFactory, context);
+                invoiceDaoHelper.populateChildren(invoices, invoicesTags, false, includeTrackingIds, entitySqlDaoWrapperFactory, context);
             }
 
             return invoices;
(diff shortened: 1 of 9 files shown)


invoice/src/main/java/org/killbill/billing/invoice/api/svcs/DefaultInvoiceInternalApi.java
```diff
@@ -90,7 +90,7 @@ private DefaultInvoice getInvoiceByIdInternal(final UUID invoiceId, final Intern
 
     @Override
     public Collection<Invoice> getUnpaidInvoicesByAccountId(final UUID accountId, final LocalDate upToDate, final InternalTenantContext context) {
-        final List<InvoiceModelDao> unpaidInvoicesByAccountId = dao.getUnpaidInvoicesByAccountId(accountId, null, upToDate, context);
```
sbrossie, line 93: yes that makes sense - you can remove he comment: This 'internal' api is used internally and since we never use the tracking ids internally, it should be safe to pass `false` for all methods in this class.

invoice/src/main/java/org/killbill/billing/invoice/dao/CBADao.java
```diff
@@ -191,7 +191,7 @@ private List<InvoiceItemModelDao> useExistingCBAFromTransaction(final BigDecimal
 
         // PERF: Computing the invoice balance is difficult to do in the DB, so we effectively need to retrieve all invoices on the account and filter the unpaid ones in memory.
         // This should be infrequent though because of the account CBA check above.
-        final List<InvoiceModelDao> allInvoices = invoiceDaoHelper.getAllInvoicesByAccountFromTransaction(false, true, invoicesTags, entitySqlDaoWrapperFactory, context);
+        final List<InvoiceModelDao> allInvoices = invoiceDaoHelper.getAllInvoicesByAccountFromTransaction(false, true, false, invoicesTags, entitySqlDaoWrapperFactory, context); //TODO_1952 - non API call, so passing includeTrackingIds=false. Revisit
```
sbrossie, line 194: Yes this makes sense as well, I think comment can go.

invoice/src/main/java/org/killbill/billing/invoice/dao/DefaultInvoiceDao.java
```diff
@@ -574,7 +574,7 @@ public BigDecimal getAccountBalance(final UUID accountId, final InternalTenantCo
             BigDecimal cba = BigDecimal.ZERO;
 
             BigDecimal accountBalance = BigDecimal.ZERO;
-            final List<InvoiceModelDao> invoices = invoiceDaoHelper.getAllInvoicesByAccountFromTransaction(false, true, invoicesTags, entitySqlDaoWrapperFactory, context);
+            final List<InvoiceModelDao> invoices = invoiceDaoHelper.getAllInvoicesByAccountFromTransaction(false, true, false, invoicesTags, entitySqlDaoWrapperFactory, context); //TODO_1952 - non API call, so passing includeTrackingIds=false. Revisit
```
sbrossie, line 577: Yes this good as well, comment can go: The method does not return the invoices

invoice/src/main/java/org/killbill/billing/invoice/optimizer/InvoiceOptimizerExp.java
```diff
@@ -80,7 +80,7 @@ public AccountInvoices getInvoices(final InternalCallContext callContext) {
         //
         final LocalDate beCutoffDt = isMaxInvoiceLimitSet ? cutoffDt.minus(maxInvoiceLimit) : null;
         final List<Invoice> existingInvoices = new LinkedList<Invoice>();
-        final List<InvoiceModelDao> invoicesByAccount = invoiceDao.getInvoicesByAccount(false, cutoffDt, null, true, callContext);
+        final List<InvoiceModelDao> invoicesByAccount = invoiceDao.getInvoicesByAccount(false, cutoffDt, null, true, false, callContext); //TODO_1952 - non API call, so passing includeTrackingIds=false. Revisit
```
sbrossie, line 83: I think this is true as well - comment can go. 

The only time we really need those in the code is when we generate an invoice, and those tracking IDs are explicitly fetched as shown [here](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/usage/RawUsageOptimizer.java#L92). So any internal use of invoices inside KB should not need those.

invoice/src/main/java/org/killbill/billing/invoice/optimizer/InvoiceOptimizerNoop.java
```diff
@@ -53,7 +53,7 @@ public AccountInvoices getInvoices(final InternalCallContext callContext) {
 
         logDisabledFeatureIfNeeded(callContext);
         final List<Invoice> existingInvoices = new LinkedList<Invoice>();
-        final List<InvoiceModelDao> invoicesByAccount = invoiceDao.getInvoicesByAccount(false, true, callContext);
+        final List<InvoiceModelDao> invoicesByAccount = invoiceDao.getInvoicesByAccount(false, true, false, callContext); //TODO_1952 - non API call, so passing includeTrackingIds=false. Revisit
```
sbrossie, line 56: Same comment as above.

sbrossie (approved):

PR #2029, opened 2024-06-18
useGlobalDefault problem

xsalefter:
As [per docs](https://killbill.github.io/slate/tenant.html#create-a-tenant), `createDefaultEmptyCatalog()` should be called only if `useGlobalDefault == true`.


xsalefter:
Closed this one as works as intended.

PR #2031, opened 2024-07-02
Work for 1739 create alignment fix

reshmabidikar:


invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
```diff
@@ -430,11 +447,25 @@ private InvoiceItem generateFixedPriceItem(final UUID invoiceId, final UUID acco
                 Preconditions.checkNotNull(currentPlan, String.format("Unexpected null Plan name event = %s", thisEvent));
 
                 final DateTime catalogEffectiveDate = thisEvent.getCatalogEffectiveDate() != null ? thisEvent.getCatalogEffectiveDate() : null;
+                LocalDate invoiceItemEndDate = null;
+                if (thisEvent.getRecurringPrice() == null && thisEvent.getPlanPhase().getPhaseType() != PhaseType.EVERGREEN) {
+                    try {
+                        if (nextEvent != null && nextEvent.getSubscriptionId().equals(thisEvent.getSubscriptionId()) && (nextEvent.getTransitionType() == SubscriptionBaseTransitionType.PHASE)) {
```
sbrossie, line 453: I would suggest here to not pass a `nextEvent` to this method when this is not the same subscription - similar to what we do for [processRecurringBillingEvents](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java#L154).

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
```diff
@@ -430,11 +447,25 @@ private InvoiceItem generateFixedPriceItem(final UUID invoiceId, final UUID acco
...
+                            invoiceItemEndDate = internalCallContext.toLocalDate(nextEvent.getEffectiveDate());
+                        }
+                        else {
+                            invoiceItemEndDate = thisEvent.getPlanPhase().getDuration().addToLocalDate(roundedStartDate);
+                        }
+
+                    } catch (final CatalogApiException e) {
+                        log.warn("Error while computing invoice item end date");
```
sbrossie, line 461: Could we lof some details about the exception as well ?

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
```diff
@@ -176,19 +179,33 @@ void processFixedBillingEvents(final UUID invoiceId, final UUID accountId, final
...
+                prevItem = currentFixedPriceItem;
             }
 
-            final InvoiceItem currentFixedPriceItem = generateFixedPriceItem(invoiceId, accountId, thisEvent, targetDate, currency, invoiceItemGeneratorLogger, internalCallContext);
-            if (!isSameDayAndSameSubscription(prevItem, thisEvent, internalCallContext) && prevItem != null) {
+            currentEvent = nextEvent;
+        }
+
+        // Process the last event
+        if (currentEvent.getTransitionType() != BCD_CHANGE) {
```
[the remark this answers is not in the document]
reshmabidikar: Fixed

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
reshmabidikar: Fixed

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
```diff
@@ -176,19 +179,33 @@ void processFixedBillingEvents(final UUID invoiceId, final UUID accountId, final
         final InvoiceItemGeneratorLogger invoiceItemGeneratorLogger = new InvoiceItemGeneratorLogger(invoiceId, accountId, "fixed", log);
 
         final Iterator<BillingEvent> eventIt = events.iterator();
+        BillingEvent currentEvent = eventIt.next();
```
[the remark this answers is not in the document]
reshmabidikar: We are [checking](https://github.com/reshmabidikar/killbill/blob/4aff74751f888f4b27e576d7b818824454f640e5/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java#L172) if events is empty (`if (events.isEmpty())`), so when we retrieve `currentEvent`, it is sure to have at least one event.

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
sbrossie: It seems like the condition `nextEvent.getSubscriptionId().equals(thisEvent.getSubscriptionId()` is not necessary is we guarantee that this method does not take events, i.e. `thisEvent` and `nextEvent` that cross subscriptions. Can you add the `@Nullable` to the argument and remove this extra condition?

invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java
```diff
@@ -176,19 +179,24 @@ void processFixedBillingEvents(final UUID invoiceId, final UUID accountId, final
...
+
+        do {
+            BillingEvent nextEvent = eventIt.hasNext() ? eventIt.next() : null;
 
-            final InvoiceItem currentFixedPriceItem = generateFixedPriceItem(invoiceId, accountId, thisEvent, targetDate, currency, invoiceItemGeneratorLogger, internalCallContext);
-            if (!isSameDayAndSameSubscription(prevItem, thisEvent, internalCallContext) && prevItem != null) {
-                proposedItems.add(prevItem);
+            if (currentEvent.getTransitionType() != BCD_CHANGE) {
+                final BillingEvent adjustedNextEvent = (currentEvent.getSubscriptionId() == (nextEvent != null ? nextEvent.getSubscriptionId() : null)) ? nextEvent : null;
```
sbrossie, line 188: The double ternary statement is hard to follow but I agree it should work... except that checking for equality of UUID should use `equals()` and not `==` but it seems like this bad code is already [there for recurring items](https://github.com/killbill/killbill/blob/master/invoice/src/main/java/org/killbill/billing/invoice/generator/FixedAndRecurringInvoiceItemGenerator.java#L154). So I suggest:

1. Fix those in this branch 
2. Independently, let's open a small PR to fix the recurring case in `master` - which mean we may have a conflict ton resolve when me merge.

sbrossie (approved):

Issue #1944, opened 2024-01-10
Issue with catalog plugin API for optional fields


sbrossie:
It looks like although some fields are optional, i.e. could be null, our code throws an NPE when plugin implementor specifies null as shown [here](https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/plugin/StandaloneCatalogMapper.java#L118)

beccagaspard:
@sbrossie I noticed the following require an empty list and do not allow null:

`ProductImp.Builder::withAvailable`, `ProductImp.Builder::withIncluded`, `PlanImp.Builder::withInitialPhases`.

PR #2032, opened 2024-07-02
https://github.com/killbill/killbill/issues/1944

KBbitsP:


catalog/src/main/java/org/killbill/billing/catalog/plugin/StandaloneCatalogMapper.java
```diff
@@ -236,7 +236,7 @@ private Iterable<Product> toDefaultProducts(final Iterable<Product> input) {
     }
 
     private Collection<Product> toFilteredDefaultProduct(final Collection<Product> input) {
-        if (!input.iterator().hasNext()) {
+        if (input == null||!input.iterator().hasNext()) {
```
reshmabidikar, line 239: Nit: Missing space after `input == null`

catalog/src/main/java/org/killbill/billing/catalog/plugin/StandaloneCatalogMapper.java
```diff
@@ -282,7 +282,7 @@ private DefaultPriceListSet toDefaultPriceListSet(final PriceList defaultPriceLi
     }
 
     private DefaultPlanPhase[] toDefaultPlanPhases(final Iterable<PlanPhase> input) {
-        if (!input.iterator().hasNext()) {
+        if (input == null||!input.iterator().hasNext()) {
```
reshmabidikar, line 285: Nit: Missing space after input == null

reshmabidikar (commented):
Would it be possible to add some beatrix tests to confirm that the fix works?

sbrossie (commented):
What about [here](https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/plugin/StandaloneCatalogMapper.java#L227): Isn't it another use case where the input could be null that is not handled properly?

catalog/src/main/java/org/killbill/billing/catalog/plugin/StandaloneCatalogMapper.java
```diff
@@ -225,6 +225,9 @@ final void populateDefaultCaseChange(final CaseChange input, final DefaultCaseCh
     }
 
     private Iterable<Product> toDefaultProducts(final Iterable<Product> input) {
+        if (input == null) {
```
sbrossie, line 228: This method is called [here](https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/plugin/StandaloneCatalogMapper.java#L109C16-L109C27), and if you follow the code it seems like we would get an NPE... 

Can you return a `Collections.emptyList()` instead?

catalog/src/main/java/org/killbill/billing/catalog/plugin/StandaloneCatalogMapper.java
```diff
@@ -254,6 +257,9 @@ private Collection<Product> toFilteredDefaultProduct(final Collection<Product> i
     }
 
     private Iterable<Plan> toDefaultPlans(final StaticCatalog staticCatalog, final Iterable<Plan> input) {
+        if (input == null) {
+            return null;
```
sbrossie, line 261: Same remark as before. Can you return a `Collections.emptyList()` instead?

sbrossie (changes requested):

reshmabidikar:
Closing this PR, will open a separate PR with the fix.

Issue #1739, opened 2022-07-06
Charged Through Date Not updating for FixedTerm phase with fixed one time charges in case of IN_ADVANCE billing


KBbitsP:
Current behavior : 

By definition, the chargedthroughdate is associated with recurring subscriptions, that is it specifies the date up to which a user has been billed for in case of a recurring subscription. Currently we do not support updating it in case of  FIXEDTERM plans.

Expected behavior : 

CTD should also be updated for a FIXEDTERM phase with fixed one time charges in case of IN_ADVANCE billing? Since the user has already been invoiced for the duration of the FIXEDTERM? Currently, the CTD is not updated in such scenarios.

reshmabidikar:
**Work done so far (Code changes in the [work-for-1739](https://github.com/killbill/killbill/tree/work-for-1739) branch):** 

1. Code changes are made to update the CTD for `TRIAL`, `DISCOUNT` and `FIXEDTERM` phases. Some side effects of this:

- A `FIXED` invoice item generated corresponding to a subscription in a `TRIAL` phase will now have a non-null end date
- Cancellations requested without a date/cancellation policy will now be done `END_OF_TERM` if the subscription is  in the `TRIAL` phase (as against IMMEDIATELY as per the earlier behavior)
2. The above changes caused a lot of Java tests to fail. These are all fixed. 
3. This also caused e2e tests to fail. These are fixed in the [killbill-integration-tests/tests-for-1739](https://github.com/killbill/killbill-integration-tests/tree/tests-for-1739) branch.



**Work Remaining:**

There is a corner case related to the change alignment rule that does not work yet. The following issues exist:
1. [These e2e tests](https://github.com/killbill/killbill-integration-tests/blob/bb75f21a27972607f0cd312f3968f56b12cb1a4c/killbill-integration-tests/core/test_catalog.rb#L180:L274) fail after the changes to update the CTD for `TRIAL/DISCOUNT/FIXEDTERM`. These need to be investigated and fixed 
2. When an addon plan is changed with `START_OF_BUNDLE` change alignment rule, we would want the ADDON CTD to be aligned with the base CTD. However, this does not happen.

PR #2041, opened 2024-07-23
1739 tests

reshmabidikar:
Tests for change plan with `START_OF_BUNDLE` alignment

### DIFF
--- a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestAddonChangeAlignment.java
+++ b/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestAddonChangeAlignment.java
@@ -1,0 +1,352 @@
+/*
+ * Copyright 2020-2024 Equinix, Inc
+ * Copyright 2014-2024 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.killbill.billing.beatrix.integration;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import org.joda.time.LocalDate;
+import org.killbill.billing.account.api.Account;
+import org.killbill.billing.api.TestApiListener.NextEvent;
+import org.killbill.billing.catalog.api.BillingActionPolicy;
+import org.killbill.billing.catalog.api.PlanPhaseSpecifier;
+import org.killbill.billing.entitlement.api.DefaultEntitlementSpecifier;
+import org.killbill.billing.entitlement.api.Subscription;
+import org.killbill.billing.platform.api.KillbillConfigSource;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class TestAddonChangeAlignment extends TestIntegrationBase {
+
+    @Override
+    protected KillbillConfigSource getConfigSource(final Map<String, String> extraProperties) {
+        final Map<String, String> allExtraProperties = new HashMap<>(extraProperties);
+        allExtraProperties.put("org.killbill.catalog.uri", "catalogs/testAddonChangeAlignment");
+        return super.getConfigSource(null, allExtraProperties);
+    }
+
+    @Test(groups = "slow")
+    public void testChangeAlignmentStartOfBundleEOTChange() throws Exception {
+        final LocalDate today = new LocalDate(2023, 8, 1);
+        clock.setDay(today);
+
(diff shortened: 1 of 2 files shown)


beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestAddonChangeAlignment.java
```diff
@@ -0,0 +1,192 @@
...
+        //ADDON CTD is 2023-08-05 as expected
+        Subscription addonSub = subscriptionApi.getSubscriptionForEntitlementId(addonEntId, false, callContext);
+        Assert.assertEquals(addonSub.getChargedThroughDate(), new LocalDate(2023, 8, 05));
+
+        clock.setDay(new LocalDate(2023, 8, 15));
+
+        //change addon with start of bundle alignment on 2023-08-15 - change will happen immediately as subscription is in TRIAL phase
+        busHandler.pushExpectedEvents(NextEvent.CHANGE, NextEvent.INVOICE);
+        final PlanPhaseSpecifier newAddOnSpec = new PlanPhaseSpecifier("BasicAOStartOfBundle-monthly");
+        addonSub.changePlan(new DefaultEntitlementSpecifier(newAddOnSpec), Collections.emptyList(), callContext);
```
sbrossie, line 79: It's a little bit confusing to follow:  We started with `BasicAO-monthly` (which has a `START_OF_BUNDLE` alignment as this is the default case) and then we move to `BasicAOStartOfBundle-monthly` (which also has a `START_OF_BUNDLE` alignment, explicitly specified).

1. Is the goal to test a change of plan from 2 products both using `START_OF_BUNDLE` alignment?
2. If so, could you modify the naming/catalog as those 2 products have the same behavior (`START_OF_BUNDLE` alignments), but one uses default behavior and it is not reflected in the name while the second uses custom rule and has it specified in the name.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestAddonChangeAlignment.java
```diff
@@ -0,0 +1,192 @@
...
+        assertListenerStatus();
+
+        //ADDON CTD is 2023-08-05 as expected
+        Subscription addonSub = subscriptionApi.getSubscriptionForEntitlementId(addonEntId, false, callContext);
+        Assert.assertEquals(addonSub.getChargedThroughDate(), new LocalDate(2023, 8, 05));
+
+        clock.setDay(new LocalDate(2023, 8, 15));
+
+        //change addon with start of bundle alignment on 2023-08-15 - change will happen immediately as subscription is in TRIAL phase
+        busHandler.pushExpectedEvents(NextEvent.CHANGE, NextEvent.INVOICE);
```
sbrossie, line 77: Could we add an assertion below on the invoices: Interestingly we end up with 2 FIXED items for the same add-on subscriptions with a different start date - so let's assert on those as this is currently our 'normal' behavior.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestAddonChangeAlignment.java
```diff
@@ -0,0 +1,192 @@
...
+        Assert.assertEquals(baseSub.getChargedThroughDate(), new LocalDate(2023, 10, 1));
+
+        // ADDON CTD is 2023-10-01
+        addonSub = subscriptionApi.getSubscriptionForEntitlementId(addonEntId, false, callContext);
+        Assert.assertEquals(addonSub.getChargedThroughDate(), new LocalDate(2023, 10, 1));
+
+    }
+
+    @Test(groups = "slow")
+    public void testChangeAlignmentStartOfBundleImmediateChange() throws Exception {
```
sbrossie, line 119: Is it true to say that this test will produce exactly same result as IMMEDIATE policy and EOT currently default to immediate?

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestAddonChangeAlignment.java
reshmabidikar: > Is the goal to test a change of plan from 2 products both using START_OF_BUNDLE alignment?

The goal is to test a change of plan to a product that uses `START_OF_BUNDLE` alignment irrespective of the default case.  I've now also renamed the products in the catalog. Let me know if this is helpful.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestAddonChangeAlignment.java
reshmabidikar: > Could we add an assertion below on the invoices

Done.

> : Interestingly we end up with 2 FIXED items for the same add-on subscriptions with a different start date

This happens only in the test written on top of #1739. It does not happen in the tests written against `master`. So, the duplicate invoice item is a side effect of changes done for #1739.

beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestAddonChangeAlignment.java
reshmabidikar: > Is it true to say that this test will produce exactly same result as IMMEDIATE policy and EOT currently default to immediate?

Yes, that is correct, changing the plan with `IMMEDIATE` policy will behave in the same way as changing the plan with default `EOT` policy (since `EOT` defaults to `IMMEDIATE` in `TRIAL` phase)

PR #2049, opened 2024-08-23
invoice: Reimplement RAW_USAGE_DATE_COMPARATOR

sbrossie:
Reimplement RAW_USAGE_DATE_COMPARATOR to not rely on underlying `hashCode()` method. Using such method in the comparison may break the contract as there is no guarantee of unicity of a hashCode and the implementation conflicts with our equals() method.

The new implementation matches the internal `hashCode()` and `equals()` method from the `RawUsageRecord` which also looks at the 5 fields from the class.


reshmabidikar (approved):

sbrossie:
Merging as this was validated:

Screen Shot 2024-08-26 at 5 08 16 PM

PR #2064, opened 2024-10-24
Strengthen nodes info changes

vnandwana:
* Updated the logic where the current node is not found.
* Updated the node initialization logic, logging an error if a node with the same name already exists.


pierre:
Could you make the same change in the CI file as in https://github.com/killbill/killbill/pull/2062/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f, to make the e2e tests run?

pierre (approved):

PR #2071, opened 2024-12-16
Fix overrides bug

reshmabidikar:
There is a bug in the code due to which it is currently not possible to override a usage price by itself, it is necessary to update either a fixed or a recurring price too. This PR attempts to address this bug.

See [this comment](https://github.com/killbill/killbill-aviate-plugin/pull/287#discussion_r1886239492).

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResourceHelpers.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResourceHelpers.java
@@ -93,7 +93,8 @@ public class SubscriptionResourceHelpers {
         if (priceOverrides != null) {
             for (final PhasePriceJson input : priceOverrides) {
                 Preconditions.checkNotNull(input);
-
+                Preconditions.checkArgument(input.getFixedPrice() != null || input.getRecurringPrice() != null || input.getUsagePrices() != null || (input.getUsagePrices() != null && !input.getUsagePrices().isEmpty()),
+                                            "At least one fixed price, one recurring price or one usage price must be overridden");
                 final List<UsagePriceOverride> usagePrices = new LinkedList<>();
                 if (input.getUsagePrices() != null) {
                     buildUsagePrices(currency, input, usagePrices);
@@ -212,8 +213,6 @@ public class SubscriptionResourceHelpers {
         final PlanPhaseSpecifier planPhaseSpecifier = spec.getPlanName() != null ?
                                                       new PlanPhaseSpecifier(spec.getPlanName(), phaseType) :
                                                       new PlanPhaseSpecifier(spec.getProductName(), spec.getBillingPeriod(), spec.getPriceListName(), phaseType);
-        final Currency resolvedCurrency = input.getFixedPrice() != null || input.getRecurringPrice() != null || (input.getUsagePrices() != null && !input.getUsagePrices().isEmpty()) ? currency : null;
-
         return new PlanPhasePriceOverride() {
 
             @Override
@@ -228,7 +227,7 @@ public class SubscriptionResourceHelpers {
 
             @Override
             public Currency getCurrency() {
-                return resolvedCurrency;
+                return currency;
             }
 
             @Override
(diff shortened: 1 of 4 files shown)


reshmabidikar:
If you agree that this is a bug, I can write a test for this fix.

reshmabidikar:
For 2/- above, I've added a test  See c29ce39. 
For 1/ above, I was not sure where to throw an exception. So I've added a validation to check that there is at least one overridden price and a test for this. With this change, there is now no need for [this check](https://github.com/killbill/killbill/blob/d7ff936609bc54a5129c080e5c450b781786d11c/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResourceHelpers.java#L215). See e941bba. If this is not what you meant please let me know and I can make changes accordingly.

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestEntitlement.java
```diff
@@ -2049,4 +2053,61 @@ public Boolean call() throws Exception {
...
+                                           ExtBusEventType.SUBSCRIPTION_CREATION);
+
+        final Subscription entitlementJson = subscriptionApi.createSubscription(input,
+                                                                                null,
+                                                                                (LocalDate) null,
+                                                                                NULL_PLUGIN_PROPERTIES,
+                                                                                requestOptions);
+        callbackServlet.assertListenerStatus();
+        final Subscription subscription = subscriptionApi.getSubscription(entitlementJson.getSubscriptionId(), requestOptions);
+        Assert.assertEquals(subscription.getState(), EntitlementState.ACTIVE);
```
sbrossie, line 2145: Any way we could check that price was correctly overridden - would `subscription#getLastActivePlan()` provide such info?

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResourceHelpers.java
```diff
@@ -93,7 +93,8 @@ public static List<PlanPhasePriceOverride> buildPlanPhasePriceOverrides(final It
         if (priceOverrides != null) {
             for (final PhasePriceJson input : priceOverrides) {
                 Preconditions.checkNotNull(input);
-
+                Preconditions.checkArgument(input.getFixedPrice() != null || input.getRecurringPrice() != null || input.getUsagePrices() != null || (input.getUsagePrices() != null && !input.getUsagePrices().isEmpty()),
```
sbrossie, line 96: I think we can remove the  `input.getUsagePrices() != null` and simply keep the `(input.getUsagePrices() != null && !input.getUsagePrices().isEmpty()`

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResourceHelpers.java
```diff
@@ -212,8 +213,6 @@ public List<UsagePriceOverride> getUsagePriceOverrides() {
         final PlanPhaseSpecifier planPhaseSpecifier = spec.getPlanName() != null ?
                                                       new PlanPhaseSpecifier(spec.getPlanName(), phaseType) :
                                                       new PlanPhaseSpecifier(spec.getProductName(), spec.getBillingPeriod(), spec.getPriceListName(), phaseType);
-        final Currency resolvedCurrency = input.getFixedPrice() != null || input.getRecurringPrice() != null || (input.getUsagePrices() != null && !input.getUsagePrices().isEmpty()) ? currency : null;
```
sbrossie: Good point, I don't think this is needed.

sbrossie (approved):

profiles/killbill/src/test/java/org/killbill/billing/jaxrs/TestEntitlement.java
reshmabidikar: Addressed in 984a82d

jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/SubscriptionResourceHelpers.java
reshmabidikar: Addressed in 984a82d

PR #2073, opened 2025-01-03
Invoice index to fix slow loading of invoices

reshmabidikar:


invoice/src/main/resources/org/killbill/billing/invoice/ddl.sql
```diff
@@ -168,7 +168,7 @@ CREATE TABLE invoice_history (
 CREATE INDEX invoice_history_target_record_id ON invoice_history(target_record_id);
 CREATE INDEX invoice_history_tenant_record_id ON invoice_history(tenant_record_id);
 CREATE INDEX invoice_history_currency ON invoice_history(currency ASC);
-
+CREATE INDEX invoice_history_account_record_id_record_id ON invoices(account_record_id, record_id);
```
pierre, line 171: Should be `invoice_history`?

pierre (changes requested):

invoice/src/main/resources/org/killbill/billing/invoice/ddl.sql
reshmabidikar: Addressed in 51f5fe6.

pierre (approved):

PR #2081, opened 2025-01-18
invoice: Initial refactoring work to extract transition times from usage code.

sbrossie:
The intent is to be able to communicate these details to a (usage) plugin to provide more flexibility and possibly have the plugin aggregate usage points in a way that will be compatible with what Kill Bill usage module expects.

In [7cea0bd](https://github.com/killbill/killbill/pull/2081/commits/7cea0bde32c2dd1e8d88a5f6d05815b5dc87ff8d), this is purely refactoring to be able to extract the computation of transitions from usage module from the usage item generation.


sbrossie:
Closing, and recreated in https://github.com/killbill/killbill/pull/2082

PR #2083, opened 2025-01-20
Fix for aviate plugin 336

reshmabidikar:
Fix for https://github.com/killbill/killbill-aviate-plugin/issues/336

### DIFF
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
@@ -764,7 +764,7 @@ public class DefaultEntitlement extends EntityBase implements Entitlement {
         final BaseEntitlementWithAddOnsSpecifier baseEntitlementWithAddOnsSpecifier = new DefaultBaseEntitlementWithAddOnsSpecifier(
                 getBundleId(),
                 getBundleExternalKey(),
-                List.of(new DefaultEntitlementSpecifier(null, null, null, getExternalKey(), null)),
+                List.of(new DefaultEntitlementSpecifier(spec.getPlanPhaseSpecifier(), spec.getBillCycleDay(), spec.getQuantity(), getExternalKey(), spec.getOverrides())),
                 null,
                 null,
                 false);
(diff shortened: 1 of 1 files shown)


entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java
```diff
@@ -764,7 +764,7 @@ public Entitlement changePlanOverrideBillingPolicy(final EntitlementSpecifier sp
         final BaseEntitlementWithAddOnsSpecifier baseEntitlementWithAddOnsSpecifier = new DefaultBaseEntitlementWithAddOnsSpecifier(
                 getBundleId(),
                 getBundleExternalKey(),
-                List.of(new DefaultEntitlementSpecifier(null, null, null, getExternalKey(), null)),
+                List.of(new DefaultEntitlementSpecifier(spec.getPlanPhaseSpecifier(), spec.getBillCycleDay(), spec.getQuantity(), getExternalKey(), spec.getOverrides())),
```
sbrossie, line 767: Am i understanding correctly that we were not passing correctly the details of the `EntitlementSpecifier` to the entitlement plugins for change operations - in this case should we also fix the other methods `changePlanWithDate`, ... ?

reshmabidikar: > Am i understanding correctly that we were not passing correctly the details of the EntitlementSpecifier to the entitlement plugins for change operations 

Yes that is correct. Yes, I think we need to fix the other `changePlanWithXXX` methods.  I'm also wondering if we should make this change in the  `cancelPlanWithXXX` methods? For example [here](https://github.com/killbill/killbill/blob/8d66b2267667b1215f042981e0832e4f82676513/entitlement/src/main/java/org/killbill/billing/entitlement/api/DefaultEntitlement.java#L344)?

sbrossie: We should pass all the info to the plugin through the  EntitlementContext `pluginContext`. In the case of cancelation, there is no `spec` though, so not clear that we are missing anything. If you see anything missing, please add it. Definitely needed for other `changePlanWithXXX`.

reshmabidikar: Taking a look again, you are right, there is no need to make changes to the `cancelXXX` methods. I've made changes to all the `changePlanWithXXX` methods in 92bc89c.

sbrossie (approved):

PR #2098, opened 2025-02-11
Export extra tables

reshmabidikar:
Added feature to export extra tables to address https://github.com/killbill/killbill-aviate-plugin/issues/354.  The `org.killbill.export.extra.tables.prefix` property needs to be set to the prefix of the tables that need to be exported. For the sake of simplicity, we are assuming that the extra table have the `account_id` and `tenant_id` columns. 

With regards to the ci failure, there is some leftover data in case of H2. So I've disabled some assertions for now. 

Finally, there are a lot of intermediate commits for the purpose of debugging the ci failure. So I would suggest we squash and merge this PR.

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/config/definition/ExportConfig.java
+++ b/util/src/main/java/org/killbill/billing/util/config/definition/ExportConfig.java
@@ -17,6 +17,8 @@
 
 package org.killbill.billing.util.config.definition;
 
+import java.util.List;
+
 import org.skife.config.Config;
 import org.skife.config.Default;
 import org.skife.config.Description;
@@ -24,8 +26,8 @@ import org.skife.config.Description;
 public interface ExportConfig extends KillbillConfig {
 
     @Config("org.killbill.export.extra.tables.prefix")
-    @Default("")
+    @Default("aviate_catalog")
     @Description("Prefix of the extra tables that need to be imported")
-    String getExtraTablesPrefix();
+    List<String> getExtraTablesPrefix();
 
 }
--- a/util/src/main/java/org/killbill/billing/util/config/definition/ExportConfig.java
+++ b/util/src/main/java/org/killbill/billing/util/config/definition/ExportConfig.java
@@ -23,9 +23,9 @@ import org.skife.config.Description;
 
 public interface ExportConfig extends KillbillConfig {
 
-    @Config("org.killbill.export.aviateCatalogTablesIncluded")
-    @Default("false")
-    @Description("Whether to include aviate catalog tables in export")
-    boolean aviateCatalogTablesIncluded();
+    @Config("org.killbill.export.extra.tables.prefix")
+    @Default("")
+    @Description("Prefix of the extra tables that need to be imported")
+    String getExtraTablesPrefix();
 
 }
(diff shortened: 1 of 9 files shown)


util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
```diff
@@ -160,24 +173,39 @@ private void exportDataForAccountAndTable(final DatabaseExportOutputStream out,
...
-
+        final TableType finalTableType = tableType;
         dbi.withHandle(new HandleCallback<Void>() {
             @Override
             public Void withHandle(final Handle handle) throws Exception {
                 final ResultIterator<Map<String, Object>> iterator = handle.createQuery(queryBuilder.toString())
-                                                                           .bind("accountRecordId", context.getAccountRecordId())
-                                                                           .bind("tenantRecordId", context.getTenantRecordId())
+                                                                           .bind("accountRecordId", finalTableType == TableType.EXTRA ? accountId : context.getAccountRecordId())
+                                                                           .bind("tenantRecordId", finalTableType == TableType.EXTRA ? tenantId : context.getTenantRecordId())
```
pierre, line 208: This doesn't sound right: `tenantId` vs `tenantRecordId`?

util/src/main/java/org/killbill/billing/util/config/definition/ExportConfig.java
```diff
@@ -0,0 +1,31 @@
...
+import org.skife.config.Config;
+import org.skife.config.Default;
+import org.skife.config.Description;
+
+public interface ExportConfig extends KillbillConfig {
+
+    @Config("org.killbill.export.extra.tables.prefix")
+    @Default("")
+    @Description("Prefix of the extra tables that need to be imported")
+    String getExtraTablesPrefix();
```
pierre, line 29: That supports only one prefix, right? How about returning a `List<String>`, so we support multiple and setting `@Default("aviate")` as the default?

util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
reshmabidikar: `tenantRecordId` is just the name of the parameter, the actual column name is `tenant_id` in case of the `aviate_catalog_xxx` tables. See [this line](https://github.com/reshmabidikar/killbill/blob/6465cc34592e4edaf9c14568e6d2d277f4c5fa4f/util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java#L76). See also [this comment](https://github.com/killbill/killbill-aviate-plugin/issues/354#issuecomment-2659996778) and [this slack thread](https://killbillio.slack.com/archives/G01C2P3V8G7/p1739799878262709).

util/src/main/java/org/killbill/billing/util/config/definition/ExportConfig.java
reshmabidikar: Addressed in 0a02711. I've set the default to `aviate_catalog`. If the default is `aviate`, the code will try to export data from all the aviate tables (like `aviate_billing_accounts`, etc.) and if these tables do not have the `tenant_id`/`account_id` columns, an exception will occur.

PR #2105, opened 2025-02-26
2012 - Fix Regression in KillBillJndiLdapRealm

reshmabidikar:


util/src/main/java/org/killbill/billing/util/security/shiro/realm/KillBillJndiLdapRealm.java
```diff
@@ -190,27 +193,28 @@ private Set<String> findLDAPGroupsForUser(final String userName, final LdapConte
...
-                        .collect(Collectors.toUnmodifiableSet());
+        final Collection<String> groups = Iterators.toStream(attributesIterator)
+                                                   // Find the attribute representing the group name
+                                                   .filter(attribute -> groupNameId.equalsIgnoreCase(attribute.getID()))
+                                                   .flatMap(groupNameAttribute -> {
+                                                 try {
+                                                     final NamingEnumeration<?> enumeration = groupNameAttribute.getAll();
+                                                     final List<?> enumerationAsList = Collections.list(enumeration);
+                                                     final List<String> values = new ArrayList<>();
+                                                     enumerationAsList.forEach(e -> values.addAll((Collection) e));
```
pierre, line 208: I'm concerned about the cast here: we don't really know what `enumeration` is (could be LDAP implementation specific?).

How about?
```
return enumerationAsList.stream().flatMap(value -> {
            if (value instanceof Collection) {
                return ((Collection<?>) value).stream().map(Object::toString);
            } else {
                return Stream.of(value.toString());
            }
        });
```

Issue #2116, opened 2025-05-02
Catalog initialization from plugin leads to lots of contention


sbrossie:
The creation of the catalog in Kill Bill from the catalog plugin API relies on reflection to populate some object fields. The code currently makes sure that any un-required field is correctly initialized - e.g. every array is allocated with a zero size - and this code currently relies on some `synchronized` statements - e.g. [here](https://github.com/killbill/killbill/blob/master/catalog/src/main/java/org/killbill/billing/catalog/CatalogSafetyInitializer.java#L122) - that end up creating a lot of contention when there are many entries and catalog calls in parallel:

```
"notifications-th" #77 prio=5 os_prio=0 cpu=6102.14ms elapsed=664.14s tid=0x0000fffe60010000 nid=0x61 waiting for monitor entry  [0x0000fffe8c3fc000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.killbill.billing.catalog.CatalogSafetyInitializer.initializeArrayIfNull(CatalogSafetyInitializer.java:132)
	- waiting to lock <0x00000000820b1470> (a java.util.HashMap)
	at org.killbill.billing.catalog.CatalogSafetyInitializer.initializeNonRequiredNullFieldsWithDefaultValue(CatalogSafetyInitializer.java:56)
	at org.killbill.billing.catalog.DefaultPlanPhase.initialize(DefaultPlanPhase.java:196)
	at org.killbill.billing.catalog.DefaultPlan.initialize(DefaultPlan.java:274)
	at org.killbill.billing.catalog.plugin.StandaloneCatalogMapper.toDefaultPlan(StandaloneCatalogMapper.java:360)
	at org.killbill.billing.catalog.plugin.StandaloneCatalogMapper.toDefaultPlans(StandaloneCatalogMapper.java:260)
	at org.killbill.billing.catalog.plugin.StandaloneCatalogMapper.toStandaloneCatalog(StandaloneCatalogMapper.java:110)
	at org.killbill.billing.catalog.plugin.VersionedCatalogMapper.toStandaloneCatalog(VersionedCatalogMapper.java:43)
	at org.killbill.billing.catalog.plugin.VersionedCatalogMapper.toVersionedCatalog(VersionedCatalogMapper.java:36)
	at org.killbill.billing.catalog.caching.DefaultCatalogCache.getCatalogFromPlugins(DefaultCatalogCache.java:179)
[...]

PR #2118, opened 2025-05-05
Issue  #2116: Catalog initialization from plugin leads to lots of contention

nick-at-finix:
I cannot think of any straightforward way to test this change with respect to the contention that we experienced. Luckily, the changes are fairly benign and take advantage of some Java 9+ features that did not exist when this class was written.

### DIFF
--- a/catalog/src/main/java/org/killbill/billing/catalog/CatalogSafetyInitializer.java
+++ b/catalog/src/main/java/org/killbill/billing/catalog/CatalogSafetyInitializer.java
@@ -20,9 +20,11 @@ package org.killbill.billing.catalog;
 import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.math.BigDecimal;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlElementWrapper;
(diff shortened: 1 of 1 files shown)


catalog/src/main/java/org/killbill/billing/catalog/CatalogSafetyInitializer.java
```diff
@@ -119,30 +119,28 @@ private static LinkedList<Field> initializeNonRequiredFields(final Class<?> aCla
     }
 
     private static void initializeFieldWithValue(final Object obj, final Field f, final Object value) throws IllegalAccessException, ClassNotFoundException {
-        synchronized (perCatalogClassNonRequiredFields) {
-            f.setAccessible(true);
+        if (f.trySetAccessible()) {
```
nick-at-finix, line 122: Java 9's trySetAccessible restricts the scope of the accessibility change to the given method only. There is no need to reset the accessibility of the field.

catalog/src/main/java/org/killbill/billing/catalog/CatalogSafetyInitializer.java
```diff
@@ -33,23 +35,23 @@
...
 public class CatalogSafetyInitializer {
 
-
     public static final Integer DEFAULT_NON_REQUIRED_INTEGER_FIELD_VALUE = -1;
     public static final Double DEFAULT_NON_REQUIRED_DOUBLE_FIELD_VALUE = (double) -1;
     public static final BigDecimal DEFAULT_NON_REQUIRED_BIGDECIMAL_FIELD_VALUE = new BigDecimal("-1");
 
-    private static final Map<Class<?>, LinkedList<Field>> perCatalogClassNonRequiredFields = new HashMap<>();
+    private static final Map<Class<?>, LinkedList<Field>> perCatalogClassNonRequiredFields = new ConcurrentHashMap<>();
+    private static final Map<Class<?>, Object> zeroLengthArrayCache = new ConcurrentHashMap<>();
```
nick-at-finix, line 43: There isn't that many types of arrays in the catalog classes, there is no need to keep creating new objects.

PR #2119, opened 2025-05-05
Fix for 2116

sbrossie:
Replace synchronized statement on a per field level with a `ConcurrentHashMap` around the main `perCatalogClassNonRequiredFields` map.


catalog/src/main/java/org/killbill/billing/catalog/CatalogSafetyInitializer.java
```diff
@@ -20,9 +20,11 @@
 import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.math.BigDecimal;
+import java.util.Deque;
```
nick-at-finix, line 23: This is unused, my bad here

Issue #2124, opened 2025-05-16
StandaloneCatalogMapper does not handle well empty catalog


sbrossie:
If a catalog plugin returns a catalog version where all plans where retired i.e it contains no entries  - see screenshot below - we end up with a NPE:

<img width="457" alt="Image" src="https://github.com/user-attachments/assets/a84a15a1-5805-4860-9919-d55768780580" />

```
java.lang.NullPointerException: null
	at org.killbill.billing.catalog.DefaultPriceListSet.initialize(DefaultPriceListSet.java:123)
	at org.killbill.billing.catalog.StandaloneCatalog.initialize(StandaloneCatalog.java:318)
	at org.killbill.billing.catalog.plugin.StandaloneCatalogMapper.toStandaloneCatalog(StandaloneCatalogMapper.java:122)
	at org.killbill.billing.catalog.plugin.VersionedCatalogMapper.toStandaloneCatalog(VersionedCatalogMapper.java:43)
	at org.killbill.billing.catalog.plugin.VersionedCatalogMapper.toVersionedCatalog(VersionedCatalogMapper.java:36)
	at org.killbill.billing.catalog.caching.DefaultCatalogCache.getCatalogFromPlugins(DefaultCatalogCache.java:179)
	at org.killbill.billing.catalog.caching.DefaultCatalogCache.getCatalog(DefaultCatalogCache.java:106)
	at org.killbill.billing.catalog.DefaultCatalogService.getCatalog(DefaultCatalogService.java:113)
	at org.killbill.billing.catalog.DefaultCatalogService.getFullCatalog(DefaultCatalogService.java:104)
	at org.killbill.billing.catalog.api.user.DefaultCatalogUserApi.getCatalog(DefaultCatalogUserApi.java:97)
	at org.killbill.billing.util.glue.KillbillApiAopModule$ProfilingMethodInterceptor$1.execute(KillbillApiAopModule.java:89)
	at org.killbill.commons.profiling.Profiling.executeWithProfiling(Profiling.java:35)
	at org.killbill.billing.util.glue.KillbillApiAopModule$ProfilingMethodInterceptor.invoke(KillbillApiAopModule.java:105)
	at org.killbill.billing.jaxrs.resources.CatalogResource.getCatalogVersions(CatalogResource.java:297)

```

PR #2125, opened 2025-05-23
Fix NPE in catalog mapper. See #2124

sbrossie:


sbrossie:
> One test failure, otherwise :shipit:

Looking at test failure forced me to change the fix. Looks this is working now

Issue #2133, opened 2025-06-18
Invoice generated incorrectly when a subscription is created on the last day of the month.


reshmabidikar:
**Steps to reproduce:**

1. Create a subscription bundle with 1 base and 1 addon with usage (both `IN_ARREAR`) on `2025-03-31`
2. Move clock to `2025-04-30` - Invoice is generated as expected
3. Record usage for the addon with `date=2025-05-15`
4. Move clock to `2025-05-30` and add another addon - Invoice is generated with invoice items for base and addon1 - This is unexpected
5. Schedule plan change for addon1 with `date=2025-05-31`
6. Move to `2025-05-31` - Invoice is generated with usage item for addon1 as well as prorated items for addon2. This is unexpected.

**Expected behavior:** Invoice should not be generated on `2025-05-30`. It should be generated on `2025-05-31` and should include recurring+usage items for addon1.

[This test](https://github.com/reshmabidikar/killbill/blob/a9f9742a04def8386b72699f8d00b01b23e5046a/beatrix/src/test/java/org/killbill/billing/beatrix/integration/TestInArrearChangeCancel.java#L132) reproduces the behavior. It is present in the [addon-last-day-of-month](https://github.com/reshmabidikar/killbill/tree/addon-last-day-of-month) branch

shaun-forgie:
Given this issue was addressed in the https://github.com/killbill/killbill/releases/tag/killbill-0.24.15 release should it not be closed?

reshmabidikar:
Thanks for pointing this out,, closing this ticket.

PR #2134, opened 2025-06-25
#2133 - Fix for addon on last day of month

reshmabidikar:
#2133 - Fix for addon on last day of month


sbrossie (commented):
I reuse your original branch and created another fix in https://github.com/killbill/killbill/pull/2136.

Initially, I wanted to implement the way you have it, i.e. modifying `advanceByNPeriods`, but then I realized it does not work, as we don't have the bcd as an input in that method - which is required to decide whether we need to align with end of month or not.

Then I found the wrapper method we needed already existed , i.e. `BillingIntervalDetail#getFutureBillingDateFor`, and so it was mater of using it in all places where we use 'advanceByNPeriods'.

I suggest we close this PR.

Issue #364, opened 2015-08-10
Janitor: Re-introduce the janitor loop for UNKNOWN payment transactions


sbrossie:
The `IncompletePaymentTransactionTask` relies on bus events to handle `UNKNOWN` or `PENDING` transactions; however there are cases (for e.g if the system shuts off during a payment transaction) where no event will be sent as `UNKNOWN`  is the original transaction status.

The idea is to re-introduce the janitor loop to take care of those.

sbrossie:
This is such an edge case, we should assess how much work is needed to do it well, if too much probably punt. I think the tricky piece is to stream all the UNKNOWN and also make sure we don't compete with current UNKNOWN handled by bus event.

PR #2140, opened 2025-06-27
Work for 364 - 

reshmabidikar:
#364 

- Ignore extra tables that do not have accountId/tenantId columns
- Handle aviate notifications table.

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
+++ b/util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
@@ -257,10 +257,10 @@ public class DatabaseExportDao {
         boolean accountIdColPresent = false;
         boolean tenantIdColPresent = false;
         for(ColumnInfo column : columnsForTable) {
-            if(!tenantIdColPresent && tableType == TableType.EXTRA && column.getColumnName().equals(tableType.getTenantRecordIdColumnName())) {
+            if(!tenantIdColPresent && tableType == TableType.EXTRA && column.getColumnName().equalsIgnoreCase(tableType.getTenantRecordIdColumnName())) {
                 tenantIdColPresent = true;
             }
-            if(!accountIdColPresent && tableType == TableType.EXTRA && column.getColumnName().equals(tableType.getAccountRecordIdColumnName())) {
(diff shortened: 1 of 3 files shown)


util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
```diff
@@ -166,14 +166,21 @@ private void exportDataForAccountAndTable(final DatabaseExportOutputStream out,
                     tableType = TableType.NOTIFICATION;
                 }
             }
+            if(tableType == TableType.EXTRA && column.getColumnName().equalsIgnoreCase(TableType.NOTIFICATION.getAccountRecordIdColumnName())) { //for aviate_notifications
```
sbrossie, line 169: Just a thought: If we were to change line 161 and handle both `if (tableType == TableType.OTHER || tableType == TableType.EXTRA )`, wouldn't we not get this for free, essentially allowing any `notifications` table from any plugins that is using notifications tables (e.g. analytics) ?

util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
```diff
@@ -246,5 +253,18 @@ public Void withHandle(final Handle handle) throws Exception {
         });
     }
 
+    private boolean areAccountIdTenantIdColsPresent(List<ColumnInfo> columnsForTable, TableType tableType) {
+        boolean accountIdColPresent = false;
+        boolean tenantIdColPresent = false;
+        for(ColumnInfo column : columnsForTable) {
```
sbrossie, line 259: We already did a loop through column name in line 152, could we leverage the same loop - or do we need to do it twice?

util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
reshmabidikar: Addressed in 1c08594

util/src/main/java/org/killbill/billing/util/export/dao/DatabaseExportDao.java
reshmabidikar: Addressed in 05d1e14.

sbrossie (approved):

PR #2169, opened 2025-10-20
killbill/killbill-platform#126 - add dedicated filter to bypass enableResponseBuffering

xsalefter:
When Jersey writing a response, it try to set Content-Length header before writing the body [1]. While this is normal for non text/stream response, it breaks SSE since data the data need to flow as a continuous stream.

`EventStreamResponseFilter` solves that incompatibility. It detects text/event-stream responses and wraps Jersey’s `ContainerResponseWriter` so `enableResponseBuffering()` always returns false. The rest of the pipeline stays untouched.

But in the end, this PR alone wont fix killbill/killbill-platform#126. We need another fix there.

---

_[1] : See the [source code](https://github.com/eclipse-ee4j/jersey/blob/7ea389305d3826ccd6e209eddf45486f58723de0/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/internal/ResponseWriter.java#L248). But I'm wondering if I'm pointed at the wrong source code. Because if this is indeed the source of truth about `enableResponseBuffering()` that simply return `true`, then how Jersey's set `ServerProperties.OUTBOUND_CONTENT_LENGTH_BUFFER` configuration into account ?_

### DIFF
--- a/profiles/killbill/src/test/java/org/killbill/billing/server/filters/TestEventStreamResponseFilter.java
+++ b/profiles/killbill/src/test/java/org/killbill/billing/server/filters/TestEventStreamResponseFilter.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2020-2025 Equinix, Inc
+ * Copyright 2014-2025 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.killbill.billing.server.filters;
+
+import java.io.ByteArrayOutputStream;
+import java.io.OutputStream;
+import java.net.URI;
+import java.util.concurrent.TimeUnit;
+
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+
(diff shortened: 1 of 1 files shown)


xsalefter:
About,

> Regarding the fix itself, could you set a breakpoint when doing regular request and doing SSE to check that it works as expected?

I hope video below is enough: 

- 00.00 : server started
- 00.25 : Set breakpoint in Intellij
- 00.29 : Perform regular request to `/catalog/versions` . The `this.eventStreamRequest` variable is `false`, and in `00.38`, it just execute line `return this.delegate.enableResponseBuffering();`
- 00.58 : Perform call to SSE endpoint `/plugins/killbill-osgi-logger/`. The `this.eventStreamRequest` variable is `true`, and in `01.07` it just return false without execute the `this.delegate.enableResponseBuffering();` line.

[platform-126-normal-request-vs-sse.webm](https://github.com/user-attachments/assets/b02944b0-3939-412a-ab02-3ea2e25b0dc4)

xsalefter:
Now in this section I'll explain about another error : `.... java.lang.IllegalStateException: s=OPEN,api=BLOCKED,sc=false,e=null ....`. In this video, I run Kill Bill server with `xsalefter:platform-126-add-eventstreamfilter` branch.

- 00.00 : Get compiled at the target branch, and show branch name.
- 00.08 : Run Kill Bill
- 00.32 : Server run and ready
- 00.38 : Call SSE endpoint via `curl`: `curl -N -H 'Accept: text/event-stream' --trace-ascii /dev/stderr --trace-time http://localhost:8080/plugins/killbill-osgi-logger/`
- 00.40 : `curl` not exit anymore. This is because `enableResponseBuffering()` now just return false.
- 00.55 : Server log still normal (no exception)
- 01.00 : Call `/catalog/versions` to make sure non heartbeat message sent to client (curl)
- 01:21 : Prepare to call SSE endpoint from new terminal tab.
- 01:30 : Connection established, the 2nd tab not exit, **but,**
- 01:35 : There's another error introduced: `java.lang.IllegalStateException: s=OPEN,api=BLOCKED,sc=false,e=null`

[platform-126-with-event-filter-branch-error.webm](https://github.com/user-attachments/assets/e3c2e08c-da8e-4da6-ac02-7aa0c63aac82)

PR #2201, opened 2026-04-20
Add requestedDate parameter in CreditResource

reshmabidikar:


sbrossie (approved):
Looks good - but maybe as a next step, let's release the java client as per [this PR](https://github.com/killbill/killbill-client-java/pull/146), and add a  `profiles/killbill/` test to verify the new query parameter?

Issue #124, opened 2013-11-18
Add-Ons subscriptions and entitlements not cancelled doing transfer


pierre:
Looking at `DefaultSubscriptionBaseTransferApi`, it doesn't look like we cancel add-ons (line 225). We also need to cancel entitlements.

I'll cook a patch.

PR #2254, opened 2026-06-15
[Entitlement] Expose BCD_UPDATE and QUANTITY_UPDATE events in subscription event stream

### LINKED ISSUE #2224: BCD_UPDATE event does not seem to be returned from subscription events
Retrieving a subscription using the [retrieve-a-subscription-by-id](https://apidocs.killbill.io/subscription#retrieve-a-subscription-by-id) returns all the fields including the subscription events in the `events` field. However, it looks like it does not include possible `BCD_UPDATE` events.

**Tasks:**

1/ Looks at the code to understand why this is not included 
2/ Make an assessment - was there a good reason?
3/ If, this is deemed to be a bug, let's fix it

jutur:
Fixes #2224.

`SubscriptionEventOrdering.toEventTypes()` fell through to the default (empty list) branch for `BCD_CHANGE` and `QUANTITY_CHANGE` transitions, so callers of `Subscription.getSubscriptionEvents()` never saw these events in the returned list — only `PHASE`, `CHANGE`, `START_BILLING`, and `STOP_BILLING` were mapped.

Add explicit case branches for `BCD_CHANGE → BCD_UPDATE` and `QUANTITY_CHANGE → QUANTITY_UPDATE`, and register both new types under `ENT_BILLING_SERVICE_NAME` in `EntitlementOrderingBase.getServiceName()`, consistent with `PHASE` and `CHANGE`.

Also explicitly lists `UNDO_CHANGE` in the ignored-event cases for clarity.

**Depends on:** killbill/killbill-api#<number> (adds `BCD_UPDATE` and `QUANTITY_UPDATE` to `SubscriptionEventType`)

A test in `TestDefaultSubscriptionApi` verifies that a `BCD_UPDATE` event appears in the subscription event stream after calling `updateBCD()`.

### DIFF
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/EntitlementOrderingBase.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/EntitlementOrderingBase.java
@@ -45,6 +45,8 @@ public abstract class EntitlementOrderingBase {
 
             case PHASE:
             case CHANGE:
+            case BCD_UPDATE:
+            case QUANTITY_UPDATE:
                 return ENT_BILLING_SERVICE_NAME;
 
             default:
--- a/entitlement/src/main/java/org/killbill/billing/entitlement/api/SubscriptionEventOrdering.java
+++ b/entitlement/src/main/java/org/killbill/billing/entitlement/api/SubscriptionEventOrdering.java
@@ -95,11 +95,16 @@ public class SubscriptionEventOrdering extends EntitlementOrderingBase {
                 return List.of(SubscriptionEventType.STOP_BILLING);
             case PHASE:
                 return List.of(SubscriptionEventType.PHASE);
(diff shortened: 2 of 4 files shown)


jutur:
CI is failing because this PR depends on killbill-api #124, which adds BCD_UPDATE and QUANTITY_UPDATE to SubscriptionEventType. The pom.xml here pins killbill-api.version to 0.54.1-SNAPSHOT (the local dev build of that change) — once killbill-api #124 is merged and 0.54.1 is released, I'll update the version here and CI should pass.

reshmabidikar:
@jutur regarding [this comment](https://github.com/killbill/killbill/pull/2254#issuecomment-4725735798) are you sure ci is failing due to https://github.com/killbill/killbill/issues/124? This seems like a very old ticket, and it is closed.

PR #2260, opened 2026-06-17
usage: expose billing transition times as a typed accessor on UsageCo…

### LINKED ISSUE #2257: Enhance usage plugin Api to pass all transitions
We introduced a plugin property [USAGE_TRANSITIONS](https://github.com/killbill/killbill/blob/killbill-0.24.18/invoice/src/main/java/org/killbill/billing/invoice/generator/UsageInvoiceItemGenerator.java#L76) to pass such transitions. Instead we should pass these as a parameter in our plugin api.

jutur:
- Adds a usageTransitions field to DefaultUsageContext and overrides getUsageTransitions() from the new UsageContext interface method
- DefaultInternalUserApi now extracts the transition map from the USAGE_TRANSITIONS plugin property and passes it to DefaultUsageContext, making it available to usage plugins via the typed accessor
- Existing plugins that read the USAGE_TRANSITIONS plugin property continue to work unchanged

Depends on killbill-plugin-api#<plugin-api-PR> which adds the default getUsageTransitions() method to UsageContext.

Fixes #2257

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,6 +62,7 @@
         <check.skip-dependency-versions>true</check.skip-dependency-versions>
         <check.spotbugs-exclude-filter-file>${main.basedir}/spotbugs-exclude.xml</check.spotbugs-exclude-filter-file>
         <killbill.version>${project.version}</killbill.version>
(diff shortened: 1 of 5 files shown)


sbrossie:
Please reach out to us prior submitting PRs.

PR #2268, opened 2026-07-10
jaxrs: clarify billingStartDate is read-only in the subscription JSON

### LINKED ISSUE #2189: Inconsistency in createSubscriptionWithAddOns: billingDate vs billingStartDate
**Description**
There is a discrepancy between the documented query parameters and the actual implementation for the createSubscriptionWithAddOns endpoint. While the standard createSubscription endpoint (and various documentation sources) suggests using billingDate, this endpoint rejects that field and instead requires billingStartDate.

**Steps to Reproduce**
    - Attempt to call the following endpoint using billingDate as a query parameter: `POST /1.0/kb/subscriptions/createSubscriptionWithAddOns?billingDate=2023-10-27`
    - Observe the error response indicating that billingDate is an unknown query field.
    - Change the query parameter to billingStartDate and notice the call succeeds.

**Expected Behavior**
The endpoint should accept billingDate to remain consistent with the standard createSubscription endpoint and current API documentation. At the very least, the documentation should be updated to reflect that this specific composite endpoint uses a different parameter name.

**Actual Behavior**
    - Parameter billingDate: Returns an error (Bds Request, Unrecognized field \"billingDate\").
    - Parameter billingStartDate: Works as expected, despite being labeled as the "wrong" field in other parts of the API documentation.

**Environment**
    Endpoint: POST http://127.0.0.1:8080/1.0/kb/subscriptions/createSubscriptionWithAddOns

jutur:
Fixes #2189 
The confusion here isn't actually a query-param inconsistency —
createSubscription, createSubscriptionWithAddOns, and
createSubscriptionsWithAddOns all already accept the same `billingDate`
query parameter. The mixup is with the `billingStartDate` field in the
request/response JSON body, which only reflects the server's computed
value on output and does nothing if set on input.

Added a schema description on that field spelling this out, so it's
clear in the generated docs that billing start is controlled via the
`billingDate` query parameter, not the JSON body.

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionJson.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionJson.java
@@ -74,6 +74,8 @@ public class SubscriptionJson extends JsonBase {
     private final EntitlementSourceType sourceType;
     private final DateTime cancelledDate;
     private final LocalDate chargedThroughDate;
+    @Schema(description = "Effective start of billing for this subscription, populated by the server. To request a specific billing start date " +
+                          "when creating a subscription, use the 'billingDate' query parameter instead of setting this field in the request body.")
     private final DateTime billingStartDate;
     private final DateTime billingEndDate;
     private final Integer billCycleDayLocal;


reshmabidikar:
Thanks for this PR @jutur. However, the request parameter is documented correctly in the [KB docs](https://apidocs.killbill.io/subscription#create-a-subscription). Also, adding a schema description to just this particular API endpoint would make it inconsistent with the rest of the KB code which does not have any schema descriptions. So, I'm going to close this PR for now.

PR #2269, opened 2026-07-10
jaxrs: fix locale handling and typo in InvoiceResource

### LINKED ISSUE #1582: Issues in InvoiceResource
- An extra `locale` parameter is specified in the [InvoiceResource.getInvoiceMPTemplate](https://github.com/killbill/killbill/blob/585d68891b06f3329379548583a33fe94b9250ea/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L801) method definition. This needs to be removed. Once this is fixed, the code for the [Kill Bill Java client](https://github.com/killbill/killbill-client-java) also needs to be regenerated.
- The Java code corresponding to the [Retrieve invoice item audit logs with history](https://killbill.github.io/slate/#invoice-retrieve-invoice-item-audit-logs-with-history-by-invoice-item-id) and [Retrieve invoice payment audit logs with history](https://killbill.github.io/slate/#invoice-retrieve-invoice-payment-audit-logs-with-history-by-invoice-payment-id) is missing from [InvoiceResource](https://github.com/killbill/killbill/blob/585d68891b06f3329379548583a33fe94b9250ea/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java). This needs to be added and the code for the [Kill Bill Java client](https://github.com/killbill/killbill-client-java) needs to be regenerated.
- There is a typo in the error message specified in the `InvoiceResource. generateDryRunInvoice` [here](https://github.com/killbill/killbill/blob/585d68891b06f3329379548583a33fe94b9250ea/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java#L401). So, when the `product name` is missing, the error message indicates that the `product category` is not specified.
- The [Upload Manual Pay Template](https://killbill.github.io/slate/#invoice-upload-the-manualpay-invoice-template-for-the-tenant) endpoint returns a location header containing a `locale` while the [Upload Invoice Template](https://killbill.github.io/slate/#invoice-upload-the-invoice-template-for-the-tenant) does not return a `locale` in the location header. In general, locale does not make sense for templates since templates are independent of locales, so this needs to be investigated further.

jutur:
Fixes #1582
- getInvoiceMPTemplate declared a 'locale' path parameter it never used (always looked up the single non-localized template); this also caused the upload endpoint's Location header to include a spurious locale segment ('/invoiceMPTemplate/en_US/') even though the manualPay template isn't stored per-locale. Removing the unused parameter fixes both.
- Fixed a copy/paste error in generateDryRunInvoice's validation message: the check on getProductName() referenced 'product category' instead of 'product name'.

Note: removing the locale path segment is a client-visible change to getInvoiceMPTemplate and requires regenerating killbill-client-java. The other two items from the original report (missing invoiceItem/ invoicePayment audit log endpoints) are already covered by InvoiceItemResource#getInvoiceItemAuditLogsWithHistory and InvoicePaymentResource#getInvoicePaymentAuditLogsWithHistory.

### DIFF
--- a/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
+++ b/jaxrs/src/main/java/org/killbill/billing/jaxrs/resources/InvoiceResource.java
@@ -469,7 +469,7 @@ public class InvoiceResource extends JaxRsResourceBase {
         if (dryRunSubscriptionSpec != null && dryRunSubscriptionSpec.getDryRunAction() != null) {
             if (SubscriptionEventType.START_BILLING.equals(dryRunSubscriptionSpec.getDryRunAction()) || SubscriptionEventType.CHANGE.equals(dryRunSubscriptionSpec.getDryRunAction())) {
                 if (dryRunSubscriptionSpec.getPlanName() == null) {
-                    verifyNonNullOrEmpty(dryRunSubscriptionSpec.getProductName(), "DryRun subscription product category should be specified when no planName is specified");
+                    verifyNonNullOrEmpty(dryRunSubscriptionSpec.getProductName(), "DryRun subscription product name should be specified when no planName is specified");
                     verifyNonNullOrEmpty(dryRunSubscriptionSpec.getBillingPeriod(), "DryRun subscription billingPeriod should be specified when no planName is specified");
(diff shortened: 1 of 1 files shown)


reshmabidikar (approved):
Thanks for the fix @jutur!

Issue #2304, opened 2026-08-20
Add Support For Partial locale in invoice/catalog translations


reshmabidikar:
Modify the backend code that retrieves invoice translations to support both exact locale matching and language-level fallback.

**Expected Behavior**
1. Exact locale match
  - First, look for translations matching the account's complete locale.
  - For example, if the account locale is `en_GB`, check for translations with the `en_GB` locale.
2. Language-level fallback
 - If no translations are found for the exact locale, fall back to the language component only.
 - For example, if the account locale is `en_GB` and no `en_GB` translations exist, look for translations with the `en` locale.
 - If `en` translations are available, use them as the fallback.

This ensures that locale-specific translations take precedence while allowing language-level translations to be used when an exact locale match is not available.

PR #2305, opened 2026-08-20
Default to language level translation if present

reshmabidikar:
- Work for #2304 (default to language level translations if present)
- Changed the default company information to use KB specific values

### DIFF
--- a/util/src/main/java/org/killbill/billing/util/LocaleUtils.java
+++ b/util/src/main/java/org/killbill/billing/util/LocaleUtils.java
@@ -40,6 +40,20 @@ public class LocaleUtils {
 
     }
 
+    public static String localeStringWithOnlyLanguage(final Locale locale, @Nullable String prefix) {
+        final StringBuilder tmp = new StringBuilder();
+        if (prefix != null) {
+            tmp.append(prefix);
+            if (!prefix.endsWith("_")) {
+                tmp.append("_");
+            }
+        }
+        tmp.append(locale.getLanguage())
+           .append("_");
+        return tmp.toString();
+
+    }
+
     // From commons-lang
     public static Locale toLocale(final String str) {
         if (str == null) {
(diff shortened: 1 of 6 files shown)


util/src/main/java/org/killbill/billing/util/LocaleUtils.java
```diff
@@ -40,6 +40,20 @@ public static String localeString(final Locale locale, @Nullable String prefix)
 
     }
 
+    public static String localeStringWithOnlyLanguage(final Locale locale, @Nullable String prefix) {
```
sbrossie: Can we instead modify the `LocaleUtils#localeString` to support both cases - language _ opt. country. 
You could call it twice using  first `locale` and then  `Locale.of(locale.getLanguage()); `

reshmabidikar: Addressed in f6c79e4

util/src/main/java/org/killbill/billing/util/LocaleUtils.java
```diff
@@ -25,19 +25,31 @@ public class LocaleUtils {
...
+    public static String localeString(final Locale locale, @Nullable final String prefix) {
+        return localeString(locale.getLanguage(), locale.getCountry(), prefix);
+    }
+
+    public static String localeString(final String language, @Nullable final String prefix) {
+        return localeString(language, "", prefix);
+    }
+
+    private static String localeString(final String language,
+                                       final String country,
```
sbrossie, line 37: Nit: I think `@Nullable` should also be for `country` ?

sbrossie (approved):

util/src/main/java/org/killbill/billing/util/LocaleUtils.java
reshmabidikar: Adding @Nullable to `country` results in a spotbug error, so skipping this.

